PR #9, opened 2019-09-10
Add validators to where we call unmarshal func

EngHabu:
This commit  https://github.com/lyft/flyteplugins/commit/99d622cf0f0ca5041e46aad172101536079ea22a relaxed the un-marshaling logic. This PR adds validate* funcs to all places where we unmarshal to ensure the parsed objects are valid.


EngHabu:
PTAL @surindersinghp @wild-endeavor @akhurana001

PR #10, opened 2019-09-10
Allow resource deletion when KillTask is called

EngHabu:
This PR adds a new executorProperty to allow K8s plugins to request deletion of owned resource when KillTask is called


EngHabu:
PTAL @surindersinghp

go/tasks/v1/flytek8s/plugin_executor.go
```diff
@@ -312,14 +312,26 @@ func (e *K8sTaskExecutor) KillTask(ctx context.Context, taskCtx types.TaskContex
...
+
 	// Retrieve the object from cache/etcd to get the last known version.
 	_, _, err = e.getResource(ctx, taskCtx, o)
 	if err != nil {
 		return err
 	}
 
-	return e.ClearFinalizers(ctx, o)
+	// Clear finalizers
+	err = e.ClearFinalizers(ctx, o)
```
surindersinghp: you may want to clear the finalizer after deleting the resource.

PR #12, opened 2019-09-17
Move Hive plugin

wild-endeavor:


kumare3:
I am not sure if you should delete the resource manager, @bnsblue / @EngHabu is the new resource manager updated in the same place

PR #13, opened 2019-09-17
Filter informer events

EngHabu:
Remove reenqueueing for Create, Delete and Generic events. Check for Resource Version on Update events.


go/tasks/v1/flytek8s/mux_handler.go
```diff
@@ -132,29 +137,42 @@ func RegisterResource(ctx context.Context, resourceToWatch runtime.Object, handl
...
+				droppedUpdateCount.Inc(newCtx)
 			}
 		},
 		DeleteFunc: func(evt event.DeleteEvent, q2 workqueue.RateLimitingInterface) {
-			err := handler.Handle(ctx, evt.Object)
-			if err != nil {
-				logger.Warnf(ctx, "Failed to handle Delete event for object [%v]", evt.Object)
-			}
 		},
 		GenericFunc: func(evt event.GenericEvent, q2 workqueue.RateLimitingInterface) {
```
kumare3: lets skip generic events too

kumare3:
Generic events are events that occur potentially from outside of the system and the type is unknown. I do not think this is our case, we are not using this to handle non k8s objects. Lets skip re-enqueuing

PR #14, opened 2019-09-18
Remove Id from Hive execution state

wild-endeavor:
* Move the Id field out of the plugin state object
* Create a wrapper object encapsulating the plugin state object that adds the Id
* Rename some of the fields to be less qubole.
* Remove the explicit Copy() command since everything is pass by value anyways.

Would like to not merge this until we test again with propeller.


wild-endeavor:
ptal @EngHabu @kumare3

PR #15, opened 2019-09-19
Optimize Command line template parsing

kumare3:
- Template parsing is used to substitute command line parameters
- It is more optimal to substitute parameters lazily rather than computing them AOT
- Previous implementaiton computed all the serialized formats ahead of time and sometimes this is very large, causing a lot of GC pressure

TODO, in the plugin refactor, we should use the io.InputReader to even more lazily resolve the inputs

### DIFF
--- a/go/tasks/v1/utils/template.go
+++ b/go/tasks/v1/utils/template.go
@@ -9,7 +9,6 @@ import (
 
 	"github.com/golang/protobuf/ptypes"
 	"github.com/lyft/flyteidl/gen/pb-go/flyteidl/core"
-	"github.com/lyft/flytestdlib/logger"
 )
 
 var inputFileRegex = regexp.MustCompile(`(?i){{\s*[\.$]Input\s*}}`)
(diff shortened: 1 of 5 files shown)


go/tasks/v1/utils/template.go
```diff
@@ -48,9 +47,12 @@ func ReplaceTemplateCommandArgs(ctx context.Context, command []string, args Comm
 	return res, nil
 }
 
-func replaceTemplateCommandArgs(_ context.Context, commandTemplate string, args *CommandLineTemplateArgs) (string, error) {
+func replaceTemplateCommandArgs(ctx context.Context, commandTemplate string, args *CommandLineTemplateArgs) (string, error) {
 	val := inputFileRegex.ReplaceAllString(commandTemplate, args.Input)
 	val = outputRegex.ReplaceAllString(val, args.OutputPrefix)
+	if args.Inputs == nil || args.Inputs.Literals == nil {
```
EngHabu: We should probably move this block down a couple of lines... basically:
If we try to reference an input in the template AND there are no inputs in args, then fail...

PR #19, opened 2019-09-26
using task context secret manager

wild-endeavor:
* Deleting the old secrets manager and using the one provided by the TaskExecutionContext instead for Hive plugin
* Adding SecretManager() to the SetupContext


wild-endeavor:
ptal @kumare3

kumare3:
LGTM

PR #22, opened 2019-09-30
Fix Redis frm off by one error

wild-endeavor:
0 is not > 0

This is preventing us from gating FRM fully.


wild-endeavor:
ptal @kumare3 @EngHabu

wild-endeavor:
💨

PR #26, opened 2019-10-09
Delete sidecar pod when task completes.

surindersinghp:


go/tasks/v1/flytek8s/plugin_executor.go
```diff
@@ -283,10 +283,20 @@ func (e *K8sTaskExecutor) CheckTaskStatus(ctx context.Context, taskCtx types.Tas
...
+	if finalStatus.Phase.IsTerminal() {
+		if en(o.GetFinalizers()) > 0 {
+			err = e.ClearFinalizers(ctx, o)
+			if err != nil {
+				return types.TaskStatusUndefined, err
+			}
+		}
+
+		// kill the object execution if still live
+		if e.handler.GetProperties().DeleteResourceOnAbort {
```
kumare3: i dont think we can do this, as this removes the idempotency property. We can do this in case of failure only if at all as it may result in an earlier failure

surindersinghp: updated the PR. I am going to add tests in a bit. Meanwhile, can you look at it?

surindersinghp: there is still an issue, dont look at it yet.

surindersinghp: this should work. I was worried if CheckNodestatus will even call task if task is successful. It seems it will so should be good.
https://github.com/lyft/flytepropeller/blob/master/pkg/controller/nodes/executor.go#L155

go/tasks/v1/flytek8s/plugin_executor.go
```diff
@@ -283,10 +312,18 @@ func (e *K8sTaskExecutor) CheckTaskStatus(ctx context.Context, taskCtx types.Tas
...
+			err = e.ClearFinalizers(ctx, o)
+			if err != nil {
+				return types.TaskStatusUndefined, err
+			}
+		}
+
+		finalStatus = types.TaskStatus{
+			Phase:        taskCtx.GetPhase(),
+			PhaseVersion: taskCtx.GetPhaseVersion(),
+			State:        storeK8sObjectStatus(k8sObjectDeleted, finalStatus.Phase),
```
EngHabu: The actual handler might be using a state on its own? I guess it's fine since we are deprecating this code anyway...

go/tasks/v1/flytek8s/k8splugin_state.go
```diff
@@ -0,0 +1,70 @@
...
+}
+
+// This status internal state of the object not read/updated by upstream components (eg. Node manager)
+type K8sObjectState struct {
+	Status        K8sObjectStatus `json:"s"`
+	TerminalPhase types.TaskPhase `json:"tp"`
+}
+
+func retrieveK8sObjectStatus(customState map[string]interface{}) (K8sObjectStatus, types.TaskPhase, error) {
+	for k, v := range customState {
```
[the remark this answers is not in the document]
surindersinghp: :)

go/tasks/v1/flytek8s/plugin_executor.go
surindersinghp: I checked, none of the handlers today are :). It will go away soon.

go/tasks/v1/flytek8s/plugin_executor.go
```diff
@@ -243,6 +243,35 @@ func (e *K8sTaskExecutor) CheckTaskStatus(ctx context.Context, taskCtx types.Tas
...
+				if IsK8sObjectNotExists(err) {
+					logger.Debugf(ctx, "the k8s object %v was found to have successfully exited after completion", taskCtx.GetTaskExecutionID().GetGeneratedName())
+				} else {
+					return types.TaskStatusUndefined, err
+				}
+			} else {
+				logger.Debugf(ctx, "deleted the k8s object %v in terminal phase", taskCtx.GetTaskExecutionID().GetGeneratedName())
+			}
+		}
+		finalStatus.Phase = terminalPhase
```
kumare3: if the terminalPhase is failed
 then set error code and error message

surindersinghp: done

go/tasks/v1/flytek8s/plugin_executor.go
```diff
@@ -243,6 +244,38 @@ func (e *K8sTaskExecutor) CheckTaskStatus(ctx context.Context, taskCtx types.Tas
...
+	objStatus, terminalPhase, err := retrieveK8sObjectState(taskCtx.GetCustomState())
+	if err != nil {
+		logger.Warningf(ctx, "Failed to retrieve object status: %v. Error: %v",
+			taskCtx.GetTaskExecutionID().GetGeneratedName(), err)
+		return types.TaskStatusUndefined, err
+	}
+	if objStatus == k8sObjectDeleted {
+		// kill the object execution if still live
+		if e.handler.GetProperties().DeleteResourceOnAbort {
+			err = instance.kubeClient.Delete(ctx, o)
```
EngHabu: I think we need to call AddObjectMetadata here before calling delete

surindersinghp: done

PR #31, opened 2019-11-13
[Array] Ensure unique env vars

EngHabu:


matthewphsmith:
Do we have a test that explicitly covers duplicate env vars?

PR #32, opened 2019-11-13
 Proposal: Common plugin interface for all external service plugins

kumare3:
Proposal: Make a simplified interface to add external service plugins, example Qubole, BigQuery, DataFlow etc

Aim: the idea is to take away the boilerplate or writing the statemachine etc for such simple plugins, including caching etc

Status: Work in progress


go/tasks/pluginmachinery/service/plugin.go
```diff
@@ -0,0 +1,74 @@
...
+
+// The resource to be sycned from the remote
+type RemoteResource interface{}
+
+// Defines a simplified interface to author plugins for k8s resources.
+type Plugin interface {
+	GetPluginProperties() PluginProperties
+	// Create a new resource using the PluginContext provided. Ideally, the remote service uses the name in the TaskExecutionMetadata
+	// to launch the resource idempotently
+	Create(ctx context.Context, tCtx PluginContext) (RemoteResourceKey, error)
```
EngHabu: I think create should -maybe optionally- return the RemoteResource too
```suggestion
	Create(ctx context.Context, tCtx PluginContext) (RemoteResourceKey, RemoteResource, error)
```

go/tasks/pluginmachinery/service/plugin.go
```diff
@@ -0,0 +1,74 @@
...
+	// Get the entire object including the status from the remote api
+	Get(ctx context.Context, key RemoteResourceKey) (RemoteResource, error)
```
EngHabu: We should probably pass some context here that has TaskExecutionMetadata
```suggestion
	Get(ctx context.Context, tCtx PluginContext, key RemoteResourceKey) (RemoteResource, error)
```

go/tasks/pluginmachinery/service/plugin.go
```diff
@@ -0,0 +1,74 @@
...
+	// Get multiple resources that match all the keys. This method is optional and should be provided in case the
+	// PluginProperties.BatchingProperties.BatchAPISupported = true
+	GetBatch(ctx context.Context, key ...RemoteResourceKey) (RemoteResource, error)
```
EngHabu: If we do not have an immediate need for Batch Api, I would rather not have this and -probably- have a separate Plugin interface for batch-like plugins...

go/tasks/pluginmachinery/service/properties.go
```diff
@@ -0,0 +1,39 @@
...
+
+type MissingResourceSemantics uint8
+
+const (
+	MissingResourceFailIfMissing MissingResourceSemantics = iota
+	MissingResourceRetryIfMissing
+)
+
+// Properties that indicate if the Batch API can be invoked. This indicates that GetBatch() can be used instead of Get
+type BatchingProperties struct {
```
EngHabu: Maybe delete

go/tasks/pluginmachinery/service/properties.go
```diff
@@ -0,0 +1,39 @@
...
+	BatchAPISupported        bool
+	MaxBatchSize             int
+	MissingResourceSemantics MissingResourceSemantics
```
EngHabu: Why is this only in Batching properties?

go/tasks/pluginmachinery/service/properties.go
```diff
@@ -0,0 +1,39 @@
...
+}
+
+// Properties that help the system optimize itself to handle the specific plugin
+type PluginProperties struct {
+	RateLimiterProperties RateLimiterProperties
```
EngHabu: We should use the ratelimiter object from stdlib

go/tasks/pluginmachinery/service/properties.go
```diff
@@ -0,0 +1,39 @@
+package service
+
+import "time"
+
+// The plugin manager automatically queries the remote API
+type RateLimiterProperties struct {
```
EngHabu: We should probably have separate rate limiters for `get` and `create`

go/tasks/pluginmachinery/service/plugin.go
kumare3: agreed

go/tasks/pluginmachinery/service/plugin.go
kumare3: agreed

go/tasks/pluginmachinery/service/plugin.go
kumare3: but no resource does it mean a failure?

go/tasks/pluginmachinery/service/properties.go
kumare3: because in batching the batch may return missing resources?

go/tasks/pluginmachinery/service/properties.go
kumare3: ya but do you want to expose that or should we expose just the properties and configure it ourselves

go/tasks/pluginmachinery/service/properties.go
kumare3: good idea

PR #33, opened 2019-11-15
Fix array log link phase to show interpreted phase instead

EngHabu:
- Fix array log link phase to show interpreted phase instead
- Add unit tests for env vars override in array tasks


go/tasks/plugins/array/awsbatch/task_links.go
```diff
@@ -70,11 +70,15 @@ func GetTaskLinks(ctx context.Context, taskMeta pluginCore.TaskExecutionMetadata
 
 	for childIdx, subJob := range job.SubJobs {
 		originalIndex := core.CalculateOriginalIndex(childIdx, state.GetIndexesToCache())
+		detailedArrayStatus := state.GetArrayStatus().Detailed
```
wild-endeavor: should this be out of the loop?

PR #34, opened 2019-11-18
qubole validation for older versions of the sdk

wild-endeavor:
Prior versions of the SDK did not fill in this value.  When this code was written, it was assumed that all prior versions of the SDK would've been deprecated, which was a bad assumption.


wild-endeavor:
ptal @EngHabu @matthewphsmith

PR #36, opened 2019-11-28
Migrate to go module and remove old Hive integration tests

honnix:
Some tests are being removed because they need to be updated and that breaks go mod.  Can add back in the future.  For some reason go mod doesn't know how to respect build tags.


honnix:
@kumare3 PTAL

honnix:
I had to delete some test files under `tests` because they are not compilable.

honnix:
@wild-endeavor PTAL. I had to delete those non-compilable code otherwise go module is sad.

wild-endeavor:
I think it's fine - those tests were written a while ago and then not kept up to date since we still don't have a cohesive integration testing story.  Let me change the title of the PR though to make them easier to find in case I ever want to resurrect them.

go.mod
```diff
@@ -0,0 +1,38 @@
...
+)
+
+replace (
+	github.com/GoogleCloudPlatform/spark-on-k8s-operator => github.com/lyft/spark-on-k8s-operator v0.1.3
+	k8s.io/api => github.com/lyft/api v0.0.0-20191031200350-b49a72c274e0
+	k8s.io/apimachinery => github.com/lyft/apimachinery v0.0.0-20191031200210-047e3ea32d7f
+	k8s.io/client-go => k8s.io/client-go v0.0.0-20191016111102-bec269661e48
+)
+
+// k8s.io/client-go kubernetes-1.17.2
```
EngHabu: If we moved that to the replace section, would it be overridden on update?

tests/hive_integration_test.go
```diff
@@ -1,201 +0,0 @@
-// +build manualintegration
```
EngHabu: Do we need to delete this?

go.mod
```diff
@@ -0,0 +1,38 @@
...
+	github.com/spf13/pflag v1.0.5
+	github.com/stretchr/testify v1.4.0
+	golang.org/x/net v0.0.0-20200114155413-6afb5195e5aa
+	google.golang.org/grpc v1.26.0
+	k8s.io/api v0.17.2
+	k8s.io/apimachinery v0.17.2
+	k8s.io/client-go v11.0.0+incompatible
+	k8s.io/klog v1.0.0 // indirect
+	k8s.io/utils v0.0.0-20200122174043-1e243dd1a584 // indirect
+	sigs.k8s.io/controller-runtime v0.4.0
```
EngHabu: We used to depend on a SHA, is this version inclusive?

tests/hive_integration_test.go
wild-endeavor: I don't think so, but they're not working anymore.  I should try to fix them at some point.

go.mod
honnix: I don't quite understand how it would end up like this. Taking a look now.

honnix: Does this kind of comment serve a special purpose? L52 should do the job if we want to pin a version.

honnix:
I ran `go mod tidy` on this PR and `go.sum` got modified. Can you try that as well @wild-endeavor ?

go.mod
wild-endeavor: https://github.com/kubernetes-sigs/controller-runtime/releases/tag/v0.4.0 includes sha 40070e2a1958c3d974ba95da883a2bd088137789

wild-endeavor:
ran go mod tidy as well.

go.mod
wild-endeavor: removed this comment, added another one.

go.mod
```diff
@@ -28,11 +28,14 @@ require (
 	sigs.k8s.io/controller-runtime v0.4.0
 )
 
+// Pin the version of client-go to something that's compatible with katrogan's fork of api and apimachinery
```
honnix: 👍

honnix: Now looks much clearer. To me it has been hard to know whether a comment serves anything in go.

PR #39, opened 2019-12-16
Make catalog async client array-aware

EngHabu:


go/tasks/pluginmachinery/catalog/reader_array_processor.go
```diff
@@ -0,0 +1,131 @@
...
+	if !casted {
+		return workqueue.WorkStatusNotDone, fmt.Errorf("wrong work item type. Received: %v", reflect.TypeOf(workItem))
+	}
+
+	if wi.cached == nil {
+		wi.cached = bitarray.NewBitSet(uint(wi.downloadRequest.Count))
+		wi.progress = bitarray.NewBitSet(uint(wi.downloadRequest.Count))
+	}
+
+	isArray := wi.downloadRequest.Count > 0
```
wild-endeavor: Can you add a comment for this?  Don't understand.  We still want the loop to run even if the Count is 0, why?  When is the count 0?

go/tasks/pluginmachinery/catalog/writer_array_processor.go
```diff
@@ -0,0 +1,116 @@
...
+				wi.request.BaseArtifactData.GetOutputPrefixPath(),
+				strconv.Itoa(int(i)))
+			if err != nil {
+				return workqueue.WorkStatusNotDone, err
+			}
+
+			// TODO: size limit is weird to be passed here...
+			outputReader = ioutils.NewRemoteFileOutputReader(ctx, wi.request.dataStore,
+				ioutils.NewRemoteFileOutputPaths(ctx, wi.request.dataStore, indexedOutputLocation),
+				int64(999999999))
```
wild-endeavor: Is this in bytes?   953MBs?

PR #40, opened 2019-12-17
Allow specifying default env vars to read from env

katrogan:


katrogan:
ping @wild-endeavor @EngHabu

PR #44, opened 2020-01-06
Hive plugin: using primary cluster labels to execute queries and manage resource; simplify hive plugin init interface 

bnsblue:
This is to simplify the init interface for hive plugin, specifically for initializing the resource manager.

There are many more lines of change than it should have been, probably because we haven't been running  `make generate` for a long while.

Related PRs:
https://github.com/lyft/flytepropeller/pull/54
https://github.com/lyft/flyteplugins-private/pull/12

Edit 1/8/2020:
Adding config and logic to make sure we use primary cluster labels to execute queries and manage resource


go/tasks/pluginmachinery/flytek8s/config/config.go
```diff
@@ -35,7 +35,7 @@ type K8sPluginConfig struct {
 	// Provide additional environment variable pairs that plugin authors will provide to containers
 	DefaultEnvVars map[string]string `json:"default-env-vars" pflag:"-,Additional environment variable that should be injected into every resource"`
 	// Provide additional environment variable pairs whose values resolve from the plugin's execution environment.
-	DefaultEnvVarsFromEnv map[string]string `json:"default-env-vars-from-env" pflag:",Additional environment variable that should be injected into every resource"`
+	DefaultEnvVarsFromEnv map[string]string `json:"default-env-vars-from-env" pflag:"-,Additional environment variable that should be injected into every resource"`
```
wild-endeavor: what does this do?

bnsblue: The pflags tool is not able to generate command for `map`s. Without the `-` the `make generate` would fail. This is to disable it

bnsblue: Let me know if it doesn't make sense.

katrogan: thank you for fixing!

go/tasks/plugins/hive/config/config.go
```diff
@@ -25,31 +25,34 @@ func MustParse(s string) config.URL {
 	return config.URL{URL: *r}
 }
 
+type ClusterConfig struct {
+	Label string `json:"label" pflag:",Label of the service cluster"`
+	Limit int    `json:"limit" pflag:",Resource quota of the service cluster"`
```
katrogan: can you add what the units are either in the pflag or variable name (e.g. LimitInFoo)

wild-endeavor:
+1 I think the only thing that is unclear is how cluster tags are handled on the Qubole side... What i mean is, if one cluster has two tags, `x` and `y`, and both are used by users, this will create two resource manager namespaces right?  I don't think this will really be a problem, just wanted to make sure I'm not missing something.

Also, I don't remember the specifics, but does this address the non-unique issue @EngHabu was talking about?

go/tasks/plugins/hive/executor.go
```diff
@@ -94,7 +96,7 @@ func (q QuboleHiveExecutor) Finalize(ctx context.Context, tCtx core.TaskExecutio
 		return errors.Wrapf(errors.CorruptedPluginState, err, "Failed to unmarshal custom state in Finalize")
 	}
 
-	return Finalize(ctx, tCtx, q.resourceNamespace, incomingState)
+	return Finalize(ctx, tCtx, q.GetResourceNamespace(), incomingState)
```
wild-endeavor: One more question - what happens if a user requests a resource from say, a new cluster in Qubole, that has not yet been added to the Config object?  It'll fail right?

bnsblue:
@wild-endeavor  
After this fix, the tokens will be in the format of "ex:project:domain:exec_id:token" so each of them will be unique universally. 
However I am not entirely familiar with the concept of tags and what it is used for, so I am not sure if this PR addresses the issue you mentioned in the first paragraph. Mind to elaborate a bit?

go/tasks/plugins/hive/executor.go
bnsblue: Good point. I think I'll panic. Let me fix it

go/tasks/plugins/hive/config/config.go
```diff
@@ -26,33 +26,42 @@ func MustParse(s string) config.URL {
...
 
 type ClusterConfig struct {
-	Label string `json:"label" pflag:",Label of the service cluster"`
-	Limit int    `json:"limit" pflag:",Resource quota (in the number of outstanding requests) of the service cluster"`
+	PrimaryLabel string   `json:"primaryLabel" pflag:",The primary label of a given service cluster"`
+	Labels       []string `json:"labels" pflag:",Labels of a given service cluster"`
+	Limit        int      `json:"limit" pflag:",Resource quota (in the number of outstanding requests) of the service cluster"`
+}
+
+type ProjectDestinationClusterConfig struct {
```
wild-endeavor: Can you rename to either ProjectDomainDestinationClusterConfig?  Or maybe just DestinationClusterConfig?

go/tasks/plugins/hive/executor.go
```diff
@@ -125,9 +127,9 @@ func InitializeHiveExecutor(ctx context.Context, iCtx core.SetupContext, cfg *co
 		return nil, err
 	}
 
-	for clusterLabel, clusterLimit := range resourceConfig {
-		namespaceWithClusterLabel := resourceNamespace.CreateSubNamespace(core.ResourceNamespace(clusterLabel))
-		if err := iCtx.ResourceRegistrar().RegisterResourceQuota(ctx, namespaceWithClusterLabel, clusterLimit); err != nil {
+	for clusterPrimaryLabel, clusterLimit := range resourceConfig {
+		namespaceWithClusterPrimaryLabel := resourceNamespace.CreateSubNamespace(core.ResourceNamespace(clusterPrimaryLabel))
+		if err := iCtx.ResourceRegistrar().RegisterResourceQuota(ctx, namespaceWithClusterPrimaryLabel, clusterLimit); err != nil {
```
wild-endeavor: If the `RegisterResourceQuota` function doesn't have a logger.Infof, can you add one here?

PR #47, opened 2020-01-21
use args instead of cmd

migueltol22:
For dynamic tasks we set the spec to use command instead of args. This causes the base image entrypoint not to be ran when specified. 

Notes:
* Verified by running dynamic tasks workflow.


vsbus:
Nice! Thank you!

PR #48, opened 2020-01-22
Add demystify success for OOMKilled

katrogan:
If a container in a Pod fails with OOMKilled, the overall Pod Phase is still marked as Succeeded leading plugins/Propeller to assume it has succeeded and fail later on with a confusing error

It's worth noting that kubectl seem to be correctly determining the pod status, because it does return OOMKilled as the status of the pod:

```
kubectl get pod -n priceoptimizeroffline-development fyxxs4ei
 NAME READY STATUS RESTARTS AGE
 fyxxs4ei 0/1 OOMKilled 0 16h
Pod status:
```

```
status:
  ...
  containerStatuses:
  ...
    state:
      terminated:
        exitCode: 0
        finishedAt: "2019-08-29T00:51:41Z"
        reason: OOMKilled
        startedAt: "2019-08-29T00:51:22Z"
  ...
  phase: Succeeded
```

### DIFF
--- a/go/tasks/pluginmachinery/flytek8s/pod_helper.go
+++ b/go/tasks/pluginmachinery/flytek8s/pod_helper.go
@@ -3,6 +3,7 @@ package flytek8s
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/lyft/flytestdlib/logger"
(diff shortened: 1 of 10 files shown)


go/tasks/pluginmachinery/flytek8s/pod_helper.go
```diff
@@ -167,6 +169,16 @@ func DemystifyPending(status v1.PodStatus) (pluginsCore.PhaseInfo, error) {
 	return pluginsCore.PhaseInfoQueued(time.Now(), pluginsCore.DefaultPhaseVersion, "Scheduling"), nil
 }
 
+func DemystifySuccess(status v1.PodStatus, info pluginsCore.TaskInfo) (pluginsCore.PhaseInfo, error) {
+	for _, status := range status.ContainerStatuses {
```
EngHabu: Should we check InitContainer Statuses as well?

katrogan: @EngHabu sure, done

PR #50, opened 2020-02-04
[MB-1972] Also check failed pod state for OOMKilled

katrogan:
Example from an OOM-d task:

```containerStatuses:
  - containerID: docker://be556e0d4edeb6435d3f45429334cb8323ea28c3b82de81f44364e2c3de43739
    image: flyteexamples:0510db793bb658155c4175d005b51bf83bdac2f5
    imageID: docker://sha256:0a6270ad43ae5f9d84fbd530206625fdb558529e3a252c118dbfd9226edfd698
    lastState: {}
    name: f0601d8477447476d919-oom-task-0
    ready: false
    restartCount: 0
    state:
      terminated:
        containerID: docker://be556e0d4edeb6435d3f45429334cb8323ea28c3b82de81f44364e2c3de43739
        exitCode: 137
        finishedAt: "2020-02-05T19:50:15Z"
        reason: OOMKilled
        startedAt: "2020-02-05T19:50:13Z"
```

### DIFF
--- a/go/tasks/pluginmachinery/flytek8s/pod_helper.go
+++ b/go/tasks/pluginmachinery/flytek8s/pod_helper.go
@@ -170,13 +170,8 @@ func DemystifyPending(status v1.PodStatus) (pluginsCore.PhaseInfo, error) {
 }
 
 func DemystifySuccess(status v1.PodStatus, info pluginsCore.TaskInfo) (pluginsCore.PhaseInfo, error) {
-	for _, status := range status.ContainerStatuses {
-		if status.State.Terminated != nil && strings.Contains(status.State.Terminated.Reason, OOMKilled) {
-			return pluginsCore.PhaseInfoRetryableFailure("OOMKilled",
-				"Pod reported success despite being OOMKilled", &info), nil
-		}
-	}
-	for _, status := range status.InitContainerStatuses {
+	for _, status := range append(
+		append(status.InitContainerStatuses, status.ContainerStatuses...), status.EphemeralContainerStatuses...) {
 		if status.State.Terminated != nil && strings.Contains(status.State.Terminated.Reason, OOMKilled) {
 			return pluginsCore.PhaseInfoRetryableFailure("OOMKilled",
 				"Pod reported success despite being OOMKilled", &info), nil
--- a/go/tasks/pluginmachinery/flytek8s/pod_helper.go
+++ b/go/tasks/pluginmachinery/flytek8s/pod_helper.go
@@ -194,47 +194,44 @@ func ConvertPodFailureToError(status v1.PodStatus) (code, message string) {
 
 	if len(status.Message) > 0 {
 		message = status.Message
-	} else {
-		for _, c := range status.InitContainerStatuses {
-			if c.LastTerminationState.Terminated != nil {
-				message += fmt.Sprintf("\r\nInit Container [%v] terminated with exit code (%v). Reason [%v]. Message: [%v].",
-					c.ContainerID,
-					c.LastTerminationState.Terminated.ExitCode,
-					c.LastTerminationState.Terminated.Reason,
-					c.LastTerminationState.Terminated.Message)
-			} else if c.State.Terminated != nil && strings.Contains(c.State.Terminated.Reason, OOMKilled) {
-				code = OOMKilled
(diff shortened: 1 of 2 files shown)


go/tasks/pluginmachinery/flytek8s/pod_helper.go
```diff
@@ -202,6 +202,9 @@ func ConvertPodFailureToError(status v1.PodStatus) (code, message string) {
 					c.LastTerminationState.Terminated.ExitCode,
 					c.LastTerminationState.Terminated.Reason,
 					c.LastTerminationState.Terminated.Message)
+			} else if c.State.Terminated != nil && strings.Contains(c.State.Terminated.Reason, OOMKilled) {
```
chanadian: Seems like we use this logic a lot (also in the method `DemystifySuccess`), perhaps we can encapsulate it in a method like `isOOMKilled()`?

go/tasks/pluginmachinery/flytek8s/pod_helper.go
```diff
@@ -212,6 +215,9 @@ func ConvertPodFailureToError(status v1.PodStatus) (code, message string) {
...
+				code = OOMKilled
+				message += fmt.Sprintf("\r\nContainer [%v] terminated due to being OOMKilled", c.ContainerID)
```
chanadian: it may help users if we can display the resource settings for their pod in the message too.

EngHabu: Should we add the c.State.Terminated.Reason to the message?

go/tasks/pluginmachinery/flytek8s/pod_helper.go
```diff
@@ -202,6 +202,9 @@ func ConvertPodFailureToError(status v1.PodStatus) (code, message string) {
 					c.LastTerminationState.Terminated.ExitCode,
```
EngHabu: Should we change the loop to be:
`for _, c := range append(status.InitContainerStatuses, status.ContainerStatuses...) {`

and just keep it as one loop?

go/tasks/pluginmachinery/flytek8s/pod_helper.go
```diff
@@ -212,6 +215,9 @@ func ConvertPodFailureToError(status v1.PodStatus) (code, message string) {
...
 					c.LastTerminationState.Terminated.Reason,
 					c.LastTerminationState.Terminated.Message)
+			} else if c.State.Terminated != nil && strings.Contains(c.State.Terminated.Reason, OOMKilled) {
```
EngHabu: Is this correct? as in `c.LastTerminationState.Terminated` will be nil in the case of OOMKilled? have you verified that?

go/tasks/pluginmachinery/flytek8s/pod_helper.go
EngHabu: I guess basically keep the message from the previous block, and just set the code as OOMKilled?

go/tasks/pluginmachinery/flytek8s/pod_helper.go
katrogan: hm i'm not a huge fan of refactoring one liners (especially using haytham's suggestion below) but let me know if you feel strongly

go/tasks/pluginmachinery/flytek8s/pod_helper.go
katrogan: we already show the resource requests and limits in the task details tab in the UI!

go/tasks/pluginmachinery/flytek8s/pod_helper.go
katrogan: yep, see: https://gist.github.com/katrogan/a4711f31af0af4b851195e5012313af0 for an example of the full pod output

go/tasks/pluginmachinery/flytek8s/pod_helper.go
katrogan: the messaging is slightly different based on the type of container so i can't think of an elegant way to pass that along

katrogan:
please take another look @EngHabu @chanadian

go/tasks/pluginmachinery/flytek8s/pod_helper.go
katrogan: sure

PR #51, opened 2020-02-05
[AWS Array] Use flytek8s logic to compute resource requirements

EngHabu:
Fix AWS Array Plugin resource computation to consolidate TaskTemplate's resource requirements, Node overrides and default resource configs when building AWS Batch Input.

Prior code only looked at TaskTemplate's resource limits, in one user-scenario, they only set the request not the limit and that code ended up passing default values instead.

This does remove code duplication as well, and ensures K8s and Batch code paths use the same logic to compute resource requirements to share future fixes.


bnsblue:
Can you add some description to the PR and link the ticket / slack thread if you have it?

go/tasks/plugins/array/awsbatch/transformer_test.go
```diff
@@ -45,6 +47,26 @@ func ref(s string) *string {
 	return &s
 }
 
+func TestResourceRequirementsToBatchRequirements(t *testing.T) {
+	memoryTests := []struct {
+		Input    string
+		Expected int64
+	}{
+		// resource Quantity gets the ceiling of values to the nearest scale
+		{"200", 1},
```
bnsblue: Typo? This shouldn't pass, right?

go/tasks/plugins/array/awsbatch/transformer_test.go
```diff
@@ -58,8 +80,8 @@ func TestArrayJobToBatchInput(t *testing.T) {
 			Environment: []*batch.KeyValuePair{
 				{Name: refStr("BATCH_JOB_ARRAY_INDEX_VAR_NAME"), Value: refStr("AWS_BATCH_JOB_ARRAY_INDEX")},
 			},
-			Memory: refInt(700),
-			Vcpus:  refInt(2),
+			Memory: refInt(1074),
```
bnsblue: Any specific reason why this and the following constants need to be changed?

go/tasks/plugins/array/awsbatch/transformer_test.go
bnsblue: Oh, I see. Rounding up to 1?

go/tasks/plugins/array/awsbatch/transformer_test.go
EngHabu: We used to have AWS Batch defaults (hard coded) that were separate from K8s defaults (driven from config). Now that we use the same code, we no longer need this diversion.

PR #52, opened 2020-02-05
Fix failed subtask outputs

EngHabu:


go/tasks/plugins/array/outputs_test.go
```diff
@@ -223,9 +247,24 @@ func TestAssembleFinalOutputs(t *testing.T) {
 		_, err := AssembleFinalOutputs(ctx, assemblyQueue, tCtx, arrayCore.PhaseSuccess, s)
 		assert.NoError(t, err)
 		assert.Equal(t, arrayCore.PhaseRetryableFailure, s.CurrentPhase)
+		assert.False(t, called)
 	})
 
 	t.Run("Not Found Queued then Succeeded", func(t *testing.T) {
+		q := &mocks.IndexedWorkQueue{}
+		called := false
+		q.On("Queue", mock.Anything, mock.Anything, mock.Anything).Return(
```
bnsblue: Is there any specific reason why you use On("Queue") instead of OnQueue() here? 
it would be great if the style can be consistent with the other calls

EngHabu: Discussed offline. The generated OnQueue doesn't let you execute a function to compute the output. It's a great piece of feedback to add though

PR #56, opened 2020-02-18
Fairness in resource manager: adding support for project-level and namespace-level cap

bnsblue:
# Fairness in Resource Manager
### Design doc (Lyft-internal) 
https://docs.google.com/document/d/1cKeD8gptdmSxB47d6uUpfVqxmsTLbfJ6jLj2XqeIajs/edit#

### Context (directly extracted from the doc)
Currently, all of the executions of the same plugin task compete to allocate the resource. There is no queue in front of each resource, nor is there any mechanism in place to prevent total domination by any single project or namespace. 

Also, currently, Flyte resource manager is Flyte-cluster-separated, meaning each of the Flyte clusters has a quota that is not shared with any other Flyte clusters. While this setting prevents the problem of all the quota being taken by one single workflow execution, it still has no fairness guarantee at the project or namespace level, so the resource quota can still be totally dominated by a single project/namespace. Also, it potentially leads to a resource underutilization problem if all the resource-intensive requests happened to be scheduled onto the same Flyte cluster.

### About this PR
This PR is an approach to achieve version-0 fairness in Flyte's resource manager. It adds support for project-level and namespace-level capping on resource usage. It allows a plugin to define the caps in the plugin's config using 
```
ProjectScopeQuotaProportionCap float64 `json:"projectScopeQuotaProportionCap" pflag:",A floating point number between 0 and 1, specifying the maximum proportion of quotas allowed to allocate to a project in the service cluster"`
```
and 
```
NamespaceScopeQuotaProportionCap float64 `json:"namespaceScopeQuotaProportionCap" pflag:",A floating point number between 0 and 1, specifying the maximum proportion of quotas allowed to allocate to a namespace in the service cluster"`
```


go/tasks/pluginmachinery/core/resource_manager.go
```diff
@@ -36,6 +36,17 @@ type ResourceRegistrar interface {
 // Resource Manager manages a single resource type, and each allocation is of size one
 type ResourceManager interface {
 	GetID() string
-	AllocateResource(ctx context.Context, namespace ResourceNamespace, allocationToken string) (AllocationStatus, error)
+	AllocateResource(ctx context.Context, namespace ResourceNamespace, allocationToken string, constraintsSpec ResourceConstraintsSpec) (AllocationStatus, error)
```
kumare3: isnt this a little weird in the API, where the plugin writer needs to provide the constraintspec?

kumare3: my preference is that for a task execution context some of these things are provided by the task handler (in propeller)

go/tasks/plugins/hive/config/config.go
```diff
@@ -54,14 +56,14 @@ var (
...
-	CommandAPIPath            config.URL                 `json:"commandApiPath" pflag:",API Path where commands can be launched on Qubole. Should be a valid url."`
-	AnalyzeLinkPath           config.URL                 `json:"analyzeLinkPath" pflag:",URL path where queries can be visualized on qubole website. Should be a valid url."`
-	TokenKey                  string                     `json:"quboleTokenKey" pflag:",Name of the key where to find Qubole token in the secret manager."`
-	LruCacheSize              int                        `json:"lruCacheSize" pflag:",Size of the AutoRefreshCache"`
-	Workers                   int                        `json:"workers" pflag:",Number of parallel workers to refresh the cache"`
-	ClusterConfigs            []ClusterConfig            `json:"clusterConfigs" pflag:"-,A list of cluster configs. Each of the configs corresponds to a service cluster"`
-	DestinationClusterConfigs []DestinationClusterConfig `json:"destinationClusterConfigs" pflag:"-,A list configs specifying the destination service cluster for (project, domain)"`
+	Endpoint                    config.URL                 `json:"endpoint" pflag:",Endpoint for qubole to use"`
```
kumare3: what changed in this file?

go/tasks/plugins/hive/execution_state.go
```diff
@@ -153,6 +153,27 @@ func composeResourceNamespaceWithClusterPrimaryLabel(ctx context.Context, tCtx c
 	return core.ResourceNamespace(clusterPrimaryLabel), nil
 }
 
+func createResourceConstraintsSpec(ctx context.Context, _ core.TaskExecutionContext, targetClusterPrimaryLabel core.ResourceNamespace) core.ResourceConstraintsSpec {
```
kumare3: i guess there is a benefit to doing this (provide the spec for every call), the benefit is it could allow dynamic adjustment of quota. One problem is that the resolution may not be deterministic if we allow dynamic adjustment

wild-endeavor:
The code itself is fine, but is there a broader description somewhere of how all this will work?  I'm sure there was a design doc that I missed.

go/tasks/plugins/hive/config/config.go
bnsblue: I think originally I added an item in that config struct but then I moved it to ClusterConfig. But my `gofmt` should act automatically on save ...weird...

go/tasks/plugins/hive/execution_state.go
bnsblue: Yes, the addition of this is to enable dynamically adjustable constraints.

bnsblue: Could you elaborate a bit on what you mean by "resolution not being deterministic?"  @kumare3

go/tasks/pluginmachinery/core/resource_manager.go
bnsblue: I designed it this way because I want to be able to allow different cap settings for different service clusters (e.g., Qubole clusters). For example, the `default` Qubole cluster should cap any namespace at 70% of the quotas, but we do NOT want the same thing in the `priceoptimizeroffline` Qubole cluster because the entire quotas of that cluster are dedicated to one team only. 

Currently, the config item `ClusterConfig` resides in the plugin's config map. That's where I find most natural for these cluster-specific cap settings.

kumare3:
> The code itself is fine, but is there a broader description somewhere of how all this will work? I'm sure there was a design doc that I missed.

Yes lets add a better description too, instead of what was added, why was it added.

bnsblue:
I had put up this doc. Take it with a grain of salt though because it was written before I started the implementation, so it might lack some implementation details. But the context and the reasonings are there

https://docs.google.com/document/d/1cKeD8gptdmSxB47d6uUpfVqxmsTLbfJ6jLj2XqeIajs/edit#heading=h.pzd821w0lwrk

bnsblue:
@kumare3 @wild-endeavor
I added a link to the doc, and also extracted a part of the text from the doc because the doc is internal to the company.

go/tasks/plugins/hive/execution_state.go
kumare3: we discussed offline. So we are going to call out the caveats and explain the semantics in the documentation

go/tasks/pluginmachinery/core/resource_manager.go
```diff
@@ -33,9 +33,44 @@ type ResourceRegistrar interface {
 	RegisterResourceQuota(ctx context.Context, namespace ResourceNamespace, quota int) error
 }
 
-// Resource Manager manages a single resource type, and each allocation is of size one
+// Resource Manager manages type, and each allocation is of size one
```
kumare3: `Resource manager manages type` -> ResourceManager provides a task type specific pooling system for Flyte Tasks.
Plugin writers can optionally request for resources in their tasks, in single quantity. (Now explain how the usage should be)

kumare3: I guess i would clean up this documentation to follow a format like
```
1. Define terms, Resource, allocation token etc...
2. describe resource manager
3. Explain usage
4. Example (Qubole that you have)
```

kumare3:
Looks Great, Thank you @bnsblue

PR #60, opened 2020-02-26
Fix AWS Batch CPU Scale

EngHabu:
Use quantity.Value() instead of quantity.ScaledValue() for CPU


anandswaminathan:
+1

go/tasks/plugins/array/awsbatch/transformer.go
```diff
@@ -169,7 +169,7 @@ func toContainerOverrides(ctx context.Context, command []string, overrides *v1.R
 
 	return &batch.ContainerOverrides{
 		Memory:      refInt(overrides.Limits.Memory().ScaledValue(resource.Mega)),
-		Vcpus:       refInt(overrides.Limits.Cpu().ScaledValue(resource.Mega)),
+		Vcpus:       refInt(overrides.Limits.Cpu().Value()),
```
katrogan: can you add a comment why we scale one and not the other?

PR #62, opened 2020-03-03
Add missing sub-task log links

EngHabu:
# TL;DR
When a second (or further) attempts are launched for a subtask, the log link doesn't show. This is because the array of Attempts in the structure doesn't get populated with a new entry until the attempt completes.
This changes ensures we do not add duplicate log links because the log stream in the base container struct will always match the last known attempt's log stream

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?
 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Complete description
Always consider the "current" container job detail not just the first time.

## Tracking Issue
https://github.com/lyft/flyte/issues/217

## Follow-up issue
_NA_


go/tasks/plugins/array/awsbatch/jobs_store.go
```diff
@@ -147,8 +150,8 @@ func updateJob(ctx context.Context, source *batch.JobDetail, target *Job) (updat
 		target.Attempts = append(target.Attempts, a)
 	}
 
-	// If no job attempts are present, try to construct one from container status
-	if len(source.Attempts) == 0 {
+	// Add the "current" log stream to log links if one exists.
+	{
```
wild-endeavor: at a glance, i don't understand why this is in a separate block.  can you add a comment?

PR #65, opened 2020-03-07
Replace lower-cased sidecar gpu resource with nvidia one

katrogan:
Addresses https://github.com/lyft/flyte/issues/180 by substituting "gpu" from the sidecar custom resources with the fully nvdia gpu resource name.

When we build Flyte container tasks we never even use "GPU", but rather the full nvidia name (https://github.com/lyft/flytepropeller/blob/f487cfe6e4d35b52f6f9cb399cceb5abf4a0595a/pkg/utils/k8s.go#L58)

I believe only sidecar tasks are (incorrectly) using the "GPU" search string in flyteplugins

### DIFF
--- a/go/tasks/plugins/k8s/sidecar/testdata/sidecar_custom
+++ b/go/tasks/plugins/k8s/sidecar/testdata/sidecar_custom
@@ -27,6 +27,9 @@
 			}],
 			"resources": {
 				"requests": {
+				    "gpu": {
+				    	"string": "2"
+				    },
 					"cpu": {
 						"string": "10"
 					}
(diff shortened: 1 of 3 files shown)


katrogan:
> add a description and link to the other PRs?

@wild-endeavor done

go/tasks/plugins/k8s/sidecar/testdata/sidecar_custom
```diff
@@ -27,13 +27,16 @@
 			}],
 			"resources": {
 				"requests": {
+				    "gpu": {
```
wild-endeavor: so the point of this test is that regardless of whether you specify `gpu` or the nvidia gpu, both work?

PR #67, opened 2020-03-07
RawDataOutput directory for every task execution

kumare3:
# TL;DR
This PR enables FlytePropeller to create data sandboxes for the executors (containers or plugins) to use as data storage. 

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Currently Flyte allows exactly once processing semantics for processes that use flytekit data handling and do not produce any side effects. This is possible because flytekit creates a random data sandbox directory (in the blobstore) which is commited (i.e., commit here implies that a user can perceive a task to be complete only when the task completion is successfully recorded by flytepropeller, this is just a trick, concurrent executions are thus disregarded only the first one to return wins/commits).

This commit promotes this sandbox creation from flytekit into flytepropeller. this enables plugins that do not use flytekit or other sdks can simply rely on the {{ .outputs.sandbox }} directory parameter that is passed as an input. Every retry gets a new random sandbox. This also enables to mark some sandboxes are active and this making is possible to perform selective garbage collection.

## Tracking Issue
https://github.com/lyft/flyte/issues/195

## Follow-up issue
https://github.com/lyft/flyte/issues/211 - Per workflow data sandboxes


go/tasks/pluginmachinery/io/iface.go
```diff
@@ -41,7 +41,17 @@ type OutputReader interface {
...
 
-// All paths where various outputs produced by the task can be placed, such that the framework can directly access them.
+// Data sandbox is the actual path where the data produced by a task can be placed. It is completely optional. The advantage
+// of using this path is to provide exactly once semantics. It is guaranteed that this path is unique for every new execution
+// of a task (across retries etc) and is constant for a specific execution.
+// As of 02/20/2020 Flytekit generates this path randomly for S3. This structure proposes migration of this logic to
+// FlytePluginMachinery so that it can be used more universally outside of Flytekit.
+type OutputDataSandbox interface {
+	// This is prefix (blob store prefix or directory) where all data produced can be stored.
+	GetOutputDataSandboxPath() storage.DataReference
```
wild-endeavor: is sandbox the best name?  That implies something to do with the sandbox deployment i feel like.

kumare3: hmm true, it is really a sandbox. Do you have any suggestions?

kumare3: I like the name sandbox for this, but lets really think of a good alternative, then i can change. Otherwise lets stick with this. So i called it DataOutputSandbox

wild-endeavor: AtRestStorage, UserStorageLocation, FlyteHardDrive, FlyteBlackBox, FlyteRecorder, UserOutputLocation, UserOutputStorage, FlyteDurableStore, etc.  I don't really care that much, but i find some Flyte things already unnecessarily confusing and would just rather err on the side of less confusing.

EngHabu:
```suggestion
	GetOutputDataSandboxPrefix() storage.DataReference
```

go/tasks/pluginmachinery/io/iface.go
```diff
@@ -51,6 +61,8 @@ type OutputFilePaths interface {
 	// A Fully qualified path (URN) where the error information should be placed as a protobuf core.ErrorDocument. It is not directly
 	// used by the framework, but could be used in the future
 	GetErrorPath() storage.DataReference
+
+	OutputDataSandbox
```
EngHabu: nit: Can you move it to the top of the interface... I find it a bit more readable..

go/tasks/pluginmachinery/io/iface.go
```diff
@@ -41,7 +41,17 @@ type OutputReader interface {
 	Read(ctx context.Context) (*core.LiteralMap, *ExecutionError, error)
 }
 
-// All paths where various outputs produced by the task can be placed, such that the framework can directly access them.
+// Data sandbox is the actual path where the data produced by a task can be placed. It is completely optional. The advantage
+// of using this path is to provide exactly once semantics. It is guaranteed that this path is unique for every new execution
+// of a task (across retries etc) and is constant for a specific execution.
+// As of 02/20/2020 Flytekit generates this path randomly for S3. This structure proposes migration of this logic to
+// FlytePluginMachinery so that it can be used more universally outside of Flytekit.
+type OutputDataSandbox interface {
```
EngHabu: Do we really need an interface here? I feel like we should just move the method into OutputFilePaths interface... that or we keep it as a separate interface but do not include it in the other interface, include it directly in the implementation struct.

go/tasks/pluginmachinery/ioutils/output_sandbox.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (r precomputedOutputSandbox) GetOutputDataSandboxPath() storage.DataReference {
+	return r.path
+}
+
+// Creates a deterministic OutputSandbox whose path is distributed based on the ShardSelector passed in.
+// Determinism depends on the outputMetadataPath
+// Potential performance problem, as creating anew randomprefixShardedOutput Sandbox may be expensive as it hashes the outputMetadataPath
+// the final OutputSandbox is created in the shard selected by the sharder at the basePath and then appended by a hashed value of the outputMetadata
+func NewRandomPrefixShardedOutputSandbox(ctx context.Context, sharder ShardSelector, basePath, outputMetadataPath storage.DataReference, store storage.ReferenceConstructor) (io.OutputDataSandbox, error) {
```
EngHabu: This is not random prefix, right? this depends on the ShardSelector. This guy is pretty deterministic otherwise... right?

go/tasks/pluginmachinery/io/iface.go
kumare3: Sure

go/tasks/pluginmachinery/io/iface.go
kumare3: Should we change the name sandbox to something else. Let me think

go/tasks/pluginmachinery/ioutils/output_sandbox.go
kumare3: it is very deterministic by design, i should rename this constructor. Let me think about the names

go/tasks/pluginmachinery/io/iface.go
wild-endeavor: agreed.

go/tasks/pluginmachinery/ioutils/precomputed_shardselector.go
```diff
@@ -0,0 +1,71 @@
...
+			sb.WriteRune(c2)
+			precomputedPrefixes = append(precomputedPrefixes, sb.String())
+		}
+	}
+
+	return NewConstantShardSelector(precomputedPrefixes), nil
+}
+
+// uses the given shards to select a shard
+func NewConstantShardSelector(shards []string) ShardSelector {
```
wild-endeavor: I still prefer to accept interfaces and return specific types.

go/tasks/plugins/array/catalog.go
```diff
@@ -416,13 +422,19 @@ func ConstructOutputReaders(ctx context.Context, dataStore *storage.DataStore, o
 	return outputReaders, nil
 }
 
-func ConstructOutputReader(ctx context.Context, dataStore *storage.DataStore, outputPrefix storage.DataReference,
+func ConstructOutputReader(ctx context.Context, dataStore *storage.DataStore, outputPrefix, baseOutputSandbox storage.DataReference,
 	index int) (io.OutputReader, error) {
-	dataReference, err := dataStore.ConstructReference(ctx, outputPrefix, strconv.Itoa(index))
+	strIndex := strconv.Itoa(index)
+	dataReference, err := dataStore.ConstructReference(ctx, outputPrefix, strIndex)
```
wild-endeavor: i don't follow.  Why are there two data references now?  there's the `dataReference` that's in the original code, and the new `outputSandbox` - what's the difference between the two and why do we need both?

go/tasks/pluginmachinery/ioutils/remote_file_output_writer.go
```diff
@@ -71,10 +72,11 @@ func (w RemoteFileOutputWriter) Put(ctx context.Context, reader io.OutputReader)
 	return fmt.Errorf("no data found to write")
 }
 
-func NewRemoteFileOutputPaths(_ context.Context, store storage.ReferenceConstructor, outputPrefix storage.DataReference) RemoteFileOutputPaths {
```
wild-endeavor: given that now there's a new datareference as part of `io.RawOutputPaths` would it help to add some more functions to `RemoteFileOutputReader`?

go/tasks/pluginmachinery/ioutils/precomputed_shardselector.go
kumare3: you cannot, the linter wont let you for non exported types

go/tasks/plugins/array/catalog.go
kumare3: so the first reference is for where the metadata is stored -> OutputPath (should have been OutputMetadata). this where essentially where outputs.pb, futures.pb etc
the new one is what is today generated by flytekit, location where data is stored for the execution from the container. This is never read by flytepropeller, but it is better to generate it from propeller

go/tasks/pluginmachinery/ioutils/remote_file_output_writer.go
kumare3: Nope, we never want to read that data. Now that we call RawOutput that should be more clear.

go/tasks/pluginmachinery/ioutils/precomputed_shardselector.go
EngHabu: But PrecomputedShardSelector is exported

go/tasks/pluginmachinery/ioutils/raw_output_path.go
```diff
@@ -0,0 +1,66 @@
...
+	if err != nil {
+		return nil, err
+	}
+	return precomputedRawOutputPaths{
+		path: path,
+	}, nil
+}
+
+// A simple Output sandbox at a given path
+func NewRawOutputPath(_ context.Context, outputSandboxPath storage.DataReference) io.RawOutputPaths {
```
EngHabu:
```suggestion
func NewRawOutputPaths(_ context.Context, outputSandboxPath storage.DataReference) io.RawOutputPaths {
```

go/tasks/pluginmachinery/ioutils/raw_output_path.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (r precomputedRawOutputPaths) GetRawOutputPrefix() storage.DataReference {
+	return r.path
+}
+
+// Creates a deterministic RawOutputPath whose path is distributed based on the ShardSelector passed in.
+// Determinism depends on the outputMetadataPath
+// Potential performance problem, as creating a new RawPath creation may be expensive as it hashes the outputMetadataPath
+// the final RawOutputPath is created in the shard selected by the sharder at the basePath and then appended by a hashed value of the outputMetadata
+func NewShardedDeterministicRawOutputPath(ctx context.Context, sharder ShardSelector, basePath, outputMetadataPath storage.DataReference, store storage.ReferenceConstructor) (io.RawOutputPaths, error) {
```
EngHabu:
```suggestion
func NewShardedDeterministicRawOutputPath(ctx context.Context, sharder ShardSelector, basePrefix, outputMetadataPrefix storage.DataReference, store storage.ReferenceConstructor) (io.RawOutputPaths, error) {
```

go/tasks/pluginmachinery/ioutils/precomputed_shardselector.go
kumare3: good point i will unexport it

PR #69, opened 2020-03-10
Presto plugin executor

lu4nm3:
# TL;DR

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [X] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
This is the executor implementation to support a new Presto task

## Tracking Issue
https://github.com/lyft/flyte/issues/203

## Follow-up issue
_NA_
OR


go/tasks/plugins/presto/client/presto_client.go
```diff
@@ -0,0 +1,66 @@
+package client
```
lu4nm3: This is just a noop Presto client for the open-source version. There will be a separate commit for one based on Mozart

lu4nm3:
meow @wild-endeavor @EngHabu @varshaparthay

go/tasks/plugins/presto/execution_state.go
```diff
@@ -0,0 +1,528 @@
...
+		return "PhaseQueryFailed"
+	}
+	return "Bad Presto execution phase"
+}
+
+type ExecutionState struct {
+	Phase ExecutionPhase
+
+	// This will store the command ID from Presto
+	CommandID string `json:"command_id,omitempty"`
```
EngHabu: We are trying to standardize on camel casing for json tags.
```suggestion
	CommandID string `json:"commandId,omitempty"`
```

go/tasks/plugins/presto/execution_state.go
```diff
@@ -0,0 +1,528 @@
...
+		return "PhaseSubmitted"
+	case PhaseQuerySucceeded:
+		return "PhaseQuerySucceeded"
+	case PhaseQueryFailed:
+		return "PhaseQueryFailed"
+	}
+	return "Bad Presto execution phase"
+}
+
+type ExecutionState struct {
```
EngHabu: nit: docs on public types

go/tasks/plugins/presto/execution_state.go
```diff
@@ -0,0 +1,528 @@
...
+func generateRandomBytes(length int) ([]byte, error) {
+	b := make([]byte, length)
+	_, err := rand.Read(b)
+	// Note that err == nil only if we read len(b) bytes.
+	if err != nil {
+		return nil, err
+	}
+
+	return b, nil
+}
```
EngHabu: You can substitute this block with `rand.String(length)` the import is coming from `k8s/apimachinary` if I recall correctly.

go/tasks/plugins/presto/execution_state.go
```diff
@@ -0,0 +1,528 @@
...
+
+func getExternalLocation(shardFormatter string, shardLength int) string {
```
EngHabu: Can you move the whole sharding thing outside of the presto plugin? maybe in some storage util in plugins/core ?

go/tasks/plugins/presto/executions_cache.go
```diff
@@ -0,0 +1,166 @@
...
+	prestoClient svc.ServiceClient,
+	cfg *config.Config,
+	scope promutils.Scope) (ExecutionsCache, error) {
+
+	q := ExecutionsCache{
+		prestoClient: prestoClient,
+		scope:        scope,
+		cfg:          cfg,
+	}
+	autoRefreshCache, err := cache.NewAutoRefreshCache("presto", q.SyncPrestoQuery, workqueue.DefaultControllerRateLimiter(), ResyncDuration, cfg.Workers, cfg.LruCacheSize, scope)
```
EngHabu: You should probably add rate limiter params to the config and construct a rate limiter here with those params...

go/tasks/plugins/presto/test_helpers.go
```diff
@@ -0,0 +1,112 @@
+package presto
```
EngHabu: Can you rename the file to `helpers_test.go` or add the annotation to only build it for Test? I know go is pretty good at trimming what's not used but why put strain on the compiler if we don't need to..

go/tasks/plugins/svc/service_client.go
```diff
@@ -0,0 +1,15 @@
...
+
+import (
+	"context"
+)
+
+type CommandStatus string
+
+//go:generate mockery -all -case=snake
+
+type ServiceClient interface {
```
EngHabu: docs

go/tasks/plugins/svc/service_client.go
```diff
@@ -0,0 +1,15 @@
+package svc
```
EngHabu: If we are not doing service plugin implementation for now, let's move this inside Presto... I think this interface needs some rework to make it generic....  @wild-endeavor what do you think?

go/tasks/plugins/presto/client/presto_client.go
```diff
@@ -0,0 +1,66 @@
...
+	"github.com/lyft/flyteplugins/go/tasks/plugins/svc"
+
+	"time"
+
+	"github.com/lyft/flyteplugins/go/tasks/plugins/presto/config"
+)
+
+const (
+	httpRequestTimeoutSecs = 30
+	//AcceptHeaderKey          = "Accept"
```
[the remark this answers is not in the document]
lu4nm3: Completely missed this, will do!

go/tasks/plugins/presto/client/presto_client.go
```diff
@@ -0,0 +1,66 @@
...
+	Schema       string `json:"schema,omitempty"`
+	Source       string `json:"source,omitempty"`
+}
+type PrestoExecuteResponse struct {
+	ID      string
+	Status  svc.CommandStatus
+	NextURI string
+}
+
+func (p *prestoClient) ExecuteCommand(
```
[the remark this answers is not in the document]
lu4nm3: Interesting. I was following what the Hive client was doing here. I'm not too familiar with things in Go yet but, when would I use a reference vs a value for a method?

go/tasks/plugins/presto/client/presto_client.go
```diff
@@ -0,0 +1,66 @@
...
+
+func (p *prestoClient) KillCommand(ctx context.Context, commandID string) error {
+	return nil
+}
+
+func (p *prestoClient) GetCommandStatus(ctx context.Context, commandID string) (svc.CommandStatus, error) {
+	return NewPrestoStatus(ctx, "UNKNOWN"), nil
+}
+
+func NewPrestoClient(cfg *config.Config) svc.ServiceClient {
```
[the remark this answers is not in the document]
lu4nm3: Oh, that's interesting. Normally from my experience in other langs (java, scala, rust, etc), you typically tie yourself to an interface instead of a specific implementation. In our case, I feel this will help when trying to swap out the Mozart client for one based on just Presto. Otherwise, won't we need to  make a copy  of the calling  code (ie. the executor) in the  private repo?

go/tasks/plugins/svc/service_client.go
lu4nm3: Sorry, can you clarify what you mean by docs? Do you mean to add comments to this file which will serve as docs? Or are there other docs I should be updating?

go/tasks/plugins/svc/service_client.go
lu4nm3: I think @wild-endeavor suggested this yesterday. I'm fine moving it.

go/tasks/plugins/presto/config/config.go
```diff
@@ -0,0 +1,67 @@
...
+	prestoConfigSection = pluginsConfig.MustRegisterSubSection(prestoConfigSectionKey, &defaultConfig)
+)
+
+// Presto plugin configs
+type Config struct {
+	Environment         config.URL           `json:"endpoint" pflag:",Endpoint for Presto to use"`
+	DefaultRoutingGroup string               `json:"defaultRoutingGroup" pflag:",Default Presto routing group"`
+	Workers             int                  `json:"workers" pflag:",Number of parallel workers to refresh the cache"`
+	LruCacheSize        int                  `json:"lruCacheSize" pflag:",Size of the AutoRefreshCache"`
+	AwsS3ShardFormatter string               `json:"awsS3ShardFormatter" pflag:", S3 bucket prefix where Presto results will be stored"`
```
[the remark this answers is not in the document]
lu4nm3: In this case, we would be limited by whatever Presto supports in terms of external tables. I'm not very familiar with Flyte's generic storage model. Do you have any ideas of how I might abstract this? Currently, this logic of generating random paths is handled by flytekit but I'm not sure if the logic there  is very generic as it appeared to be tied to S3.

go/tasks/plugins/presto/config/config.go
```diff
@@ -0,0 +1,67 @@
...
+type RoutingGroupConfig struct {
+	Name                             string  `json:"primaryLabel" pflag:",The name of a given Presto routing group"`
+	Limit                            int     `json:"limit" pflag:",Resource quota (in the number of outstanding requests) of the routing group"`
+	ProjectScopeQuotaProportionCap   float64 `json:"projectScopeQuotaProportionCap" pflag:",A floating point number between 0 and 1, specifying the maximum proportion of quotas allowed to allocate to a project in the routing group"`
+	NamespaceScopeQuotaProportionCap float64 `json:"namespaceScopeQuotaProportionCap" pflag:",A floating point number between 0 and 1, specifying the maximum proportion of quotas allowed to allocate to a namespace in the routing group"`
+}
+
+var (
+	defaultConfig = Config{
+		Environment:         URLMustParse("https://prestoproxy-internal.lyft.net:443"),
```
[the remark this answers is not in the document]
lu4nm3: @EngHabu I had a larger question with regards to configs, namely, where would I create the actual config file for the Presto plugin? I couldn't find any in this repo

go/tasks/plugins/presto/client/presto_status.go
```diff
@@ -0,0 +1,41 @@
...
+const (
+	PrestoStatusUnknown   svc.CommandStatus = "UNKNOWN"
+	PrestoStatusQueued    svc.CommandStatus = "QUEUED"
+	PrestoStatusRunning   svc.CommandStatus = "RUNNING"
+	PrestoStatusFinished  svc.CommandStatus = "FINISHED"
+	PrestoStatusFailed    svc.CommandStatus = "FAILED"
+	PrestoStatusCancelled svc.CommandStatus = "CANCELLED"
+)
+
+var PrestoStatuses = map[svc.CommandStatus]struct{}{
```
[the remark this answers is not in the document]
lu4nm3: When I do this:
```
var PrestoStatuses = sets.String{
	PrestoStatusUnknown:   {},
	PrestoStatusQueued:    {},
	PrestoStatusRunning:   {},
	PrestoStatusFinished:  {},
	PrestoStatusFailed:    {},
	PrestoStatusCancelled: {},
}
```
I get the error messages about:
```
Cannot use 'PrestoStatusUnknown' (type svc.CommandStatus) as type string
```
Am I doing this correctly?

go/tasks/plugins/presto/execution_state.go
lu4nm3: Hmm, if this is the case, then it appears it would be inconsistent with what you guys have for the Hive  plugin: https://github.com/lyft/flyteplugins/blob/master/go/tasks/plugins/hive/execution_state.go#L54
Do you guys prefer camelCase or snake_case?

go/tasks/plugins/presto/execution_state.go
lu4nm3: Where are docs located?

go/tasks/plugins/presto/client/presto_client.go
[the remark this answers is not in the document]
wild-endeavor: The Hive code should probably be changed.  by value here is fine, esp since the underlying http client is already a pointer.

go/tasks/plugins/presto/config/config.go
```diff
@@ -0,0 +1,67 @@
...
+		DefaultRoutingGroup: "adhoc",
+		Workers:             15,
+		LruCacheSize:        2000,
+		AwsS3ShardFormatter: "s3://lyft-modelbuilder/{}/",
```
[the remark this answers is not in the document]
wild-endeavor: Leave it out, but we should put in canonical examples that work with the sandbox (docker desktop/minikube) deployment.

go/tasks/plugins/presto/config/config.go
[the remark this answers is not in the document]
wild-endeavor: Here: https://github.com/lyft/flytepropeller-private/tree/master/artifacts/overlays/production/propeller/plugins. (i think a private link is okay).

go/tasks/plugins/presto/execution_state.go
wild-endeavor: camel I'm pretty sure.

go/tasks/plugins/presto/config/config.go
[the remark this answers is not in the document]
EngHabu: [Qubole Plugin Config for OSS](https://github.com/lyft/flyte/blob/master/kustomize/overlays/sandbox/propeller/plugins/qubole/config.yaml)

[Mozart Plugin Config For Lyft](https://github.com/lyft/flytepropeller-private/blob/master/artifacts/overlays/production/propeller/plugins/mozart/config.yaml)

go/tasks/plugins/presto/execution_state.go
```diff
@@ -0,0 +1,528 @@
...
+	if !ok {
+		logger.Errorf(ctx, "Error casting cache object into ExecutionState")
+		return currentState, errors.Errorf(errors.CacheFailed, "Failed to cast [%v]", cachedItem)
+	}
+
+	// If there were updates made to the state, we'll have picked them up automatically. Nothing more to do.
+	return cachedExecutionState.ExecutionState, nil
+}
+
+func MapExecutionStateToPhaseInfo(state ExecutionState) core.PhaseInfo {
```
wild-endeavor: does this function still work?  I feel like we need to add more phase versions.

The Hive state machine is a one pass waterfall.  it just goes through each state linearly.  This state machine goes back and revisits the states because there's the query count.  We should make each time around a new phase version.

go/tasks/plugins/presto/client/presto_status.go
```diff
@@ -0,0 +1,41 @@
...
+	PrestoStatusQueued:    {},
+	PrestoStatusRunning:   {},
+	PrestoStatusFinished:  {},
+	PrestoStatusFailed:    {},
+	PrestoStatusCancelled: {},
+}
+
+func NewPrestoStatus(ctx context.Context, state string) svc.CommandStatus {
+	upperCased := strings.ToUpper(state)
+	if strings.Contains(upperCased, "FAILED") {
```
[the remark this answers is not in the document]
lu4nm3: Not entirely sure, I sort of copied this from the existing qubole_status. Probably not necessary to single this out.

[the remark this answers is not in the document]
lu4nm3: Actually, I think it was needed because Hive/Qubole has different error/failure modes and I think this maps them all to a single Failure/Error. Going  to leave this in for now

go/tasks/plugins/presto/config/config.go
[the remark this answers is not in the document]
lu4nm3: Thanks! I'll create separate PRs to add configs to these repos

go/tasks/plugins/presto/execution_state.go
lu4nm3: That method seems to generate a random string without vowels, I'm guessing that's ok?

go/tasks/plugins/presto/execution_state.go
lu4nm3: This sharding logic currently only exists in flytekit for Hive I believe. Since currently, only the Presto plugin uses this, can we keep it together with the rest of the Presto logic?

go/tasks/plugins/presto/client/presto_status.go
```diff
@@ -0,0 +1,41 @@
...
+	"strings"
+
+	"github.com/lyft/flyteplugins/go/tasks/plugins/svc"
+	"github.com/lyft/flytestdlib/logger"
+)
+
+// This type is meant only to encapsulate the response coming from Presto as a type, it is
+// not meant to be stored locally.
+const (
+	PrestoStatusUnknown   svc.CommandStatus = "UNKNOWN"
```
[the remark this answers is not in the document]
lu4nm3: I tried this but ran into a problem:
```
enumer: can't handle non-integer constant type CommandStatus
go/tasks/plugins/presto/client/presto_status.go:10: running "enumer": exit status 1
```
I think this might be why the `qubole_status.go` doesn't use the generator.

go/tasks/pluginmachinery/core/phase.go
```diff
@@ -180,6 +180,10 @@ func PhaseInfoSuccess(info *TaskInfo) PhaseInfo {
 	return phaseInfo(PhaseSuccess, DefaultPhaseVersion, nil, info)
 }
 
+func PhaseInfoSuccessWithVersion(version uint32, info *TaskInfo) PhaseInfo {
```
EngHabu: I don't think we consider the version on success.. success phase is terminal, so once you hit that, there should be no expectation the system will call your plugin ever again.

go/tasks/plugins/presto/client/presto_client.go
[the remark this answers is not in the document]
EngHabu: Well, interfaces in Go are kind of the inverse in Java.
The idiomatic way is something like this:
`func MyFunc(m SpecificInterface) SpecificType`
So your function takes an interface of exactly the functions it needs and returns a specific type..

I do not understand what do you mean by copying calling code?

go/tasks/plugins/svc/service_client.go
EngHabu: ah I meant godocs so yeah comments on the public types, fields and funcs... ideally also a package level comment to explain what this is

go/tasks/pluginmachinery/core/phase.go
lu4nm3: Hmm, so the reason why I wanted to add a version to success is that this is a status that will occur up to 5 times for every Presto query due to how the mini-state machine is implemented where it executes a total of 5 Presto queries for each Presto task. After each  of these 5 queries finishes successfully, its status will be  successful and is handled here: https://github.com/lyft/flyteplugins/pull/69/files?file-filters%5B%5D=.go#diff-6a65d420579b33b62766e71e98eafdf7R112

go/tasks/plugins/presto/client/presto_client.go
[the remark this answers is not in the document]
lu4nm3: So given that I made the client specific to Presto, my original comment is largely void now. 

What I was trying to say is that if we made the function return a specific type:
```
func NewNoopPrestoClient(cfg *config.Config) PrestoClient {
```

Then in the calling code that expects  this client, we would also need to specifically have it accept a `PrestoClient`:
```
func InitializePrestoExecutor(
	ctx context.Context,
	iCtx core.SetupContext,
	cfg *config.Config,
	resourceConfig map[string]int,
	prestoClient client.PrestoClient) (core.Plugin, error) {
```
Normally, I think this would be ok. But what I was trying to say (perhaps in a not so clear manner), is that if one day we wanted to abstract the client to be a more generic `ServiceClient` to use with not only Presto but also Hive, etc, then it's better if we rely on this interface (`ServiceClient`) as it would give us the flexibility on swapping out the client as we see fit. However, as I'm writing this, I realize that making this abstract is probably a larger refactoring effort that will likely change even this disregard this explanation haha. Sorry I think I'm rambling at this point

go/tasks/pluginmachinery/core/phase.go
wild-endeavor: As each query finishes, even though the internal state might move to `PhaseQuerySucceeded`, the overall phase that's sent to Admin should still be running.  Only when everything is truly done, should the plugin send a success.  Does this make sense?

lu4nm3: I see. So we should keep things as  Running and only use Success once at the very end once everything is done?

lu4nm3: @wild-endeavor I went ahead and updated things so that the success phase is only returned at the very end once all 5 queries have executed

go/tasks/plugins/presto/client/presto_status.go
```diff
@@ -0,0 +1,42 @@
...
+const (
+	PrestoStatusUnknown   PrestoStatus = "UNKNOWN"
+	PrestoStatusQueued    PrestoStatus = "QUEUED"
+	PrestoStatusRunning   PrestoStatus = "RUNNING"
+	PrestoStatusFinished  PrestoStatus = "FINISHED"
+	PrestoStatusFailed    PrestoStatus = "FAILED"
+	PrestoStatusCancelled PrestoStatus = "CANCELLED"
+)
+
+var PrestoStatuses = map[PrestoStatus]struct{}{
```
EngHabu: I could swear I left this feedback but I can't see it... 
Can you see if you can use this [instead](https://github.com/lyft/flyteplugins/blob/dc076e467ed5fbf34fce3506db676a8556376463/go/tasks/pluginmachinery/core/transition.go#L5).

lu4nm3: @EngHabu I think  this was your comment: https://github.com/lyft/flyteplugins/pull/69#discussion_r391834458
I replied saying that it doesn't seem like it's supported for strings.

go/tasks/plugins/presto/client/presto_client.go
```diff
@@ -0,0 +1,63 @@
+package client
```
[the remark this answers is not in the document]
lu4nm3: Sounds good, I'll rename to `noop_presto_client.go`

go/tasks/plugins/presto/execution_state.go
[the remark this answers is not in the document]
wild-endeavor: I feel like writing it like this and abstracting later is easier.

go/tasks/plugins/presto/execution_state.go
```diff
@@ -0,0 +1,569 @@
...
+	catalog = prestoQuery.Catalog
+	schema = prestoQuery.Schema
+	statement = prestoQuery.Statement
+
+	inputs, err := tCtx.InputReader().Get(ctx)
+	if err != nil {
+		return "", "", "", "", err
+	}
+
+	statement, routingGroup, catalog, schema, err = presto.InterpolateInputs(ctx, *inputs, statement, routingGroup, catalog, schema)
```
lu4nm3: We interpolate the `statement`, `routing_config`, `catalog`, `schema` with inputs from users. See the comments I wrote in the utils file. I tested this locally and it works

go/tasks/plugins/presto/execution_state.go
```diff
@@ -0,0 +1,566 @@
...
+	catalog := prestoQuery.Catalog
+	schema := prestoQuery.Schema
+	statement := prestoQuery.Statement
+
+	inputs, err := tCtx.InputReader().Get(ctx)
+	if err != nil {
+		return "", "", "", "", err
+	}
+
+	routingGroup, catalog, schema, statement, err = presto.InterpolateInputs(ctx, *inputs, routingGroup, catalog, schema, statement)
```
lu4nm3: This is where  user inputs get interpolated for the query, routing group, catalog, and  schema.  See comments  in the interpolator file for more details

go/tasks/plugins/presto/client/presto_status.go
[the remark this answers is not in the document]
wild-endeavor: Add this as a comment?

go/tasks/plugins/presto/config/config.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+type RoutingGroupConfig struct {
+	Name                             string  `json:"name" pflag:",The name of a given Presto routing group"`
+	Limit                            int     `json:"limit" pflag:",Resource quota (in the number of outstanding requests) of the routing group"`
+	ProjectScopeQuotaProportionCap   float64 `json:"projectScopeQuotaProportionCap" pflag:",A floating point number between 0 and 1, specifying the maximum proportion of quotas allowed to allocate to a project in the routing group"`
+	NamespaceScopeQuotaProportionCap float64 `json:"namespaceScopeQuotaProportionCap" pflag:",A floating point number between 0 and 1, specifying the maximum proportion of quotas allowed to allocate to a namespace in the routing group"`
+}
+
+type RateLimiter struct {
```
wild-endeavor: In light of the other conversations we've had, should we rename this struct?

lu4nm3: Sure. How does  `CacheConfig` sound?

go/tasks/plugins/presto/utils/input_interpolator.go
```diff
@@ -0,0 +1,137 @@
...
+		return "", "", "", "", err
+	}
+
+	// Remove implicit inputs from the rest of the inputs used for interpolation
+	delete(inputsAsStrings, "__implicit_routing_group")
+	delete(inputsAsStrings, "__implicit_catalog")
+	delete(inputsAsStrings, "__implicit_schema")
+
+	statement = interpolate(inputsAsStrings, statement)
+	routingGroup = interpolate(inputsAsStrings, routingGroup)
```
wild-endeavor: actually, @EngHabu can you chime in here?  We only want to do statement right?  Everything else we just leave as is?

go/tasks/plugins/presto/executor_metrics.go
```diff
@@ -0,0 +1,34 @@
...
+func getPrestoExecutorMetrics(scope promutils.Scope) ExecutorMetrics {
+	return ExecutorMetrics{
+		Scope: scope,
+		ReleaseResourceFailed: labeled.NewCounter("presto_released_resource_failed",
+			"Error releasing allocation token for Presto", scope),
+		AllocationGranted: labeled.NewCounter("presto_allocation_granted",
+			"Allocation request granted for Presto", scope),
+		AllocationNotGranted: labeled.NewCounter("presto_allocation_not_granted",
+			"Allocation request did not fail but not granted for Presto", scope),
+		ResourceWaitTime: scope.MustNewSummaryWithOptions("presto_resource_wait_time",
```
wild-endeavor: Do we want to add "Seconds"/"_seconds" to this metric name?  Not sure if we have a convention.  Haytham?

lu4nm3: Not sure. I copied this largely from the  Hive version. Not sure how units were handled in that case.

go/tasks/plugins/presto/utils/input_interpolator.go
lu4nm3: Well, this approach would provide a bit more flexibility to users in terms of how much they are able to customize their Presto tasks. For example, if they want  their query to  only include  the table  name:
```
SELECT * FROM table_name
```
Then they could set the `catalog`  and  `schema` as inputs whose values perhaps come from a previous task and thus you wouldn't limit users to only be able to interpolate the query statement itself. Not sure  if interpolating only  the query will be enough for users. If so, then I can definitely simplify this. Your call.

go/tasks/plugins/presto/utils/input_interpolator.go
```diff
@@ -0,0 +1,137 @@
...
+	"regexp"
+	"strings"
+
+	"github.com/golang/protobuf/ptypes"
+	pb "github.com/lyft/flyteidl/gen/pb-go/flyteidl/core"
+	"github.com/lyft/flytestdlib/logger"
+)
+
+// Matches any pair of open/close mustaches that contains a variable inside (e.g. `{{ abc }}`)
+var inputMustacheRegex = regexp.MustCompile(`{{\s*[^\s]+\s*}}`)
```
EngHabu: Let's not have our own matching and our own format here... please use the one that already exist [here](https://github.com/lyft/flyteplugins/blob/dc076e467ed5fbf34fce3506db676a8556376463/go/tasks/pluginmachinery/utils/template.go#L34).

go/tasks/plugins/presto/client/presto_client.go
```diff
@@ -0,0 +1,35 @@
...
+type PrestoExecuteArgs struct {
+	RoutingGroup string `json:"routingGroup,omitempty"`
+	Catalog      string `json:"catalog,omitempty"`
+	Schema       string `json:"schema,omitempty"`
+	Source       string `json:"source,omitempty"`
+	User         string `json:"user,omitempty"`
+}
+
+// Representation of a response after submitting a query to Presto
+type PrestoExecuteResponse struct {
```
EngHabu: nit: add json tags

go/tasks/plugins/presto/client/presto_status.go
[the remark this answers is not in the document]
EngHabu: Here is an example that works:
https://play.golang.org/p/DOSciZ7joY1

I needed to implement UnmarshalJSON()...

go/tasks/plugins/presto/config/config.go
```diff
@@ -0,0 +1,78 @@
...
+
+	"github.com/lyft/flytestdlib/config"
+	"github.com/lyft/flytestdlib/logger"
+
+	pluginsConfig "github.com/lyft/flyteplugins/go/tasks/config"
+)
+
+const prestoConfigSectionKey = "presto"
+
+func URLMustParse(s string) config.URL {
```
EngHabu: lower case? 
```suggestion
func urlMustParse(s string) config.URL {
```

go/tasks/plugins/presto/config/config.go
```diff
@@ -0,0 +1,78 @@
...
+	ProjectScopeQuotaProportionCap   float64 `json:"projectScopeQuotaProportionCap" pflag:",A floating point number between 0 and 1, specifying the maximum proportion of quotas allowed to allocate to a project in the routing group"`
+	NamespaceScopeQuotaProportionCap float64 `json:"namespaceScopeQuotaProportionCap" pflag:",A floating point number between 0 and 1, specifying the maximum proportion of quotas allowed to allocate to a namespace in the routing group"`
+}
+
+type RateLimiter struct {
+	Name         string          `json:"name" pflag:",The name of the rate limiter"`
+	SyncPeriod   config.Duration `json:"syncPeriod" pflag:",The duration to wait before the cache is refreshed again"`
+	Workers      int             `json:"workers" pflag:",Number of parallel workers to refresh the cache"`
+	LruCacheSize int             `json:"lruCacheSize" pflag:",Size of the cache"`
+	MetricScope  string          `json:"metricScope" pflag:",The prefix in Prometheus used to track metrics related to Presto"`
```
EngHabu: I don't think you need this... you just create it as a const in your code

go/tasks/plugins/presto/config/config.go
```diff
@@ -0,0 +1,78 @@
...
+}
+
+var (
+	defaultConfig = Config{
+		Environment:         URLMustParse(""),
+		DefaultRoutingGroup: "adhoc",
+		DefaultUser:         "flyte-default-user@lyft.com",
```
EngHabu: We should, generally, only fill in defaults with values that give you a good default behavior. This user, even if it works in Lyft, won't work outside...

go/tasks/plugins/presto/config/config.go
```diff
@@ -0,0 +1,78 @@
...
+		RoutingGroupConfigs: []RoutingGroupConfig{{Name: "adhoc", Limit: 250}, {Name: "etl", Limit: 100}},
+		RateLimiter: RateLimiter{
+			Name:         "presto",
+			SyncPeriod:   config.Duration{Duration: 5 * time.Second},
+			Workers:      15,
+			LruCacheSize: 2000,
```
EngHabu:
```suggestion
			LruCacheSize: 10000,
```

go/tasks/plugins/presto/executor_metrics.go
EngHabu: It doesn't look like we are using them though...

lu4nm3: I believe it's used somewhere in the executor

go/tasks/plugins/presto/utils/input_interpolator.go
lu4nm3: @wild-endeavor and I discussed this last week and came to the conclusion that we can't just use this out-of-the-box as inputs in the Presto task (as well as how interpolation is  done on the query plus routing_group, catalog, and schema) are being handled a little differently

go/tasks/plugins/presto/config/config.go
lu4nm3: This is part of the change for one of your earlier comments where we should make the cache/rate limiter configurable. Are there things that should be  configurable and others that shouldnt?

go/tasks/plugins/presto/config/config.go
lu4nm3: I made it  public because I call the method in one of the unit test

wild-endeavor: I like URL.

go/tasks/plugins/presto/config/config.go
wild-endeavor: sure

go/tasks/plugins/presto/client/presto_status.go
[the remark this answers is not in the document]
lu4nm3: @EngHabu Sorry, I'm still a little confused as I think this example you posted  uses:
```
type TransitionType int
```
where the type is `int`. In my cause, the status is a `string` and based on my attempts above, it doesn't seem to work on non-integer types. Am I missing something?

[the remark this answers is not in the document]
EngHabu: Does it need to be a string though?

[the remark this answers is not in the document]
EngHabu: Ok, this is not a big deal... if it doesn't seem obvious to change, that's fine...

go/tasks/plugins/presto/config/config.go
EngHabu: Yeah, I would say you should drop Name and MetricScope from here...

lu4nm3:
Ready for another  look 🙏

PR #71, opened 2020-03-11
Enable Resource Manager for k8s array plugin

migueltol22:
# TL;DR
Use resource manager for k8s array plugin as a rate limiter to k8s.

Example Config that could be used for resource manager:
```
resourceConfig:
  - primaryLabel: "tokens"
  - limit: 300
```

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/lyft/flyte/issues/201

## Follow-up issue
_NA_
OR


kumare3:
@DouglasCurbelo / @migueltol22 can we please create a issue for this and add it to the description

go/tasks/plugins/array/k8s/monitor.go
```diff
@@ -78,6 +80,15 @@ func CheckSubTasksState(ctx context.Context, tCtx core.TaskExecutionContext, kub
 
 		actualPhase := phaseInfo.Phase()
 		if phaseInfo.Phase().IsSuccess() {
+
+			// Release token
+			resourceNamespace := core.ResourceNamespace(tCtx.TaskExecutionMetadata().GetOwnerID().Namespace)
```
kumare3: release has to be done in case of success/failure or aborts right? I would do this in Finalize() as no matter what once you want to finish the task, we invoke finalize and thus this would be the right place to do token release.

kumare3: Also writing a simplified interface that launches, monitors, aborts and finalizes one job might be  nicer way of arranging the code - wdyt?

go/tasks/plugins/array/k8s/monitor.go
```diff
@@ -78,6 +80,15 @@ func CheckSubTasksState(ctx context.Context, tCtx core.TaskExecutionContext, kub
 
 		actualPhase := phaseInfo.Phase()
 		if phaseInfo.Phase().IsSuccess() {
```
DouglasCurbelo: What happen if the pod fail? Do we release the token as well?

go/tasks/plugins/array/k8s/launcher.go
```diff
@@ -94,6 +96,16 @@ func LaunchSubTasks(ctx context.Context, tCtx core.TaskExecutionContext, kubeCli
 
 		pod = ApplyPodPolicies(ctx, config, pod)
 
+		// Allocate Token
+		resourceNamespace := core.ResourceNamespace(pod.Namespace)
+		allocationStatus, err := tCtx.ResourceManager().AllocateResource(ctx, resourceNamespace, pod.Name, core.ResourceConstraintsSpec{})
```
DouglasCurbelo: Nob question, here we allocate resources using pod.Namespace, by on the initialization step, we :
```
iCtx.ResourceRegistrar().RegisterResourceQuota(ctx, core.ResourceNamespace(primaryLabel), tokenLimit)
```

Does that mean that we only can request tokens with the same label we used during registration? If so, does enforcement of limits per namespace would require registering quotas for each namespace on initialization?
@kumare3 ?

DouglasCurbelo: https://github.com/lyft/flyteplugins/blob/master/go/tasks/plugins/hive/executor.go#L127?

DouglasCurbelo:
@kumare3 : https://github.com/lyft/flyte/issues/201

go/tasks/plugins/array/k8s/monitor.go
migueltol22: So my question with doing this in finalize is that currently finalize will be called after all of the subtasks get run. What if we have 1000 subtasks and only 300 resources in that case we'll never release a resource (assuming my understanding is correct). I do agree that this should be done where each subtask has its own launch, monitor, etc. Any ideas on best way to handle this?  Would introducing a new phase such as `processingChildren` be a good approach?

go/tasks/plugins/array/k8s/launcher.go
migueltol22: I also have the same question about how resource namespace and requests are used

bnsblue: I don't think I understand your question, but let me provide some details regarding RM to see if that resolves your concern.
 
RM is basically a pooling system for plugins. A plugin requests to create one or more pools during setup time (using `ResourceRegistrar().RegisterResourceQuota()`). At the end of the setup time, the RM will be created based on the valid pool-creation requests, and the actual pools will be created. During execution time, the plugin send token-allocate and token-release requests to RM. Note that the term`ResourceNamespace` here has nothing to do with the users' namespace or k8s namespace (i.e. `<project>-<domain>`). `ResourceNamespace` here just means the name of a token pool, nothing else. So how you want to use it is totally up to you. 

I just realized that the argument names used in the function signature are misleading in this case. Those names made sense when I first wrote RM, but it seems like it has become confusing for its current form. I'll try to refactor the naming soon. Sorry for the confusion.

go/tasks/plugins/array/k8s/monitor.go
migueltol22: @kumare3 @EngHabu any thoughts on comment above?

go/tasks/plugins/array/k8s/config.go
```diff
@@ -32,13 +32,19 @@ var (
...
+}
+
 // Defines custom config for K8s Array plugin
 type Config struct {
 	DefaultScheduler     string `json:"scheduler" pflag:",Decides the scheduler to use when launching array-pods."`
 	MaxErrorStringLength int    `json:"maxErrLength" pflag:",Determines the maximum length of the error string returned for the array."`
 	MaxArrayJobSize      int64  `json:"maxArrayJobSize" pflag:",Maximum size of array job."`
 	OutputAssembler      workqueue.Config
 	ErrorAssembler       workqueue.Config
+	TokenConfigs         TokenConfig
```
EngHabu: Maybe ResourcesConfig?

go/tasks/plugins/array/k8s/config.go
```diff
@@ -32,13 +32,19 @@ var (
 	configSection = config.MustRegisterSection(configSectionKey, defaultConfig)
 )
 
+type TokenConfig struct {
+	primaryLabel string
```
EngHabu: These will need to be exported e.g.:
```suggestion
	PrimaryLabel string `json:"primaryLabel" pflag:",What is this?"`
```

go/tasks/plugins/array/k8s/monitor.go
```diff
@@ -24,73 +23,100 @@ import (
...
 	newState = currentState
-
 	msg := errorcollector.NewErrorMessageCollector()
-	newArrayStatus := arraystatus.ArrayStatus{
+	newArrayStatus := &arraystatus.ArrayStatus{
 		Summary:  arraystatus.ArraySummary{},
 		Detailed: arrayCore.NewPhasesCompactArray(uint(currentState.GetExecutionArraySize())),
 	}
 
+	if int64(currentState.GetExecutionArraySize()) > config.MaxArrayJobSize {
```
katrogan, line 48: nit: why not do this check before declaring all the variables above?

go/tasks/plugins/array/k8s/monitor.go
```diff
@@ -24,73 +23,100 @@ import (
...
+			ChildIdx:         childIdx,
+			MessageCollector: &msg,
 		}
 
-		if phaseInfo.Info() != nil {
-			logLinks = append(logLinks, phaseInfo.Info().Logs...)
+		// The first time we enter this state we will launch every subtask. On subsequent rounds, the pod
+		// has already been created so we return a Success value and continue with the Monitor step.
+		var status TaskStatus
+		status, err = task.Launch(ctx, tCtx, kubeClient)
```
katrogan, line 94: sorry if i'm missing something obvious here, but if the task is not terminal we always launch? what if it's already running?

go/tasks/plugins/array/k8s/task.go
```diff
@@ -0,0 +1,232 @@
...
+	MessageCollector *errorcollector.ErrorMessageCollector
+}
+
+type TaskStatus int8
+
+const (
+	Success TaskStatus = iota
+	Error
+	Waiting
+	ReturnState
```
katrogan, line 40: it's not immediately clear what this TaskStatus represents. can you add a comment explaining 'ReturnState'?

go/tasks/plugins/array/k8s/task.go
```diff
@@ -0,0 +1,232 @@
...
+
+func (t Task) Launch(ctx context.Context, tCtx core.TaskExecutionContext, kubeClient core.KubeClient) (TaskStatus, error) {
+	podTemplate, _, err := FlyteArrayJobToK8sPodTemplate(ctx, tCtx)
+	if err != nil {
+		return Error, errors2.Wrapf(ErrBuildPodTemplate, err, "Failed to convert task template to a pod template for a task")
+	}
+
+	var args []string
+	if len(podTemplate.Spec.Containers) > 0 {
+		args = append(podTemplate.Spec.Containers[0].Command, podTemplate.Spec.Containers[0].Args...)
```
katrogan, line 51: do we always assume that there is at most one container in the podspec? should we error if there are more? 0? in line 60 and elsewhere you reference the Containers[0] again without this check

go/tasks/plugins/array/k8s/task.go
```diff
@@ -0,0 +1,232 @@
...
+		}
+	}
+
+	t.NewArrayStatus.Detailed.SetItem(t.ChildIdx, bitarray.Item(actualPhase))
+	t.NewArrayStatus.Summary.Inc(actualPhase)
+
+	return Success, nil
+}
+
+func (t Task) Abort() {}
```
katrogan, line 138: why is this a no-op?

go/tasks/plugins/array/k8s/task.go
```diff
@@ -0,0 +1,232 @@
...
+		},
+		ObjectMeta: metav1.ObjectMeta{
+			Name:      podName,
+			Namespace: tCtx.TaskExecutionMetadata().GetNamespace(),
+		},
+	}
+
+	err := kubeClient.GetClient().Delete(ctx, pod)
+	if err != nil {
+		if k8serrors.IsNotFound(err) {
```
katrogan, line 156: if the pod doesn't exist, would it make sense to deallocate the resource just as a safeguard below?

go/tasks/plugins/array/k8s/task.go
```diff
@@ -0,0 +1,232 @@
...
+	if err != nil {
+		logger.Errorf(ctx, "Error releasing allocation token [%s] in Finalize [%s]", podName, err)
+		return err
+	}
+
+	return nil
+
+}
+
+func allocateResource(ctx context.Context, tCtx core.TaskExecutionContext, config *Config, podName string, childIdx int, arrayStatus *arraystatus.ArrayStatus) (bool, error) {
```
katrogan, line 174: what's the bool in the response represent?

go/tasks/plugins/array/core/state.go
```diff
@@ -238,11 +239,18 @@ func SummaryToPhase(ctx context.Context, minSuccesses int64, summary arraystatus
...
+		} else if phase.IsWaitingForResources() {
+			totalWaitingForResources += count
 		} else {
 			totalRunning += count
 		}
 	}
 
+	if totalWaitingForResources > 0 {
+		logger.Infof(ctx, "Array is still running and waiting for resources totalWaitingForResources[%v]", totalWaitingForResources)
+		return PhaseCheckingSubTaskExecutions
```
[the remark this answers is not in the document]
migueltol22: Good point. Updated.

go/tasks/plugins/array/k8s/monitor.go
migueltol22: updated.

go/tasks/plugins/array/k8s/monitor.go
migueltol22: Yes that is correct. If it's already running we will get `alreadyExists` error and proceed with the rest of the flow in monitor.

go/tasks/plugins/array/k8s/task.go
migueltol22: So there are essentially two state machines one for the entire Job and one for the subtasks. There are cases where we would like to return the entire job state based off of the outcome of a subtask. I couldn't come up with a great name. Any ideas?(will add a comment after deciding on name)

go/tasks/plugins/array/k8s/task.go
migueltol22: I'm not entirely sure. This was the code that was there before hand.

go/tasks/plugins/array/k8s/task.go
migueltol22: So abort in the handle step had never been implemented it was just calling finalize. Can update to actually call abort. Just have it as it because that was the preexisting implementation.

go/tasks/plugins/array/k8s/task.go
migueltol22: That makes sense. Updated.

go/tasks/plugins/array/k8s/task.go
migueltol22: So in the case we do not get a resource we do not want to continue with the flow. We'd like to proceed to the next iteration in the for loop. I'm thinking a better solution would be to just return the allocation status and have the check being done by the caller of allocate Resource. Thoughts?

go/tasks/plugins/array/k8s/monitor.go
katrogan: thanks for explaining

go/tasks/plugins/array/k8s/task.go
katrogan: hm it seems odd that we check before de-referencing here but now below, can we maybe check and just throw an error?

go/tasks/plugins/array/k8s/task.go
katrogan: that sounds reasonable!

go/tasks/plugins/array/k8s/task.go
migueltol22: Updated to throw error. Not sure if this is what you meant or throw error below?

go/tasks/plugins/array/k8s/task.go
migueltol22: Updated.

go/tasks/plugins/array/k8s/launcher.go
```diff
@@ -68,113 +59,25 @@ func applyPodTolerations(_ context.Context, cfg *Config, pod *corev1.Pod) *corev
...
-
-	currentState.SetPhase(arrayCore.PhaseCheckingSubTaskExecutions, 0).SetReason("Job launched.")
-	currentState.SetArrayStatus(arrayStatus)
-
-	return currentState, nil
-}
-
-func TerminateSubTasks(ctx context.Context, tMeta core.TaskExecutionMetadata, kubeClient core.KubeClient,
-	errsMaxLength int, currentState *arrayCore.State) error {
+func TerminateSubTasks(ctx context.Context, tCtx core.TaskExecutionContext, kubeClient core.KubeClient, config *Config,
```
EngHabu, line 62: I would either rename it to FinalizeSubTasks or change it to call Abort & Finalize instead

go/tasks/plugins/array/k8s/monitor.go
```diff
@@ -24,73 +23,100 @@ import (
...
 	for childIdx, existingPhaseIdx := range currentState.GetArrayStatus().Detailed.GetItems() {
 		existingPhase := core.Phases[existingPhaseIdx]
+		indexStr := strconv.Itoa(childIdx)
+		podName := formatSubTaskName(ctx, tCtx.TaskExecutionMetadata().GetTaskExecutionID().GetGeneratedName(), indexStr)
 		if existingPhase.IsTerminal() {
 			// If we get here it means we have already "processed" this terminal phase since we will only persist
 			// the phase after all processing is done (e.g. check outputs/errors file, record events... etc.).
+
+			// Since we know we have already "processed" this terminal phase we can safely deallocate resource
+			err = deallocateResource(ctx, tCtx, config, childIdx)
```
EngHabu, line 70: I think this can have an issue, in the last round (where SummaryToPhase would return a different phase), we will not dealocate resources for the tasks that JUST completed.

go/tasks/plugins/array/k8s/config.go
```diff
@@ -34,11 +34,17 @@ var (
...
+	PrimaryLabel string `json:"primaryLabel" pflag:",PrimaryLabel of a given service cluster"`
+	Limit        int    `json:"limit" pflag:",Resource quota (in the number of outstanding requests) for the cluster"`
+}
+
 // Defines custom config for K8s Array plugin
 type Config struct {
 	DefaultScheduler     string            `json:"scheduler" pflag:",Decides the scheduler to use when launching array-pods."`
 	MaxErrorStringLength int               `json:"maxErrLength" pflag:",Determines the maximum length of the error string returned for the array."`
 	MaxArrayJobSize      int64             `json:"maxArrayJobSize" pflag:",Maximum size of array job."`
+	ResourceConfig       ResourceConfig    `json:"resourceConfig" pflag:"-,ResourceConfiguration to limit number of resources used by k8s-array."`
```
EngHabu, line 47: Any reason for the `-` in pflag? this will make it not generate pflags options... desired or copy paste error?

go/tasks/plugins/array/k8s/task.go
EngHabu: I think it's also confusing that we return Success from Launch when we successfully launch and we return the same Success from Monitor() when the task finishes successfully.
Maybe you just need LaunchResult and MonitorResult (two separate enums)

go/tasks/plugins/array/k8s/task.go
```diff
@@ -0,0 +1,237 @@
...
+
+	t.NewArrayStatus.Detailed.SetItem(t.ChildIdx, bitarray.Item(actualPhase))
+	t.NewArrayStatus.Summary.Inc(actualPhase)
+
+	return Success, nil
+}
+
+func (t Task) Abort() {}
+
+func (t Task) Finalize(ctx context.Context, tCtx core.TaskExecutionContext, kubeClient core.KubeClient) error {
```
EngHabu, line 145: So this is the implementation of Abort.. Finalize should just try to deallocateResource

migueltol22: I see. Updated abort and finalize.

go/tasks/plugins/array/k8s/config.go
migueltol22: Good catch. Had it originally like that because I was using a map and pflag doesn't support that type. Updated.

go/tasks/plugins/array/k8s/monitor.go
migueltol22: I see. But we can run into issues if we call after the task just completed as well. Any ideas on how to handle this?

go/tasks/plugins/array/k8s/launcher.go
migueltol22: I updated it to call Abort & Finalize.

go/tasks/plugins/array/k8s/task.go
migueltol22: Updated to use two seperate enums. Still unsure about the name for `returnState`. Any ideas for a better name?

go/tasks/plugins/array/k8s/monitor.go
migueltol22: Actually shouldn't finalized be call before returning so it would deallocate the resource there in this scenario?

go/tasks/plugins/array/k8s/config.go
```diff
@@ -34,11 +34,17 @@ var (
 	configSection = config.MustRegisterSection(configSectionKey, defaultConfig)
 )
 
+type ResourceConfig struct {
+	PrimaryLabel string `json:"primaryLabel" pflag:",PrimaryLabel of a given service cluster"`
```
EngHabu, line 38: This is the cluster name that will be used in resource manager?

go/tasks/plugins/array/k8s/launcher.go
```diff
@@ -68,113 +59,29 @@ func applyPodTolerations(_ context.Context, cfg *Config, pod *corev1.Pod) *corev
...
-			},
+	for childIdx := 0; childIdx < size; childIdx++ {
+		task := Task{
+			ChildIdx: childIdx,
+			Config:   config,
+			State:    currentState,
 		}
-
-		err := kubeClient.GetClient().Delete(ctx, pod)
+		err := task.Finalize(ctx, tCtx, kubeClient)
```
EngHabu, line 73: We should abort first then finalize

go/tasks/plugins/array/k8s/task.go
```diff
@@ -0,0 +1,256 @@
...
+				return err
+			}
+			return nil
+		}
+
+		return err
+	}
+
+	// Deallocate Resource
+	err = deallocateResource(ctx, tCtx, t.Config, t.ChildIdx)
```
EngHabu, line 179: Is this needed?

go/tasks/plugins/array/k8s/config.go
migueltol22: Yes, that is correct.

go/tasks/plugins/array/k8s/launcher.go
migueltol22: Updated.

go/tasks/plugins/array/k8s/task.go
migueltol22: No, removed.

go/tasks/plugins/array/k8s/monitor.go
kumare3: @migueltol22 good question, so I would always release on successful completion of a task. The finalize of the one job should probably be invoked in 2 places, immediately after success and one in the finalize of the parent. If the finalize previously succeeded it would just be a no-op. Thus you will not have to worry about not release resources.

go/tasks/plugins/array/k8s/launcher.go
```diff
@@ -31,6 +21,7 @@ const (
 	ErrSubmitJob              errors2.ErrorCode = "SUBMIT_JOB_FAILED"
 	JobIndexVarName           string            = "BATCH_JOB_ARRAY_INDEX_VAR_NAME"
 	FlyteK8sArrayIndexVarName string            = "FLYTE_K8S_ARRAY_INDEX"
+	ResourcesPrimaryLabel     string            = "token"
```
EngHabu, line 24: nit: should this be something like "current_cluster" or something?

go/tasks/plugins/array/k8s/monitor.go
EngHabu: ic. I missed that you attempt to deallocate tokens in finalize anyway... awesome

go/tasks/plugins/array/k8s/task.go
EngHabu: I think we attempt to deallocateResources in Finalize. Do we need to do it here again?

migueltol22: No we do not. Updated.

go/tasks/plugins/array/k8s/launcher.go
migueltol22: Actually this value isn't being used. The value specified as part of the config is being used. Removed.

go/tasks/plugins/array/core/state.go
```diff
@@ -238,11 +239,18 @@ func SummaryToPhase(ctx context.Context, minSuccesses int64, summary arraystatus
...
 				totalFailures += count
 			}
+		} else if phase.IsWaitingForResources() {
+			totalWaitingForResources += count
 		} else {
 			totalRunning += count
 		}
 	}
 
+	if totalWaitingForResources > 0 {
```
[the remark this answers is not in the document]
migueltol22: That is correct. I updated the existing function to check for `totalRunning+totalSuccesses+totalWaitingForResources < minSuccesses` instead of just `totalRunning+totalSuccesses < minSuccesses`

go/tasks/plugins/array/k8s/task.go
EngHabu: @katrogan ^^

katrogan: yes, thank you!

DouglasCurbelo:
@katrogan @migueltol22, sorry I stop following this change. Is the change ready to merge, if not what is missing?

go/tasks/plugins/array/k8s/executor.go
```diff
@@ -164,5 +168,14 @@ func GetNewExecutorPlugin(ctx context.Context, iCtx core.SetupContext) (core.Plu
 		return nil, err
 	}
 
+	if IsResourceConfigSet() {
+		primaryLabel := GetConfig().ResourceConfig.PrimaryLabel
+		limit := GetConfig().ResourceConfig.Limit
+		if err := iCtx.ResourceRegistrar().RegisterResourceQuota(ctx, core.ResourceNamespace(primaryLabel), limit); err != nil {
```
bnsblue, line 174: @EngHabu regarding the resource name: We discussed earlier that we might need to add flyte cluster information to the resource name as a preparation step for moving toward a centralized Redis. Do you think the change should be done in this PR or a separate one?  cc @migueltol22 just FYI

EngHabu: I might have been mistaken. Maybe there is no way to get the cluster name... we might have to do something in our deployment to do that... might not be so trivial... let me ask on the channel..

PR #72, opened 2020-03-23
Add NodeSelector Config for k8s array

migueltol22:
# TL;DR
Add support for specifying node selector as part of k8s array plugin config

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/lyft/flyte/issues/220

## Follow-up issue
_NA_
OR

### DIFF
--- a/go/tasks/plugins/array/k8s/config.go
+++ b/go/tasks/plugins/array/k8s/config.go
@@ -32,6 +32,11 @@ var (
 	configSection = config.MustRegisterSection(configSectionKey, defaultConfig)
 )
 
+type NodeSelectorConfig struct {
+	key   string `json:"key", pflag:",Node Selector key.`
+	value string `json:"value", pflag:",Node Selector value."`
+}
+
 // Defines custom config for K8s Array plugin
 type Config struct {
 	DefaultScheduler     string `json:"scheduler" pflag:",Decides the scheduler to use when launching array-pods."`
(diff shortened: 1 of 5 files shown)


go/tasks/plugins/array/k8s/launcher.go
```diff
@@ -52,6 +52,14 @@ func ApplyPodPolicies(_ context.Context, cfg *Config, pod *corev1.Pod) *corev1.P
 	return pod
 }
 
+func applyNodeSelectorLabels(_ context.Context, cfg *Config, pod *corev1.Pod) *corev1.Pod {
```
vsbus: should function start with uppercase? : `ApplyNodeSelectorLabels`

surindersinghp:
please add some tests

go/tasks/plugins/array/k8s/config.go
```diff
@@ -32,13 +32,19 @@ var (
 	configSection = config.MustRegisterSection(configSectionKey, defaultConfig)
 )
 
+type NodeSelectorConfig struct {
+	key   string `json:"key", pflag:",Node Selector key.`
```
EngHabu: I think Pod's NodeSelector supports multiple keys here right? and it does an AND between them or something?
we should match that behavior then

surindersinghp: agreed, like we do for interruptible: https://github.com/lyft/flyteplugins/blob/master/go/tasks/pluginmachinery/flytek8s/config/config.go#L51

migueltol22: Thanks for the reference! Updated.

go/tasks/plugins/array/k8s/launcher.go
migueltol22: Since I'm not exporting this function, it should be fine lowercase.

vsbus: ah, gotcha! 👍

PR #74, opened 2020-04-02
Fix PhaseStatus that gets displayed for Presto tasks

lu4nm3:
# TL;DR
When running a Presto task, the status of each of the 5 queries is not displayed properly.
Screen Shot 2020-04-02 at 2 04 27 PM

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

## Complete description
Due to how the state machine the Presto plugin uses goes back and forth between various states, I keep track of the of both the current and previous states and use the previous state for the status that gets displayed to users.

## Tracking Issue
https://github.com/lyft/flyte/issues/237

### DIFF
--- a/go/tasks/plugins/presto/execution_state.go
+++ b/go/tasks/plugins/presto/execution_state.go
@@ -56,7 +56,8 @@ func (p ExecutionPhase) String() string {
 }
 
 type ExecutionState struct {
-	Phase ExecutionPhase
+	CurrentPhase  ExecutionPhase
+	PreviousPhase ExecutionPhase
 
 	// This will store the command ID from Presto
 	CommandID string `json:"commandId,omitempty"`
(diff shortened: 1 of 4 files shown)


wild-endeavor:
So what's the end result of this?  The text for each log link still moves through the phases right? Like the first one will appear queued, then running, then succeeded, and then the next one will show up queued. etc.

If we're storing the individual query IDs, and they're always run serially, and you only launch the next one if the current one succeeds, then we can also just always assume that all the previous ones are in a succeeded state right?

lu4nm3:
@wild-endeavor Yeah that seems like a reasonable assumption to make. The part where it sets the previous status for the last time (ie. succeeded) is [here](https://github.com/lyft/flyteplugins/pull/74/files#diff-6a65d420579b33b62766e71e98eafdf7R133).

lu4nm3:
The problem that was happening is that after each query, we set the phase to "queued" (so that it can keep advancing through all 5 queries). However, because this is set to "queued" it keeps that phase through to when it gets logged and that's why it shows up that way in the UI

PR #75, opened 2020-04-03
Add a rate limiter and use the RawOutputPrefix in Presto plugin

lu4nm3:
# TL;DR
- Add rate limiter configs for the Presto client
- Replace hard-coded external location with `GetRawOutputPrefix`

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [X] Unit tests added

### DIFF
--- a/go/tasks/plugins/presto/config/config.go
+++ b/go/tasks/plugins/presto/config/config.go
@@ -40,6 +40,11 @@ type RefreshCacheConfig struct {
 	LruCacheSize int             `json:"lruCacheSize" pflag:",Size of the cache"`
 }
 
+type RateLimiterConfig struct {
+	Rate  int64 `json:"rate" pflag:",Allowed rate of calls per second."`
+	Burst int   `json:"burst" pflag:",Allowed burst rate of calls."`
+}
+
 var (
 	defaultConfig = Config{
 		Environment:         URLMustParse(""),
@@ -51,6 +56,10 @@ var (
 			SyncPeriod:   config.Duration{Duration: 5 * time.Second},
 			Workers:      15,
 			LruCacheSize: 10000,
+		},
+		RateLimiterConfig: RateLimiterConfig{
+			Rate:  15,
+			Burst: 20,
 		},
 	}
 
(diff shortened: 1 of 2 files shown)


go/tasks/plugins/presto/config/config.go
```diff
@@ -40,6 +40,11 @@ type RefreshCacheConfig struct {
 	LruCacheSize int             `json:"lruCacheSize" pflag:",Size of the cache"`
 }
 
+type RateLimiterConfig struct {
+	Rate  int64 `json:"rate" pflag:",Allowed rate of calls per second."`
+	Burst int   `json:"burst" pflag:",Allowed burst rate of calls."`
```
wild-endeavor, line 45:
```suggestion
	Burst int   `json:"burst" pflag:",Allowed burst rate of calls per second."`
```

go/tasks/plugins/presto/config/config.go
```diff
@@ -52,6 +57,10 @@ var (
 			Workers:      15,
 			LruCacheSize: 10000,
 		},
+		RateLimiterConfig: RateLimiterConfig{
+			Rate:  15,
```
wild-endeavor, line 61: What will this be used for?  Will it gate _all_ calls to Mozart?  Or just Presto?  Will it be for only creation calls?  Or both creation calls and all the status checking calls in the auto-refresh cache as well?  Can you add some comments about how best to calculate a good number for this config?

wild-endeavor:
Can you create and link issue as well?  Or use an existing issue?

go/tasks/plugins/presto/config/config.go
lu4nm3: This will be used for all calls going to Mozart only  for Presto. Sure I can add some comments.

lu4nm3: In terms of the values themselves, I sort of took them from the awsbatch rate limiter. I'm sure we can adjust them later based on how the Presto cluster behaves

lu4nm3:
Made changes from feedback and linked issue

PR #78, opened 2020-04-11
SchedulerName config

vsbus:
# TL;DR

Add SchedulerName option that will propagate to PodSpec

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

I've updated code to use this flag in 2 places:
* go/tasks/pluginmachinery/flytek8s/pod_helper.go (regular tasks)

## Tracking Issue

## Follow-up issue
_NA_
OR


go/tasks/plugins/k8s/sidecar/sidecar.go
```diff
@@ -62,6 +62,7 @@ func validateAndFinalizeContainers(
 	}
 	pod.Spec.Containers = finalizedContainers
 	pod.Spec.Tolerations = flytek8s.GetPodTolerations(taskCtx.TaskExecutionMetadata().IsInterruptible(), resReqs...)
+	pod.Spec.SchedulerName = flytek8s.GetPodTolerations(taskCtx.TaskExecutionMetadata().SchedulerName(), resReqs...)
```
vsbus, line 65: @EngHabu @surindersinghp @migueltol22 is it correct to update sidecar with this option?

Should I also update go/tasks/plugins/k8s/spark/spark.go?

vsbus:
PTAL @migueltol22 @EngHabu @surindersinghp 

This is my first time contributing to Flyte code. Please LMK what should I do to run certain tests/checks. Thanks!

go/tasks/plugins/k8s/sidecar/sidecar.go
migueltol22: This shouldn't be necessary. If my understanding is correct this value will already be populated for all tasks that run on kubernetes if set, so we shouldn't need to modify any of the tasks.

migueltol22:
@vsbus for running unit tests you can run `make test_unit`

vsbus:
Tests were failing in the sidecar code. reverter the sidecar file

vsbus:
PTAL @EngHabu

go/tasks/pluginmachinery/flytek8s/pod_helper.go
```diff
@@ -43,6 +43,7 @@ func ToK8sPodSpec(ctx context.Context, taskExecutionMetadata pluginsCore.TaskExe
 			Tolerations:        GetPodTolerations(taskExecutionMetadata.IsInterruptible(), c.Resources),
 			ServiceAccountName: taskExecutionMetadata.GetK8sServiceAccount(),
 			NodeSelector:       config.GetK8sPluginConfig().InterruptibleNodeSelector,
+			SchedulerName:      config.GetK8sPluginConfig().SchedulerName,
```
surindersinghp, line 46: I will say that we set scheduler only if it is set.

surindersinghp:
looks good overall.

go/tasks/pluginmachinery/flytek8s/pod_helper.go
vsbus: Agree. I tested it and when the flag is unset in .yml file, the pod is using `default-scheduler`. So even with writing a value here the behavior is still matching expectations

PR #87, opened 2020-05-12
Read the stderr into termination reason when container fails

EngHabu:
# TL;DR
This change will automatically read the pod log and set it as the termination reason when the container exits with non-zero exit code. In case of flytekit, and it fails to write error.pb, the log will have more information to surface to users.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Tracking Issue
https://github.com/lyft/flyte/issues/308


kumare3:
No parsing just dropping a better error message

kumare3:
I was thinking if we should use this always?

EngHabu:
I think it'll almost always give you a better idea.... I'm thinking we can further limit the size of the message on our side... 
What are your concerns?

PR #90, opened 2020-05-15
Remote Plugins Interface and Implementation

EngHabu:
# TL;DR
_Please replace this text with a description of what this PR accomplishes._

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


go/tasks/pluginmachinery/core/resource_manager.go
```diff
@@ -4,21 +4,23 @@ import (
...
 )
 
-type AllocationStatus string
+//go:generate enumer -type=AllocationStatus -trimprefix=AllocationStatus
+
+type AllocationStatus int
 
 const (
 	// This is the enum returned when there's an error
-	AllocationUndefined AllocationStatus = "ResourceGranted"
```
kumare3, line 11: isnt this a breaking change?

kumare3: not sure, but do we store this in redis?

go/tasks/pluginmachinery/remote/properties.go
```diff
@@ -0,0 +1,66 @@
...
+		},
+		WriteRateLimiter: RateLimiterProperties{
+			QPS:   20,
+			Burst: 200,
+		},
+	}
+)
+
+// The plugin manager automatically queries the remote API
+type RateLimiterProperties struct {
```
kumare3, line 33: why not config? seems to be our standardized name?

go/tasks/pluginmachinery/remote/properties.go
```diff
@@ -0,0 +1,66 @@
...
+	// Queries per second from one process to the remote service
+	QPS int `json:"qps" pflag:",Defines the max rate of calls per second."`
+
+	// Maximum burst size
+	Burst int `json:"burst" pflag:",Defines the maximum burst size."`
+}
+
+type CachingProperties struct {
+	// Max number of Resource's to be stored in the local cache
+	Size int `json:"size" pflag:",Defines the maximum number of items to cache."`
```
kumare3, line 43: we will have a default max right?

go/tasks/pluginmachinery/remote/plugin.go
```diff
@@ -0,0 +1,81 @@
...
+
+	// Returns a handle to the Task's execution metadata.
+	TaskExecutionMetadata() pluginsCore.TaskExecutionMetadata
+
+	// Provides an output sync of type io.OutputWriter
+	OutputWriter() io.OutputWriter
+}
+
+// The resource to be sycned from the remote
+type ResourceMeta = interface{}
```
wild-endeavor, line 57: Why is this called "Meta"?  What is meta about it?  It seems like it's the resource itself, not metadata about the resource.

go/tasks/pluginmachinery/internal/remote/state.go
```diff
@@ -0,0 +1,45 @@
...
+)
+
+func (p Phase) IsTerminal() bool {
+	return p == PhaseSucceeded || p == PhaseFailed
+}
+
+type State struct {
+	Phase Phase `json:"phase"`
+
+	// ResourceKeys contain the resource keys of the launched resources. If this is empty, it means
```
wild-endeavor, line 29: What are ResourceKeys?

go/tasks/pluginmachinery/internal/remote/allocation_token.go
```diff
@@ -0,0 +1,88 @@
+package remote
```
wild-endeavor, line 1: Can we think about this name "remote"?  I like the idea of service or something.  A call to create a K8s pod is also remote right?  Like, it's not done in Propeller or in the plugins - it has to contact the K8s API server, which is remote.  I feel like the defining feature of this whole concept is not that it's remote but rather that it's a non-Flyte controlled service that creates some entities on behalf of Flyte.  Should/could this framework be used to interact with external services that don't offer a read/get endpoint?  What about an external service that doesn't offer read after write consistency?

go/tasks/pluginmachinery/core/resource_manager.go
EngHabu: We do not store this value anywhere as far as I can tell...

go/tasks/pluginmachinery/remote/plugin.go
EngHabu: You're right, it's actually metadata about the resource yes.

go/tasks/pluginmachinery/internal/remote/allocation_token.go
EngHabu: I know that was a long time ago... I'm finally coming back to this... Any suggestions? service also applies to k8s resources (we call ApiServer which is, at the end of the day, a service)... I don't feel strongly either way...

Bringing @kumare3 to the naming fun discussion...

kumare3:
will TAL tomorrow

PR #91, opened 2020-05-19
Mark Hive and Presto query failures as Retryable to allow Propeller to decide the behavior

EngHabu:
# TL;DR
Hive and Presto plugins report Permanent Failure when the query fails. This signals to Propeller that it should not retry the queries. Given that Hive & Presto do not report detailed failure reasons, it's not easy to figure out whether the query is in fact retryable or not. Returning Retryable all the time would at least enable Propeller to use user's retry strategy to resubmit these queries.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [X] Unit tests added

## Tracking Issue


kumare3:
LGTM

PR #92, opened 2020-05-26
Allow multiple arguments to be substituted in one run

kumare3:
# TL;DR
A query like
```sql
     SELECT
        	COUNT(*) as total_count
    	FROM
        	hive.events.{{ .Inputs.table }}
    	WHERE
        	ds = '{{ .Inputs.ds }}' AND hr = {{ .Inputs.hr }} AND min = {{ .Inputs.min }}
```
should be substituted. 

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/lyft/flyte/issues/320

## Follow-up issue
https://github.com/lyft/flyte/issues/321

### DIFF
--- a/go/tasks/pluginmachinery/utils/error_collection.go
+++ b/go/tasks/pluginmachinery/utils/error_collection.go
@@ -1,0 +1,19 @@
+package utils
+
+import (
+	"fmt"
+	"strings"
+)
+
+type ErrorCollection struct {
+	Errors []error
+}
+
+func (e ErrorCollection) Error() string {
+	sb := strings.Builder{}
+	for idx, err := range e.Errors {
+		sb.WriteString(fmt.Sprintf("%v: %v\r\n", idx, err))
+	}
+
+	return sb.String()
+}
--- a/go/tasks/pluginmachinery/utils/template.go
+++ b/go/tasks/pluginmachinery/utils/template.go
@@ -51,7 +51,7 @@ func ReplaceTemplateCommandArgs(ctx context.Context, command []string, in io.Inp
 	return res, nil
 }
 
-func transformVarNameToStringVal(ctx context.Context, varName string, inputs *core.LiteralMap) (string, error){
+func transformVarNameToStringVal(ctx context.Context, varName string, inputs *core.LiteralMap) (string, error) {
 	inputVal, exists := inputs.Literals[varName]
 	if !exists {
 		return "", fmt.Errorf("requested input is not found [%s]", varName)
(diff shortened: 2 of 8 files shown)


go/tasks/pluginmachinery/utils/error_collection.go
```diff
@@ -0,0 +1,19 @@
+package utils
+
+import (
+	"fmt"
+	"strings"
+)
+
+type ErrorCollection struct {
```
EngHabu, line 8: Can you use this? https://github.com/lyft/flytestdlib/blob/537f86093d86af270aab300742ee1a56f2905885/config/errors.go#L11

go/tasks/pluginmachinery/utils/error_collection.go
```diff
@@ -0,0 +1,19 @@
+package utils
```
wild-endeavor, line 1: i think maybe plugins doesn't use this but can we move this to https://github.com/lyft/flytestdlib/blob/master/errors/error.go eventually?

go/tasks/pluginmachinery/utils/template.go
```diff
@@ -51,47 +51,51 @@ func ReplaceTemplateCommandArgs(ctx context.Context, command []string, in io.Inp
...
-		return val, errors.Wrapf(err, "failed to bind a value to inputName [%s]", inputName)
-	}
-	if endIdx >= len(val) {
-		return val[:startIdx] + v, nil
-	}
+	totalMatches := 0
+	var errs ErrorCollection
+	val = inputVarRegex.ReplaceAllStringFunc(val, func(s string) string {
+		totalMatches++
+		matches := inputVarRegex.FindAllStringSubmatch(s, 1)
```
EngHabu, line 84: Do we need to do this?

go/tasks/pluginmachinery/utils/error_collection.go
kumare3: i saw that, its not in the errors package but in config. isn't that weird?

go/tasks/pluginmachinery/utils/template.go
kumare3: yup i will remove

go/tasks/pluginmachinery/utils/error_collection.go
kumare3: yes i agree we should move, its replicated in a few places. I will create an issue

PR #93, opened 2020-05-28
Add success and failure metrics for Hive and Presto

anandswaminathan:
Adding metrics for Hive and Presto success and failure.

### DIFF
--- a/go/tasks/plugins/hive/executor.go
+++ b/go/tasks/plugins/hive/executor.go
@@ -61,6 +61,14 @@ func (q QuboleHiveExecutor) Handle(ctx context.Context, tCtx core.TaskExecutionC
 	// If no error, then infer the new Phase from the various states
 	phaseInfo := MapExecutionStateToPhaseInfo(outgoingState, q.quboleClient)
 
+	if phaseInfo.Phase() == core.PhaseSuccess {
+		q.metrics.Succeeded.Inc(ctx)
+	}
+
+	if phaseInfo.Phase() == core.PhaseRetryableFailure || phaseInfo.Phase() == core.PhasePermanentFailure {
+		q.metrics.Failed.Inc(ctx)
+	}
+
 	if err := tCtx.PluginStateWriter().Put(pluginStateVersion, outgoingState); err != nil {
 		return core.UnknownTransition, err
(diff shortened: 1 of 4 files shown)


anandswaminathan:
PTAL @lu4nm3 @wild-endeavor

go/tasks/plugins/hive/executor.go
```diff
@@ -61,6 +61,14 @@ func (q QuboleHiveExecutor) Handle(ctx context.Context, tCtx core.TaskExecutionC
 	// If no error, then infer the new Phase from the various states
 	phaseInfo := MapExecutionStateToPhaseInfo(outgoingState, q.quboleClient)
 
+	if phaseInfo.Phase() == core.PhaseSuccess {
```
wild-endeavor, line 64: @EngHabu This uses the barrier transition - this guarantees for the most part that this will only trigger once per successful task right?

anandswaminathan: Yup. Uses the response from `MapExecutionStateToPhaseInfo`

kumare3: yes you are right.

kumare3: @anandswaminathan  should we not do this in the common task layer? and use task_type as the label?

kumare3:
Why should we do this per plugin and not in the common layer?

anandswaminathan:
@kumare3 Just cleaner code and metrics. If I do in the common layer, I will need to have a map of each types and success/failure and so on. Here it is correctly nested.

go/tasks/plugins/hive/executor.go
EngHabu: Agreed with Ketan, I think it should move higher up in the stack with task_type as a label (or even part of the metric key)

kumare3:
@anandswaminathan I think if you do not do this in the common layer, we will have to repeat it for every Plugin. Which I think is further overhead and place for divergence. I would like that all metrics for every plugin look identical

anandswaminathan:
@kumare3 Makes sense. Will come with PR in Propeller then.

PR #94, opened 2020-06-03
Add AWS Batch metrics for queued

anandswaminathan:
## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin


go/tasks/plugins/array/awsbatch/monitor.go
```diff
@@ -63,9 +63,13 @@ func CheckSubTasksState(ctx context.Context, taskMeta core.TaskExecutionMetadata
 		Detailed: arrayCore.NewPhasesCompactArray(uint(currentState.GetExecutionArraySize())),
 	}
 
+	queued := 0
 	for childIdx, subJob := range job.SubJobs {
 		actualPhase := subJob.Status.Phase
 		originalIdx := arrayCore.CalculateOriginalIndex(childIdx, currentState.GetIndexesToCache())
+		if subJob.Status.Phase == core.PhaseQueued {
+			queued++
```
wild-endeavor, line 71: Remind me again... 
1) what is this metric labeled by?  just project and domain?  Is execution id in there?
2) This counts the number of each element of the batch array task right?  So if a dynamic task yields 10 tasks that gets run as 1 array task, and 5 are queued and 5 are running, this counter will be incremented to 5.

anandswaminathan: 1) By Wf- "Project:domain:name"
2) Yes. It indicates the number of queued task monitored in every loop.

wild-endeavor: do you mean the workflow name? or the execution name?

wild-endeavor: if there are two executions of the same workflow (and hence the same task), and each one has 5 running and 5 queued, what will this counter report?

anandswaminathan: It will do add(5) twice. Depends on the poll interval between the two workflows in propeller. You would see 5 and then 10 in prometheus

wild-endeavor: cool, thanks.

PR #95, opened 2020-06-04
Set the executor prod prefix in k8s

catalinii:
Spark 2.4.6 changed the name of the executor pod name to not contain the execution id. This PR fixes the executor name to have the flyte execution id.


kumare3:
Hey @catalinii thank you for the change. We usually try to create an issue (and you can definite that the problem is 2.4.6) and then associate this PR with the issue. This way when we close the milestone for the month, we automatically get changelog

catalinii:
Added https://github.com/lyft/flyte/issues/344

akhurana001:
LGTM 👍

PR #96, opened 2020-06-15
Presto - add option to use namespace as the user

wild-endeavor:
# TL;DR
In an effort to reduce the occurrence of the error in the issue, this makes the user that is sent when creating a Presto query the K8s namespace instead of the default username.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
This changes the user to be the K8s namespace.

## Tracking Issue
https://github.com/lyft/flyte/issues/348

## Follow-up issue
NA

### DIFF
--- a/go/tasks/plugins/presto/config/config.go
+++ b/go/tasks/plugins/presto/config/config.go
@@ -76,6 +76,7 @@ var (
 		Environment:         URLMustParse(""),
 		DefaultRoutingGroup: "adhoc",
 		DefaultUser:         "flyte-default-user",
+		UseNamespaceAsUser:  true,
 		RoutingGroupConfigs: []RoutingGroupConfig{{Name: "adhoc", Limit: 100}, {Name: "etl", Limit: 25}},
 		RefreshCacheConfig: RefreshCacheConfig{
 			Name:         "presto",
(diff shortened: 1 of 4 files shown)


go/tasks/plugins/presto/config/config.go
```diff
@@ -101,6 +102,7 @@ type Config struct {
 	Environment            config.URL           `json:"environment" pflag:",Environment endpoint for Presto to use"`
 	DefaultRoutingGroup    string               `json:"defaultRoutingGroup" pflag:",Default Presto routing group"`
 	DefaultUser            string               `json:"defaultUser" pflag:",Default Presto user"`
+	UseNamespaceAsUser     bool                 `json:"useNamespaceAsUser" pflag:",Use the K8s namespace as the user"`
```
EngHabu, line 105: Do you see a world where we pass in the actual user principal as the user? for interactive usecases maybe?
should this be like:
```
UserParamSource []UserParamSource `json:"userParamSource" pflag:",An ordered list of sources to fallback to to fill in the user parameter when creating a presto job"
```

```
//go:generate enumer --type=UserParamSource --timprefix=UserParamSource --json

type UserParamSource int

const (
    UserParamSourceDefaultUser UserParamSource = iota
    UserParamSourceNamespace
    UserParamSourceEmpty
    UserParamSourceLoggedInUser
)
```

wild-endeavor: sure why not.  will use this instead

wild-endeavor: though I don't see the need to make this a list, can I just make it one thing for now?

wild-endeavor: actually, this adds a fair bit of code, i'll add this if/when the need arises.

PR #98, opened 2020-06-16
Presto - Reduce number of queries run

wild-endeavor:
# TL;DR
This reduces the number of actual queries sent to Presto from 5 to 2.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Previously, two tables were being created and then dropped.  This changes it to just be one query.

## Tracking Issue
https://github.com/lyft/flyte/issues/347

## Follow-up issue
NA


lu4nm3:
Looks good. You should also update the docs [here](https://github.com/lyft/flyte/blob/master/rsts/user/tasktypes/presto.rst) to reflect these changes

PR #100, opened 2020-06-23
SageMaker on Flyte: GO plugin [Alpha]

bnsblue:
# TL;DR
_Please replace this text with a description of what this PR accomplishes._

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


go/tasks/plugins/k8s/sagemaker/sagemaker.go
```diff
@@ -0,0 +1,320 @@
...
+					MaxNumberOfTrainingJobs: ToInt64Ptr(10),
+					MaxParallelTrainingJobs: ToInt64Ptr(5),
+				},
+				Strategy: "Bayesian",
+				HyperParameterTuningJobObjective: &commonv1.HyperParameterTuningJobObjective{
+					Type:       "Minimize",
+					MetricName: ToStringPtr("validation:error"),
+				},
+				// TODO: need to implement a factory that map names to different types of parameter range
+				ParameterRanges: &commonv1.ParameterRanges{
```
kumare3, line 129: this is not using your params?

bnsblue: I hadn't pushed my changes when you commented

go/tasks/plugins/k8s/sagemaker/constants.go
```diff
@@ -0,0 +1,41 @@
...
+
+import commonv1 "github.com/aws/amazon-sagemaker-operator-for-k8s/api/v1/common"
+
+const (
+	pluginID          = "aws_sagemaker_hpo"
+	sagemakerTaskType = "aws_sagemaker_hpo"
+)
+
+
+const (
```
kumare3, line 11: lets keep the constants in the file (hpo) where we use it

go/tasks/plugins/k8s/sagemaker/constants.go
```diff
@@ -0,0 +1,41 @@
...
+	BayesianSageMakerAPIHyperParameterTuningJobStrategyType commonv1.HyperParameterTuningJobStrategyType = "Bayesian"
+	RandomSageMakerAPIHyperParameterTuningJobStrategyType commonv1.HyperParameterTuningJobStrategyType = "Random"
+)
+
+const (
```
kumare3, line 16: should these be enumer?

go/tasks/plugins/k8s/sagemaker/constants.go
kumare3: should these be enumer?

go/tasks/plugins/k8s/sagemaker/constants.go
kumare3: if so should they be in the protobuf?

go/tasks/plugins/k8s/sagemaker/constants.go
bnsblue: These contants need have to  this type `commonv1.HyperParameterTuningJobStrategyType` tho, which is just a  redefinition of string.

In `github.com/aws/amazon-sagemaker-operator-for-k8s/api/v1/common/sagemaker_api.go`
```Go
type HyperParameterTuningJobStrategyType string
```

go/tasks/plugins/k8s/sagemaker/constants.go
bnsblue: This is a mapping from our protobuf type to sagemaker's API type tho?

go/tasks/plugins/k8s/sagemaker/hpo_job.go
```diff
@@ -0,0 +1,319 @@
...
+	if err != nil {
+		return nil, errors.Wrapf(err, "failed to convert hpo job config literal to spec type")
+	}
+
+	// If the container is part of the task template you can access it here
+	container := taskTemplate.GetContainer()
+
+	// When adding env vars there are some default env vars that are available, you can pass them through
+	envVars := flytek8s.DecorateEnvVars(ctx, flytek8s.ToK8sEnvVar(container.GetEnv()), taskCtx.TaskExecutionMetadata().GetTaskExecutionID())
+	_ = envVars
```
kumare3, line 111: ?

go/tasks/plugins/k8s/sagemaker/hpo_job.go
```diff
@@ -0,0 +1,319 @@
...
+					},
+				},
+				OutputDataConfig: &commonv1.OutputDataConfig{
+					S3OutputPath: awssagemaker.ToStringPtr(outputPath),
+				},
+				ResourceConfig: &commonv1.ResourceConfig{
+					InstanceType:   sagemakerHPOJob.GetTrainingJob().GetTrainingJobConfig().GetInstanceType(),
+					InstanceCount:  awssagemaker.ToInt64Ptr(sagemakerHPOJob.GetTrainingJob().GetTrainingJobConfig().GetInstanceCount()),
+					VolumeSizeInGB: awssagemaker.ToInt64Ptr(sagemakerHPOJob.GetTrainingJob().GetTrainingJobConfig().GetVolumeSizeInGb()),
+					VolumeKmsKeyId: awssagemaker.ToStringPtr(""), // TODO: add to proto and flytekit
```
kumare3, line 177: interesting, this is to support encrypted data i guess

bnsblue: Yes it is. I haven't added that to flytekit yet because I am not sure whether we should allow users to encrypt

go/tasks/plugins/k8s/sagemaker/hpo_job.go
bnsblue: This comes from the original implementation in your awsflyteplugin repo. I wasn't sure what the purpose was so I kept it here for now. Is it just for demonstration purpose?

go/tasks/plugins/k8s/sagemaker/utils.go
```diff
@@ -0,0 +1,218 @@
...
+	var err error
+	if specifiedAlg := job.GetAlgorithmSpecification().GetAlgorithmName(); specifiedAlg != sagemakerSpec.AlgorithmName_CUSTOM {
+		// Built-in algorithm mode
+		apiAlgorithmName := getAPIAlgorithmName(specifiedAlg)
+
+		// Getting the version
+		ver := job.GetAlgorithmSpecification().GetAlgorithmVersion()
+		if ver == "" {
+			// user didn't specify a version -> use the latest
+			ver, err = getLatestSemVer(cfg, apiAlgorithmName, cfg.Region)
```
kumare3, line 131: this is interesting, this introduces potential non-determinism to the plugin execution? I guess it is fine, but good to note

bnsblue: ah good point...

bnsblue:
🔨

copilot/go.mod
```diff
@@ -7,7 +7,7 @@ require (
 	github.com/gogo/protobuf v1.3.1
 	github.com/golang/protobuf v1.4.2
 	github.com/imdario/mergo v0.3.9 // indirect
-	github.com/lyft/flyteidl v0.17.32
+	github.com/lyft/flyteidl v0.17.37-0.20200727195325-5aef14f0da69
```
kumare3, line 10: once we merge flyteidl, we can just get this updated

go/tasks/plugins/k8s/sagemaker/config/config.go
```diff
@@ -0,0 +1,43 @@
...
+)
+
+const sagemakerConfigSectionKey = "sagemaker"
+
+var (
+	defaultConfig = Config{
+		RoleArn: "default",
+		Region:  "us-east-1",
+		// https://docs.aws.amazon.com/sagemaker/latest/dg/sagemaker-algo-docker-registry-paths.html
+		AlgorithmPrebuiltImages: map[string]map[string]map[string]string{
```
kumare3, line 16: does this work now? It was not working right

go/tasks/plugins/k8s/sagemaker/constants.go
```diff
@@ -0,0 +1,45 @@
...
+	trainingJobTaskType     = "sagemaker_training_job_task"
+)
+
+const (
+	hyperparameterTuningJobTaskPluginID = "sagemaker_hyperparameter_tuning"
+	hyperparameterTuningJobTaskType     = "sagemaker_hyperparameter_tuning_job_task"
+)
+
+const (
+	BayesianSageMakerAPIHyperParameterTuningJobStrategyType commonv1.HyperParameterTuningJobStrategyType = "Bayesian"
```
kumare3, line 16: It looks like we can just do `commonv1.HyperParameterTuningJobStrategyType(proto_enum.string())`
Is that not right?

kumare3: reason is you wont need a "switch" case.  thus if we add some new types, we just have to add it to the proto.

go/tasks/plugins/k8s/sagemaker/constants.go
kumare3: look at the previous comment

go/tasks/plugins/k8s/sagemaker/utils.go
```diff
@@ -0,0 +1,281 @@
...
+	awssagemaker "github.com/aws/amazon-sagemaker-operator-for-k8s/controllers/controllertest"
+	"github.com/lyft/flyteidl/gen/pb-go/flyteidl/core"
+	sagemakerSpec "github.com/lyft/flyteidl/gen/pb-go/flyteidl/plugins/sagemaker"
+	"github.com/lyft/flyteplugins/go/tasks/plugins/k8s/sagemaker/config"
+	"github.com/pkg/errors"
+
+	"github.com/golang/protobuf/proto"
+)
+
+func getAPIHyperParameterTuningJobStrategyType(
```
kumare3, line 21: if we just typecast you would not need this. And that should be ok right?

kumare3: all the switch statements below?

go/tasks/plugins/k8s/sagemaker/config/config.go
bnsblue: argh... I don't think this works but I am not sure how to fix it...... @kumare3 mind giving some hint?

go/tasks/plugins/k8s/sagemaker/utils.go
```diff
@@ -0,0 +1,281 @@
...
+	}
+	return op
+}
+
+func deleteConflictingStaticHyperparameters(
+	ctx context.Context,
+	staticHPs []*commonv1.KeyValuePair,
+	tunableHPMap map[string]*sagemakerSpec.ParameterRangeOneOf) []*commonv1.KeyValuePair {
+
+	w := 0 // write position
```
kumare3, line 268: why do you need this, why not just use append? Anyways you are creating a new array. I think you just need to change `make([]type, 0, len(...))`
What this will do is create a slice of length `len` but start at position 0. So now you can just append

go/tasks/plugins/k8s/sagemaker/utils.go
kumare3: comment?

go/tasks/plugins/k8s/sagemaker/utils.go
bnsblue: Isn't it going to create new slice whenever `append` is called?

go/tasks/plugins/k8s/sagemaker/config/config.go
kumare3: I should try this, let me try.
But
``` map[string]X
where X {
map[string]Y
}```
might work

Comment (kumare3) [reply go/tasks/plugins/k8s/sagemaker/config/config.go:16]: @bnsblue I will pull your branch and try out

File: go/tasks/plugins/k8s/sagemaker/utils.go (hunk 10 above)
Comment (bnsblue) [reply go/tasks/plugins/k8s/sagemaker/utils.go:21]: how do you type cast from 

```go
const (
	HyperparameterTuningStrategy_BAYESIAN HyperparameterTuningStrategy_Value = 0
	HyperparameterTuningStrategy_RANDOM   HyperparameterTuningStrategy_Value = 1
)
```

to 
 ```go
const (
	BayesianSageMakerAPIHyperParameterTuningJobStrategyType commonv1.HyperParameterTuningJobStrategyType = "Bayesian"
	RandomSageMakerAPIHyperParameterTuningJobStrategyType   commonv1.HyperParameterTuningJobStrategyType = "Random"
)
```
?

File: go/tasks/plugins/k8s/sagemaker/utils.go (hunk 11 above)
Comment (bnsblue) [reply go/tasks/plugins/k8s/sagemaker/utils.go:268]: Are these two approaches incurring the same number of memory block re-allocations? but yeah your suggestion is more elegant

File: go/tasks/plugins/k8s/sagemaker/utils.go (hunk 10 above)
Comment (bnsblue) [reply go/tasks/plugins/k8s/sagemaker/utils.go:21]: @kumare3 ^

Comment (bnsblue) [reply go/tasks/plugins/k8s/sagemaker/utils.go:21]: oh I saw your code snippet above

Comment (bnsblue) [reply go/tasks/plugins/k8s/sagemaker/utils.go:21]: wait that doesn't work right?

File: copilot/go.mod (hunk 7 above)
Comment (kumare3) [reply copilot/go.mod:10]: you should update this now? - Flyteidl is merged

File: go/tasks/plugins/k8s/sagemaker/config/config.go [hunk 12]
```diff
@@ -0,0 +1,86 @@
+package config
+
+import pluginsConfig "github.com/lyft/flyteplugins/go/tasks/config"
+
+//go:generate pflags Config --default-var=defaultConfig
+
+const sagemakerConfigSectionKey = "sagemaker"
+
+//var (
```
kumare3, line 9: delete this section?

go/tasks/plugins/k8s/sagemaker/sagemaker.go
```diff
@@ -0,0 +1,581 @@
...
+	if err != nil {
+		return pluginsCore.PhaseInfoUndefined, err
+	}
+
+	occurredAt := time.Now()
+
+	switch trainingJob.Status.TrainingJobStatus {
+	case trainingjobController.ReconcilingTrainingJobStatus:
+		logger.Errorf(ctx, "Job stuck in reconciling status, assuming retryable failure [%s]", trainingJob.Status.Additional)
+		// TODO talk to AWS about why there cannot be an explicit condition that signals AWS API call errors
```
kumare3, line 460: did we talk to AWS about this?

kumare3: don't we get this problem often, especially if we are having some permissions problem in the begining and then it works? But, I guess it is ok to start off. Maybe one thing we could do is, put a simple time, check, if we see this status, but its not been atleast a couple minutes then we ignore?

kumare3:
Looks good to me, one comment then we should be good to go!

copilot/go.mod
bnsblue: yes

go/tasks/plugins/k8s/sagemaker/sagemaker.go
bnsblue: @kumare3 from my experience during the past couple of weeks, every time I got this problem it was due to an actual error in the CRD. I can't recall if I had seen any error that was due to transient permission errors.

bnsblue:
💨

PR #101, opened 2020-06-25
Set Max Limit on number of executors to be launched in spark plugin

migueltol22:
# TL;DR
Allows configuration to set a max limit on spark executor by specifying `spark.executor.instances:<max_limit>` as part of the default spark configuration.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/lyft/flyte/issues/366

## Follow-up issue
_NA_
OR

### DIFF
--- a/go/tasks/plugins/k8s/spark/spark.go
+++ b/go/tasks/plugins/k8s/spark/spark.go
@@ -31,6 +31,8 @@ import (
 const KindSparkApplication = "SparkApplication"
 const sparkDriverUI = "sparkDriverUI"
 const sparkHistoryUI = "sparkHistoryUI"
+const sparkExecutorKey = "spark.executor.instances"
+const sparkExecutorLimitKey = "spark.executor.instances.limit"
 
 var sparkTaskType = "spark"
 
@@ -81,8 +83,8 @@ func min(a, b int) int {
 
 func populateSparkConfig(sparkConfig, userSparkConfig map[string]string) error {
 	for k, v := range userSparkConfig {
-		if k == "spark.executor.instances" && sparkConfig["spark.executor.instances"] != "" {
-			limit, err := strconv.Atoi(sparkConfig["spark.executor.instances"])
+		if k == sparkExecutorKey && sparkConfig[sparkExecutorLimitKey] != "" {
+			limit, err := strconv.Atoi(sparkConfig[sparkExecutorLimitKey])
 			if err != nil {
 				return err
 			}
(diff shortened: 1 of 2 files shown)


go/tasks/plugins/k8s/spark/spark.go
```diff
@@ -71,6 +72,33 @@ func validateSparkJob(sparkJob *plugins.SparkJob) error {
...
+func min(a, b int) int {
+	if a < b {
+		return a
+	}
+	return b
+}
+
+func populateSparkConfig(sparkConfig, userSparkConfig map[string]string) error {
+	for k, v := range userSparkConfig {
+		if k == "spark.executor.instances" && sparkConfig["spark.executor.instances"] != "" {
```
akhurana001, line 84: The current behavior of the config value is to be the default value if the users don't provide one. This will change its behavior as the max.

 Let's add a a separate section for limits ?

migueltol22: Updated with separate section.

go/tasks/plugins/k8s/spark/spark.go
```diff
@@ -30,13 +31,20 @@ import (
...
 // Spark-specific configs
 type Config struct {
 	DefaultSparkConfig    map[string]string `json:"spark-config-default" pflag:",Key value pairs of default spark configuration that should be applied to every SparkJob"`
 	SparkHistoryServerURL string            `json:"spark-history-server-url" pflag:",URL for SparkHistory Server that each job will publish the execution history to."`
+	SparkLimits           Limits            `json:"spark-limits" pflag:", Spark Limits that can be set for each job execution."`
+}
+
+// Spark Limits config
+type Limits struct {
+	ExecutorLimit string `json:"executor-limit" pflag:",Max limit on number of executors allowed to be launched."`
```
akhurana001, line 47: nit: would suggest renaming to ExecutorCountLimit or similar just to be more explicit in-case  if we want to add memory or cpu limits as well.

akhurana001:
looks good.

go/tasks/plugins/k8s/spark/spark.go
migueltol22: Updated.

kumare3:
@migueltol22 is this required and are you merging it?

PR #107, opened 2020-07-31
[Bug fix] SageMaker case sensitivity

bnsblue:
# TL;DR
A lot of the fields in SageMaker CRDs are case sensitive, but only some of them are validated at CRD level. The bug this PR fixed was detected by the operator.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
A lot of the fields in SageMaker CRDs are case sensitive, but only some of them are validated at CRD level. The bug this PR fixed was detected by the operator. We use `strings.Title(strings.ToLower(<protoc enum>.String()))` to fix it.

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/plugins/k8s/sagemaker/utils.go
+++ b/go/tasks/plugins/k8s/sagemaker/utils.go
@@ -120,6 +120,7 @@ func buildParameterRanges(hpoJobConfig *sagemakerSpec.HyperparameterTuningJobCon
 	}
 
 	for prName, pr := range prMap {
+		scalingTypeString := strings.Title(strings.ToLower(pr.GetContinuousParameterRange().GetScalingType().String()))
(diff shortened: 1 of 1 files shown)


kumare3:
Have we tested sufficient surface area, that this is not an issue with other enums? All Protobuf enums are capitalized, i would presume all of them would have a problem? no?

kumare3:
👍

bnsblue:
@kumare3 I’ve double checked that, except for the “InputContentType” enum which requires a explicit translation, all the other enums that I use in SageMaker plugin are wrapped inside the same strings.Title(strings.ToLower()) thing

PR #109, opened 2020-08-05
Add System Level Timeout

migueltol22:
# TL;DR
Handles system level timeouts in 2 cases, PodQueued and PhaseInitializing.

PodQueued is part of interruptible, where if a pod is stuck in pending for a time longer than the maxSystemLevelTimeout,  we can assume that there are not enough resources in the ASG and we will then error out as a system level error and retry on the non-interruptible ASG. 

PhaseInitializing is when some sort of system level error occurs that leaves us in this state. We will error with a system level error and retry launching the pod.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR
_https://github.com/lyft/flyte/issues/382


go/tasks/pluginmachinery/flytek8s/config/config.go
```diff
@@ -76,6 +76,8 @@ type K8sPluginConfig struct {
 	SchedulerName string `json:"scheduler-name" pflag:",Defines scheduler name."`
 	// Flyte CoPilot Configuration
 	CoPilot FlyteCoPilotConfig `json:"co-pilot" pflag:",Co-Pilot Configuration"`
+	// Set system level timeout. If timeout reached pod will be relaunched.
+	MaxSystemLevelTimeout int64 `json:"maxSystemLevelTimeout" pflag:",Value to be used for system level timeouts in minutes."`
```
EngHabu, line 80:
```suggestion
	MaxSystemLevelTimeout config.Duration `json:"maxSystemLevelTimeout" pflag:",Value to be used for system level timeouts in minutes."`
```

go/tasks/pluginmachinery/flytek8s/pod_helper.go
```diff
@@ -96,27 +96,54 @@ func BuildIdentityPod() *v1.Pod {
...
+	status := pod.Status
+	timeout := config.GetK8sPluginConfig().MaxSystemLevelTimeout
 	for _, c := range status.Conditions {
 		switch c.Type {
 		case v1.PodScheduled:
 			if c.Status == v1.ConditionFalse {
+				// If the pod is interruptible and is waiting to be scheduled for an extended amount of time,  it is possible there are
+				// no spot instances availabled in the AZ. In this case, we timeout with a system level error and will retry on a
+				// non spot instance AZ.
+				if val, ok := pod.ObjectMeta.Labels["interruptible"]; ok {
```
EngHabu, line 111: Please move the `interruptible` string into a const

go/tasks/pluginmachinery/flytek8s/pod_helper.go
```diff
@@ -96,27 +96,54 @@ func BuildIdentityPod() *v1.Pod {
...
+func DemystifyPending(pod *v1.Pod) (pluginsCore.PhaseInfo, error) {
 	// Search over the difference conditions in the status object.  Note that the 'Pending' this function is
 	// demystifying is the 'phase' of the pod status. This is different than the PodReady condition type also used below
+	status := pod.Status
+	timeout := config.GetK8sPluginConfig().MaxSystemLevelTimeout
 	for _, c := range status.Conditions {
 		switch c.Type {
 		case v1.PodScheduled:
 			if c.Status == v1.ConditionFalse {
+				// If the pod is interruptible and is waiting to be scheduled for an extended amount of time,  it is possible there are
```
EngHabu, line 108: ConditionFalse _I think_ means this condition is no longer the "current" but it'll never be removed from the list of conditions. I do not think we should apply the timeout in this case...

go/tasks/pluginmachinery/flytek8s/pod_helper.go
```diff
@@ -153,6 +180,17 @@ func DemystifyPending(status v1.PodStatus) (pluginsCore.PhaseInfo, error) {
 							finalMessage := fmt.Sprintf("%s|%s", c.Message, containerStatus.State.Waiting.Message)
 							switch reason {
 							case "ErrImagePull", "ContainerCreating", "PodInitializing":
```
EngHabu, line 182: Anand mentioned there are a couple more errors we should add here... 
@anandswaminathan can you remind me?

go/tasks/pluginmachinery/flytek8s/pod_helper.go
migueltol22: Yep took a look at the docs and it appears it means that the condition is not applicable. Removed.

go/tasks/pluginmachinery/flytek8s/pod_helper.go
migueltol22: done.

go/tasks/pluginmachinery/flytek8s/config/config.go
migueltol22: Done and updated checks.

go/tasks/pluginmachinery/flytek8s/pod_helper.go
```diff
@@ -108,15 +111,28 @@ func DemystifyPending(status v1.PodStatus) (pluginsCore.PhaseInfo, error) {
...
 			//  Pod status looks like this:
 			// 	message: '0/1 nodes are available: 1 Insufficient memory.'
 			//  reason: Unschedulable
 			// 	status: "False"
 			// 	type: PodScheduled
+
+			// If the pod is interruptible and is waiting to be scheduled for an extended amount of time,  it is possible there are
+			// no spot instances availabled in the AZ. In this case, we timeout with a system level error and will retry on a
+			// non spot instance AZ.
+			if val, ok := pod.ObjectMeta.Labels[Interruptible]; ok {
```
EngHabu, line 127: Should we add a check here for `if c.Status == v1.ConditionTrue` or is Unschedulable a terminal condition that once entered, the pod will never exit?

migueltol22: I found this comment for Unschedulable 
```	
// PodReasonUnschedulable reason in PodScheduled PodCondition means that the scheduler
// can't schedule the pod right now, for example due to insufficient resources in the cluster
```

However for the status that is being checked is the `ConditionStatus` and if it is in `ConditionTrue` then appears to mean that resource **is in** that condition.

```
// These are valid condition statuses. "ConditionTrue" means a resource is in the condition.
// "ConditionFalse" means a resource is not in the condition. "ConditionUnknown" means kubernetes
// can't decide if a resource is in the condition or not. In the future, we could add other
// intermediate conditions, e.g. ConditionDegraded.
```

migueltol22: Added the check.

go/tasks/pluginmachinery/flytek8s/pod_helper.go
anandswaminathan: Yes. We can follow up, should be easy to add

go/tasks/pluginmachinery/flytek8s/pod_helper.go
```diff
@@ -153,6 +169,17 @@ func DemystifyPending(status v1.PodStatus) (pluginsCore.PhaseInfo, error) {
 							finalMessage := fmt.Sprintf("%s|%s", c.Message, containerStatus.State.Waiting.Message)
 							switch reason {
 							case "ErrImagePull", "ContainerCreating", "PodInitializing":
+
+								// If we are in any of these states for an extended period of time there could be a system level error.
+								// To help mitigate the pod being stuck in this state we have a system level timeout that will error out
+								// as a system error and retry launching the pod.
+								if timeout > 0 && timeout > time.Since(status.StartTime.Time) {
+									return pluginsCore.PhaseInfoRetryableFailure(
```
anandswaminathan, line 177: Return system error? 

https://github.com/lyft/flyteplugins/blob/master/go/tasks/pluginmachinery/core/phase.go#L192

anandswaminathan:
cc @kumare3 as haytham is out.

go/tasks/pluginmachinery/flytek8s/config/config.go
```diff
@@ -77,7 +77,7 @@ type K8sPluginConfig struct {
 	// Flyte CoPilot Configuration
 	CoPilot FlyteCoPilotConfig `json:"co-pilot" pflag:",Co-Pilot Configuration"`
 	// Set system level timeout. If timeout reached pod will be relaunched.
-	MaxSystemLevelTimeout config2.Duration `json:"maxSystemLevelTimeout" pflag:"-,Value to be used for system level timeouts in minutes."`
+	MaxSystemLevelTimeout config2.Duration `json:"max-system-level-timeout" pflag:"-,Value to be used for system level timeouts in minutes."`
```
kumare3, line 80: can we improve the description. I do not follow the description actually

kumare3:
@migueltol22 are we reviving this?

PR #112, opened 2020-08-11
Adding retry policy to be available to the plugins

catalinii:
https://github.com/lyft/flyte/issues/469


go/tasks/pluginmachinery/core/exec_metadata.go
```diff
@@ -11,6 +12,7 @@ import (
 type TaskOverrides interface {
 	GetResources() *v1.ResourceRequirements
 	GetConfig() *v1.ConfigMap
+	GetRetryStrategy() *v1alpha1.RetryStrategy
```
kumare3, line 15: Please do not add references to the CRD. These should be references to the core Protobuf objects. This is a cyclic dependency

kumare3:
@catalinii this is incorrect. PluginMachinery only references the flyteIDL (proto) defs. the existence of CRD is unknown at this point

kumare3:
you want to return - https://sourcegraph.lyft.net/github.com/lyft/flyteidl@e146f4964c55d7fa7077e078aab62efe237997cb/-/blob/protos/flyteidl/core/literals.proto#L138:9

go/tasks/plugins/k8s/spark/spark.go
```diff
@@ -112,6 +113,11 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
 		sparkEnvVars[envVar.Name] = envVar.Value
 	}
 
+	if taskCtx.TaskExecutionMetadata().GetOverrides().GetRetryStrategy() != nil {
+		sparkEnvVars["FLYTE_MAX_ATTEMPTS"] = strconv.Itoa(int(taskCtx.TaskExecutionMetadata().GetOverrides().GetRetryStrategy().Retries))
```
kumare3, line 117: why do you need these env variables? Initially when we discussed we talked that you want metrics on retries. Adding env variables may add things to the contract. What is the goal of these variables?

kumare3: if we want to add env vars -- we should add it here, as this is the contract that will be maintained.
https://github.com/lyft/flyteplugins/blob/master/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go#L36,

catalinii: @kumare3 by exposing the retry number we can do canary based on retryNumber. I have updated the attemptNumber in k8s_resource_adds. The only left problem is related to maxattempt (which I do not think is yet implemented on the propeller/admin side).

kumare3: you are right Max attempt number is not available in the ID at the moment. @akhurana001 and you can work on that together?

kumare3 (approved):

go/tasks/plugins/k8s/spark/spark.go
catalinii: Will work the max attempt as part of a separate PR under the same task

PR #115, opened 2020-08-26
Raw output prefix

wild-endeavor:
# TL;DR
This will pipe through the raw output data prefix, as set by the [flytekit PR](https://github.com/lyft/flytekit/pull/146/files#diff-937f5bf70e8c655b5e2731a288bc3db8R17), through admin/propeller/propeller's node execution context, to container based plugins that call the replaceTemplateArgs.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
* Add regex, use regex to fill in the container args.
* Also accidentally ran `make generate` so some stuff got updated.

## Tracking Issue
https://github.com/lyft/flyte/issues/211

## Follow-up issue
NA

### DIFF
--- a/go/tasks/pluginmachinery/utils/template_test.go
+++ b/go/tasks/pluginmachinery/utils/template_test.go
@@ -42,11 +42,12 @@ func (d dummyInputReader) Get(ctx context.Context) (*core.LiteralMap, error) {
 }
 
 type dummyOutputPaths struct {
-	outputPath storage.DataReference
+	outputPath          storage.DataReference
+	rawOutputDataPrefix storage.DataReference
 }
 
 func (d dummyOutputPaths) GetRawOutputPrefix() storage.DataReference {
-	panic("should not be called")
+	return d.rawOutputDataPrefix
 }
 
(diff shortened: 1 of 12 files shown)


go/tasks/pluginmachinery/utils/template_test.go
```diff
@@ -338,4 +342,18 @@ func TestReplaceTemplateCommandArgs(t *testing.T) {
 			"output/blah",
 		}, actual)
 	})
+
+	t.Run("sub raw output data prefix", func(t *testing.T) {
```
kumare3, line 346: can we create a test, where we are replacing other things with raw-output prefix?

wild-endeavor: mmmm... yeah.  sorry does not compute.  can you elaborate?

wild-endeavor: done

PR #119, opened 2020-09-03
Revert "Revert "Presto - Reduce number of queries run""

wild-endeavor:
Reverts lyft/flyteplugins#99

https://github.com/lyft/flyte/issues/347


EngHabu:
Thank you! this is about *2.5x* improvement in presto query speeds if I'm reading it correctly.

EngHabu (approved):

PR #121, opened 2020-09-30
Introduce default plugin for task type

katrogan:
# TL;DR
Introduce default plugin for task type

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
Next step: update propeller plugin registration to properly register the default plugin handler for a task type

## Tracking Issue
https://github.com/lyft/flyte/issues/516

## Follow-up issue
_NA_


go/tasks/pluginmachinery/k8s/plugin.go
```diff
@@ -31,6 +31,10 @@ type PluginEntry struct {
 	// Boolean that indicates if this plugin can be used as the default for unknown task types. There can only be
 	// one default in the system
 	IsDefault bool
+	// A list of all task types for which this plugin should be default handler when multiple registered plugins
+	// support the same task type. This must be a subset of RegisteredTaskTypes and at most one default per task type
+	// is supported.
+	DefaultForTaskTypes []pluginsCore.TaskType
```
EngHabu, line 37: Should the choice of which plugin is the default of which type be defined as a config at the registry level?

katrogan: @EngHabu do you mean in RegisteredTaskTypes above?

katrogan: ping @EngHabu

katrogan:
friendly ping @EngHabu

EngHabu:
Ah so the way this will work is, each plugin will declare the types it can act as the default plugin for, then in Propeller's config, we can choose which task types each plugin is the default for?

So if pluginA said I can be the default for taskA & taskB and pluginB said it can be the default for taskB.
In Propeller's config, we can say:

- EnabledPlugins
   - pluginA:
      defaultFor:
      - taskA
      - taskC
   - pluginB:
      defaultFor:
      - taskB

This will complain about taskC (because it was never declared in pluginA?) but otherwise it'll be happy to accept the config... Did I get that right?

EngHabu (approved):

katrogan:
@EngHabu yes exactly - the gating in propeller ensures you can't declare yourself a default for a task type you don't support as defined in plugins

PR #124, opened 2020-10-06
Adding logic to insert the MPI-enabling hyperparameter in SageMaker plugin

bnsblue:
# TL;DR
This PR adds the logic in SageMaker plugins to read from the task spec and figure out if it should make SageMaker initialize MPI master and workers when executing the task.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [x] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
This PR adds the logic in SageMaker plugins to read from the task spec and figure out if it should make SageMaker initialize MPI master and workers when executing the task.

## Tracking Issue
https://github.com/lyft/flyte/issues/452

## Follow-up issue
_NA_


go/tasks/plugins/k8s/sagemaker/custom_training.go
```diff
@@ -94,6 +94,14 @@ func (m awsSagemakerPlugin) buildResourceForCustomTrainingJob(
 		Value: strconv.FormatBool(true),
 	})
 
+	// inject sagemaker_mpi_enabled into hyperparameters if the user code designates MPI as its distributed training framework
+	if sagemakerTrainingJob.GetTrainingJobResourceConfig().GetDistributedProtocol() == flyteSageMakerIdl.DistributedProtocol_MPI {
```
kumare3, line 98: can it be nil? or what is the default?

kumare3:
One comment, then looks good to me

go/tasks/plugins/k8s/sagemaker/custom_training.go
bnsblue: fixed

go/tasks/plugins/k8s/sagemaker/custom_training.go
```diff
@@ -94,6 +94,26 @@ func (m awsSagemakerPlugin) buildResourceForCustomTrainingJob(
 		Value: strconv.FormatBool(true),
 	})
 
+	if sagemakerTrainingJob.GetTrainingJobResourceConfig() == nil {
+		logger.Errorf(ctx, "TrainingJobResourceConfig is nil")
+		return nil, pluginErrors.Errorf(pluginErrors.BadTaskSpecification, "TrainingJobResourceConfig is nil")
+	}
+
+	if sagemakerTrainingJob.GetTrainingJobResourceConfig().GetDistributedProtocol() == flyteSageMakerIdl.DistributedProtocol_MPI && sagemakerTrainingJob.GetTrainingJobResourceConfig().GetInstanceCount() > 1 {
+		// inject sagemaker_mpi_enabled=true into hyperparameters if the user code designates MPI as its distributed training framework and the instance count is > 1
```
katrogan, line 103: what happens when sagemaker_mpi_enabled=true but instance count == 0? is this an error or is the fallback behavior (injecting sagemaker_mpi_enabled=false) intended?

bnsblue: @katrogan This particular scenario leads to an error captured by the SageMaker CRD's own validation with the following error message: 
```
spec.resourceConfig.instanceCount: Invalid value: 1: spec.resourceConfig.instanceCount in body should be greater than or equal to 1
```

katrogan: got it. do we want to explicitly pass along this misconfiguration error in this case or is the intention to just fall back to the false case?

bnsblue: @katrogan 

That's actually a good catch. I just found a document (https://aws.amazon.com/blogs/machine-learning/multi-gpu-and-distributed-training-using-horovod-in-amazon-sagemaker-pipe-mode/) saying that MPI works on a single instance with multiple GPUs. I also did an experiment and found that MPI can even work on a single instance with one GPU (although that means there's no benefit and even has overhead, it still "works" without an error)


Back to our discussion, so I guess what you are suggesting is we don't do this override on `sagemaker_mpi_enabled` and just let it be carried over to the CRD and let the CRD's validation take care of it?

katrogan: yeah exactly. i would prefer to explicitly report misconfiguration errors to users rather than silently impact behavior and possibly lead to confusion. but i also have very little context with sagemaker in general, so i'm not sure if it's a preferable user experience to mask the error and disable MPI

bnsblue: @katrogan Great suggestion. I've fixed it in the following commits. Please take another look

go/tasks/plugins/k8s/sagemaker/custom_training_test.go
```diff
@@ -74,6 +75,87 @@ func Test_awsSagemakerPlugin_BuildResourceForCustomTrainingJob(t *testing.T) {
...
+		taskTemplate := generateMockTrainingJobTaskTemplate("the job", tjObj)
+
+		trainingJobResource, err := awsSageMakerTrainingJobHandler.BuildResource(ctx, generateMockCustomTrainingJobTaskContext(taskTemplate, false))
+		assert.NoError(t, err)
+		assert.NotNil(t, trainingJobResource)
+
+		trainingJob, ok := trainingJobResource.(*trainingjobv1.TrainingJob)
+		assert.True(t, ok)
+
+		// Since the instance count is 1, even though the distributed protocol is MPI, we should find sagemaker_mpi_enabled=false in the hyperparameters
```
katrogan, line 114: is this comment still accurate?

katrogan (dismissed):

go/tasks/plugins/k8s/sagemaker/custom_training_test.go
bnsblue: goo catch!

katrogan (approved):

PR #128, opened 2020-10-13
Use flyte configuration defaults for sidecar pod spec

katrogan:
# TL;DR
Use flyte configuration defaults for sidecar pod spec.

I tried to be clever and use a patch strategy that doesn't require hand curation for every additional field we write a default for, but I was unsuccessful. I tried dumping pod specs to json and using a merge patch that way but it failed for list fields :( 

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
User reported (thank you @jeevb)

## Tracking Issue
https://github.com/lyft/flyte/issues/548

## Follow-up issue
_NA_

### DIFF
--- a/copilot/data/download.go
+++ b/copilot/data/download.go
@@ -25,7 +25,7 @@ import (
 
 type Downloader struct {
 	format core.DataLoadingConfig_LiteralMapFormat
-	store *storage.DataStore
+	store  *storage.DataStore
 	// TODO support download mode
 	mode core.IOStrategy_DownloadMode
 }
--- a/go/tasks/plugins/k8s/sidecar/sidecar.go
+++ b/go/tasks/plugins/k8s/sidecar/sidecar.go
@@ -4,8 +4,6 @@ import (
 	"context"
 	"fmt"
 
-	v1 "k8s.io/api/core/v1"
-
 	"github.com/lyft/flyteidl/gen/pb-go/flyteidl/plugins"
 
 	"github.com/lyft/flyteplugins/go/tasks/pluginmachinery"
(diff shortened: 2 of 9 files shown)


go/tasks/plugins/k8s/sidecar/sidecar.go
```diff
@@ -27,6 +27,23 @@ const (
 
 type sidecarResourceHandler struct{}
 
+func mergePodSpecs(flyteConfiguredPodSpec, userSpecified *v1.PodSpec) {
```
kumare3, line 30: merge seems weird and seems will get out of date everytime we change. What if we change this so that the merging is done by the flytek8s pod_helper.

so we can have a function
flytek8s.UpdatePod(basePod v1.PodSpec)?

katrogan: yeah I agree the manual merge is clunky (see PR comment, I couldn't successfully find a way to dynamically merge the flyte defaults w/ the user specified pod spec). updated to use your suggestion

katrogan:
PTAL @kumare3

katrogan:
@kumare3 I had to remove the `nvidia.com/gpu` resource toleration from the config because an the Interruptible test was complaining - I'm not sure what changed to make that an invalid name

copilot/data/download.go
```diff
@@ -25,7 +25,7 @@ import (
 
 type Downloader struct {
 	format core.DataLoadingConfig_LiteralMapFormat
-	store *storage.DataStore
+	store  *storage.DataStore
```
kumare3, line 28: Format?

kumare3 (approved):

copilot/data/download.go
katrogan: yep, i ran `make goimports` but I can revert though if it's annoying to mess with git blame

PR #130, opened 2020-10-15
Introduce a fail fast plugin

katrogan:
# TL;DR
Introduce a fail fast plugin which can be used as a default implementation for cases where Flyte deployments are not set-up to support certain tasks.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
In certain cases it would be useful to leverage configurable plugins as a whitelist mechanism without necessarily having a fallback implementation. In this case a fail-fast plugin implementation would be useful to have as the default plugin implementation for a task type and whitelisted projects/domains could be added for the actual plugin handler.

## Tracking Issue
https://github.com/lyft/flyte/issues/552

## Follow-up issue
_NA_


go/tasks/pluginmachinery/bundle/fail_fast.go
```diff
@@ -0,0 +1,58 @@
...
+func failFastPluginLoader(_ context.Context, _ core.SetupContext) (core.Plugin, error) {
+	return &failFastHandler{}, nil
+}
+
+func init() {
+	// TODO(katrogan): Once we move pluginmachinery to flyteidl make these task types named constants that flyteplugins
+	// can reference in other handler definitions.
+	// NOTE: these should match the constants defined flytekit
+	taskTypes := []core.TaskType{
+		"container", "sidecar", "container_array", "hive", "presto",
```
wild-endeavor, line 49: should we add everything?

wild-endeavor (approved):

go/tasks/pluginmachinery/bundle/fail_fast.go
katrogan: sure, wasn't sure what the state of sagemaker/pytorch was but I can add too

go/tasks/pluginmachinery/bundle/fail_fast.go
```diff
@@ -0,0 +1,59 @@
...
+	return failFastExecutorName
+}
+
+func (h failFastHandler) GetProperties() core.PluginProperties {
+	return core.PluginProperties{}
+}
+
+func (h failFastHandler) Handle(_ context.Context, _ core.TaskExecutionContext) (core.Transition, error) {
+	occuredAt := time.Now()
+	return core.DoTransition(core.PhaseInfoFailure("AlwaysFail",
```
kumare3, line 26: to the message can we add the task_type?

kumare3: Also other information like TaskID?

katrogan: I don't see that exposed in the [TaskExecutionContext](https://github.com/lyft/flyteplugins/blob/master/go/tasks/pluginmachinery/core/exec_metadata.go#L24) @kumare3 - should i add it or am I looking in the wrong place?

katrogan:
ping @kumare3

go/tasks/pluginmachinery/bundle/fail_fast.go
kumare3: https://github.com/lyft/flyteplugins/blob/master/go/tasks/pluginmachinery/core/exec_metadata.go#L28

kumare3 (approved):

PR #134, opened 2020-10-27
Update Spark Plugin to v1beta2

akhurana001:
Move spark plugin to  v1beta2 multi-version of Operator. Ref: https://github.com/lyft/flyte/issues/573

### DIFF
--- a/go/tasks/plugins/k8s/spark/spark.go
+++ b/go/tasks/plugins/k8s/spark/spark.go
@@ -17,7 +17,7 @@ import (
 	"github.com/lyft/flyteplugins/go/tasks/pluginmachinery/utils"
 	"k8s.io/client-go/kubernetes/scheme"
 
-	sparkOp "github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1beta1"
+	sparkOp "github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1beta2"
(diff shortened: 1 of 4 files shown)


go/tasks/plugins/k8s/spark/spark.go
```diff
@@ -17,7 +17,7 @@ import (
 	"github.com/lyft/flyteplugins/go/tasks/pluginmachinery/utils"
 	"k8s.io/client-go/kubernetes/scheme"
 
-	sparkOp "github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1beta1"
+	sparkOp "github.com/GoogleCloudPlatform/spark-on-k8s-operator/pkg/apis/sparkoperator.k8s.io/v1beta2"
```
kumare3, line 20: thats it no changes in the API?

kumare3:
When you do this, can you also change the oss flyte deployment to deploy the new version?

kumare3 (approved):

go/tasks/plugins/k8s/spark/spark.go
akhurana001: There are some change like types of some fields in the Spec has changed but we don't use that in flyteplugins i.e. only set those config via spark_conf. The  one other change is see is that status is now a sub-resource and the operator needs additional perms for the same.

akhurana001:
> When you do this, can you also change the oss flyte deployment to deploy the new version?

Will do in https://github.com/lyft/flyte/issues/573 once I create a new propeller release .

PR #138, opened 2020-11-06
Update Spark Plugin

akhurana001:
# TL;DR
SparkOperator v1beta2 expects Driver/Executor cores in the CRD else it defaults these values if they are not set which then takes precedence over spark_conf values effectively leading to no way of specifying these as part of spark_conf. This looks like an un-intended behavior in the operator but for now, let's add these values at both places to be safe. 

Ref: https://github.com/lyft/flyte/issues/585

Filed an Operator issue in https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/issues/1055 

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue
https://github.com/lyft/flyte/issues/585
## Follow-up issue
_NA_
OR

### DIFF
--- a/go/tasks/plugins/k8s/spark/spark.go
+++ b/go/tasks/plugins/k8s/spark/spark.go
@@ -169,6 +169,24 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
 	}
 	sparkConfig["spark.kubernetes.executor.podNamePrefix"] = taskCtx.TaskExecutionMetadata().GetTaskExecutionID().GetGeneratedName()
 
+	// Add driver/executor defaults to CRD Driver/Executor Spec as well.
+	cores, err := strconv.Atoi(sparkConfig["spark.driver.cores"])
+	if err == nil {
+		driverSpec.Cores = intPtr(int32(cores))
+	}
+	driverSpec.Memory = strPtr(sparkConfig["spark.driver.memory"])
+
+	execCores, err := strconv.Atoi(sparkConfig["spark.executor.cores"])
+	if err == nil {
+		executorSpec.Cores = intPtr(int32(execCores))
+	}
+
+	execCount, err := strconv.Atoi(sparkConfig["spark.executor.instances"])
+	if err == nil {
+		executorSpec.Instances = intPtr(int32(execCount))
+	}
+	executorSpec.Memory = strPtr(sparkConfig["spark.executor.memory"])
+
 	j := &sparkOp.SparkApplication{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       KindSparkApplication,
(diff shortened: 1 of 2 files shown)


go/tasks/plugins/k8s/spark/spark.go
```diff
@@ -169,6 +169,24 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
 	}
 	sparkConfig["spark.kubernetes.executor.podNamePrefix"] = taskCtx.TaskExecutionMetadata().GetTaskExecutionID().GetGeneratedName()
 
+	// Add driver/executor defaults to CRD Driver/Executor Spec as well.
+	cores, err := strconv.Atoi(sparkConfig["spark.driver.cores"])
```
kumare3, line 173: so there is no way to run flyte with older spark operator right? otherwise this will fail

kumare3: also will this make it possible to run Spark now in client mode?

kumare3 (approved):

go/tasks/plugins/k8s/spark/spark.go
akhurana001: The latest plugin version creates v1beta2 so will need this fix to work correctly. The older versions should continue to work.

Client mode  will need a more work to run the submit-job as a pod and set its resources correctly i.e. driver memory + cores .

go/tasks/plugins/k8s/spark/spark.go
```diff
@@ -169,6 +169,24 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
...
+	execCores, err := strconv.Atoi(sparkConfig["spark.executor.cores"])
+	if err == nil {
+		executorSpec.Cores = intPtr(int32(execCores))
+	}
+
+	execCount, err := strconv.Atoi(sparkConfig["spark.executor.instances"])
+	if err == nil {
+		executorSpec.Instances = intPtr(int32(execCount))
+	}
+	executorSpec.Memory = strPtr(sparkConfig["spark.executor.memory"])
```
EngHabu, line 188: What happens if it doesn't exist? It'll be a nil pointer? and that's an ok setting?

go/tasks/plugins/k8s/spark/spark_test.go
```diff
@@ -316,6 +317,19 @@ func TestBuildResourceSpark(t *testing.T) {
 	assert.Equal(t, sj.PythonApplicationType, sparkApp.Spec.Type)
 	assert.Equal(t, testArgs, sparkApp.Spec.Arguments)
 	assert.Equal(t, testImage, *sparkApp.Spec.Image)
+
```
EngHabu, line 320: Can you add tests for nil cases (when the ocnfig doesn't exist

PR #140, opened 2020-11-12
Fix log links in K8s Array tasks

anandswaminathan:
Currently Log links are not sent to Admin/UI for K8s batch array task due to a bug. Similarly there were other cases that were breaking.

* Log links for remote cluster execution.
* Log links for finished tasks
* Introduces finalizer for pods created by K8s array task.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

PTAL @EngHabu @katrogan

### DIFF
--- a/go/tasks/logs/logging_utils.go
+++ b/go/tasks/logs/logging_utils.go
@@ -9,13 +9,15 @@ import (
 	v1 "k8s.io/api/core/v1"
 )
 
-func GetLogsForContainerInPod(ctx context.Context, pod *v1.Pod, index uint32, nameSuffix string) ([]*core.TaskLog, error) {
+func GetLogsForContainerInPodCustomURL(ctx context.Context, pod *v1.Pod, index uint32, nameSuffix, kubernetesURL string) ([]*core.TaskLog, error) {
 	var logs []*core.TaskLog
 	logConfig := GetLogConfig()
 
 	logPlugins := map[string]logUtils.LogPlugin{}
 
-	if logConfig.IsKubernetesEnabled {
+	if kubernetesURL != "" {
+		logPlugins["Kubernetes Logs"] = logUtils.NewKubernetesLogPlugin(kubernetesURL)
+	} else if logConfig.IsKubernetesEnabled {
 		logPlugins["Kubernetes Logs"] = logUtils.NewKubernetesLogPlugin(logConfig.KubernetesURL)
 	}
 	if logConfig.IsCloudwatchEnabled {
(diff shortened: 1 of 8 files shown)


go/tasks/logs/logging_utils.go
```diff
@@ -9,13 +9,15 @@ import (
...
 
-func GetLogsForContainerInPod(ctx context.Context, pod *v1.Pod, index uint32, nameSuffix string) ([]*core.TaskLog, error) {
+func GetLogsForContainerInPodCustomURL(ctx context.Context, pod *v1.Pod, index uint32, nameSuffix, kubernetesURL string) ([]*core.TaskLog, error) {
 	var logs []*core.TaskLog
 	logConfig := GetLogConfig()
 
 	logPlugins := map[string]logUtils.LogPlugin{}
 
-	if logConfig.IsKubernetesEnabled {
+	if kubernetesURL != "" {
```
EngHabu, line 18:
```suggestion
	if len(kubernetesURL) > 0 {
```

go/tasks/plugins/array/k8s/task.go
```diff
@@ -82,6 +104,7 @@ func (t Task) Launch(ctx context.Context, tCtx core.TaskExecutionContext, kubeCl
 	pod = ApplyPodPolicies(ctx, t.Config, pod)
 	pod = applyNodeSelectorLabels(ctx, t.Config, pod)
 	pod = applyPodTolerations(ctx, t.Config, pod)
+	pod = addPodFinalizer(pod)
```
EngHabu, line 107: We should gate this on the same or similar config: https://github.com/lyft/flyteplugins/blob/master/go/tasks/pluginmachinery/flytek8s/config/config.go#L53

go/tasks/plugins/array/k8s/task.go
```diff
@@ -180,8 +234,26 @@ func (t Task) Finalize(ctx context.Context, tCtx core.TaskExecutionContext, kube
...
+	if err != nil {
+		if !k8serrors.IsNotFound(err) {
+			logger.Errorf(ctx, "Error fetching pod [%s] in Finalize [%s]", podName, err)
+			return err
+		}
+	} else {
+		pod = clearFinalizer(pod)
+		err := kubeClient.GetClient().Update(ctx, pod)
+		if err != nil {
+			logger.Errorf(ctx, "Error updating pod finalizer [%s] in Finalize [%s]", podName, err)
```
EngHabu, line 251: We should check for notFound here again... and not error if it's

katrogan:
LGTM

go/tasks/plugins/array/k8s/task.go
anandswaminathan: Why? If pod is found, how can it disappear before update ? I guess no harm in adding.

PR #141, opened 2020-11-17
Add flyte startup time to the spark configuration

catalinii:
# TL;DR
This sets an environment variable with the time the spark plugin creates the job

## Type
 - [ ] Bug Fix
 - [ x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Just defined an environment variable

## Tracking Issue
https://github.com/lyft/flyte/issues/618

### DIFF
--- a/go/tasks/plugins/k8s/spark/spark.go
+++ b/go/tasks/plugins/k8s/spark/spark.go
@@ -168,6 +168,7 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
 		sparkConfig["spark.kubernetes.executor.limit.cores"] = sparkConfig["spark.executor.cores"]
 	}
 	sparkConfig["spark.kubernetes.executor.podNamePrefix"] = taskCtx.TaskExecutionMetadata().GetTaskExecutionID().GetGeneratedName()
+	sparkConfig["spark.kubernetes.driverEnv.FLYTE_START_TIME"] = strconv.FormatInt(time.Now().UnixNano()/1000000, 10)
 
 	// Add driver/executor defaults to CRD Driver/Executor Spec as well.
 	cores, err := strconv.Atoi(sparkConfig["spark.driver.cores"])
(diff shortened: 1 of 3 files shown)


go/tasks/plugins/k8s/spark/spark.go
```diff
@@ -168,6 +168,7 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
 		sparkConfig["spark.kubernetes.executor.limit.cores"] = sparkConfig["spark.executor.cores"]
 	}
 	sparkConfig["spark.kubernetes.executor.podNamePrefix"] = taskCtx.TaskExecutionMetadata().GetTaskExecutionID().GetGeneratedName()
+	sparkConfig["spark.kubernetes.driverEnv.FLYTE_START_TIME"] = strconv.FormatInt(time.Now().UnixNano() / 1000000, 10)
```
akhurana001, line 171: @EngHabu @anandswaminathan do we have a start timestamp in metadata which can use here ?

anandswaminathan: But is that what he wants to use here? or just time crd was created?

catalinii: @anandswaminathan capturing end to end latency is needed to capture cudtomer feeling, we can add also spark operator separately.

anandswaminathan: @catalinii If that is the case, you need to follow anmol's comment, and try to use started_at of the node. 

Here you are only trying to capture time the crd would be created.

catalinii: Not seeing ExecutableNodeStatus in the plugins and was not able to identify the start time before that. If bring the node startup time into the plugins is too big of a deal we can leave it as it.

akhurana001: I believe @anandswaminathan mentioned  started_at in the node_status but that's not passed to the plugin: https://github.com/lyft/flytepropeller/blob/eca2accc887dda4fd9519333d69821f342209d27/pkg/apis/flyteworkflow/v1alpha1/node_status.go#L160 .. @anandswaminathan thoughts ?

anandswaminathan (approved):

PR #142, opened 2020-11-23
Add interpolation to Hive queries

wild-endeavor:
# TL;DR
Runs Hive queries through the interpolation engine so that things get filled in.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
This is a pretty small change:
* Run the Hive query through the interpolation engine.
* Introduces the task execution ID's generated name as something that can be used to fill in, with the key

        {{ .PerRetryUniqueKey }}

However, in unit testing, we found that we could not add the `tCtx` argument to the `ReplaceTemplateCommandArgs` function because it caused an import cycle.  I moved that function into it's own file in `core` but that caused another import cycle because the `core/mocks` package imports `core` itself for some reason.  Resolved by making a new `coreutils` package rather than re-implementing a mock, which was the other option.

## Tracking Issue

## Follow-up issue
NA


go/tasks/plugins/hive/execution_state.go
```diff
@@ -470,3 +479,53 @@ func IsNotYetSubmitted(e ExecutionState) bool {
...
+	}
+
+	externalLocation, err := tCtx.DataStore().ConstructReference(ctx, tCtx.OutputWriter().GetRawOutputPrefix(), "")
+	if err != nil {
+		logger.Errorf(ctx, "Error getting external location: [%s]", err)
+		return currentState, err
+	}
+
+	outputs := taskTemplate.Interface.Outputs.GetVariables()
+	if len(outputs) != 0 && len(outputs) != 1 {
```
EngHabu, line 500: Do you want to validate the LiteralType of the output?

go/tasks/pluginmachinery/coreutils/template.go
```diff
@@ -0,0 +1,55 @@
...
+// 		what to expect each literal type to be serialized as.
+// If a command isn't a valid template or failed to evaluate, it'll be returned as is.
+// NOTE: I wanted to do in-place replacement, until I realized that in-place replacement will alter the definition of the
+// graph. This is not desirable, as we may have to retry and in that case the replacement will not work and we want
+// to create a new location for outputs
+func ReplaceTemplateCommandArgs(ctx context.Context, tExecMeta core.TaskExecutionMetadata, command []string, in io.InputReader,
+	out io.OutputFilePaths) ([]string, error) {
+
+	var perRetryUniqueKey = tExecMeta.GetTaskExecutionID().GetGeneratedName()
+	perRetryUniqueKey = startsWithAlpha.ReplaceAllString(perRetryUniqueKey, "a")
```
kumare3, line 33: Lets add a TODO here to make the generatedID itself follow some of these rules

go/tasks/pluginmachinery/coreutils/template.go
```diff
@@ -0,0 +1,55 @@
+package coreutils
```
kumare3, line 1: can we not move all files here? this looks a little off?

go/tasks/plugins/hive/execution_state.go
```diff
@@ -27,7 +31,7 @@ const (
 	PhaseNotStarted ExecutionPhase = iota
 	PhaseQueued                    // resource manager token gotten
 	PhaseSubmitted                 // Sent off to Qubole
-
+	PhaseWriteOutputFile
```
kumare3, line 34: Why are we adding this? if so we might need to add a transformer in Propeller - https://github.com/lyft/flytepropeller/blob/master/pkg/controller/nodes/task/transformer.go#L41

go/tasks/plugins/hive/execution_state.go
```diff
@@ -84,6 +90,9 @@ func HandleExecutionState(ctx context.Context, tCtx core.TaskExecutionContext, c
 	case PhaseSubmitted:
 		newState, transformError = MonitorQuery(ctx, tCtx, currentState, executionsCache)
 
+	case PhaseWriteOutputFile:
```
kumare3, line 93: why is this a separate phase?

go/tasks/pluginmachinery/coreutils/template.go
kumare3: I definitely want to do this, but in the past this replacement has been very expensive. With extremely large queries that may happen again, I think at-least a TODO that will measure the time it takes to replace

go/tasks/plugins/hive/execution_state.go
wild-endeavor: The bit of code that monitors the query status currently is done inside the cache handler, that just runs in goroutines in the background.  For something like writing to S3, it made more sense to make it an explicit plugin state.

kumare3 (approved):
lgtm

PR #146, opened 2021-01-21
Introduce WebAPI Plugin Interface, an example & AWS athena plugin

EngHabu:
# TL;DR
Introduce WebAPI Interface, backing structures and an example plugin that executes launchplans.

image

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [X] Plugin

## Are all requirements met?

 - [X] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
This PR increases the types of plugins users can write to add a WebAPI plugin. These are plugins that executes workflows by calling a remote service and periodically monitoring the progress of the task.

## Tracking Issue


go/tasks/pluginmachinery/core/resource_manager.go
```diff
@@ -4,21 +4,23 @@ import (
 	"context"
 )
 
-type AllocationStatus string
+//go:generate enumer -type=AllocationStatus -trimprefix=AllocationStatus
+
+type AllocationStatus int
```
kumare3, line 9: this would be a breaking change right?

EngHabu: No, this is not serialized anywhere... just a return type for the resource manager... I checked

go/tasks/aws/config.go
```diff
@@ -18,22 +21,21 @@ const ConfigSectionKey = "aws"
...
-	AccountID            string          `json:"accountId" pflag:",AWS Account Identifier."`
-	Retries              int             `json:"retries" pflag:",Number of retries."`
-	MaxErrorStringLength int             `json:"maxErrorLength" pflag:",Maximum size of error messages."`
-	CatalogCacheTimeout  config.Duration `json:"catalog-timeout" pflag:"\"5s\",Timeout duration for checking catalog for all batch tasks"`
+	Region               string            `json:"region" pflag:",AWS Region to connect to."`
+	AccountID            string            `json:"accountId" pflag:",AWS Account Identifier."`
+	Retries              int               `json:"retries" pflag:",Number of retries."`
+	LogLevel             aws.ClientLogMode `json:"logLevel" pflag:"-,Defines the Sdk Log Level."`
+	MaxErrorStringLength int               `json:"maxErrorLength" pflag:",Maximum size of error messages."`
+	CatalogCacheTimeout  config.Duration   `json:"catalog-timeout" pflag:"\"5s\",Timeout duration for checking catalog for all batch tasks"`
```
kumare3, line 38: this is interesting that we add catalogcache timeout in aws config no?

go/tasks/pluginmachinery/internal/webapi/allocation_token.go
```diff
@@ -0,0 +1,95 @@
...
+
+func SetClockForTest(clck clock2.Clock) {
+	clock = clck
+}
+
+func init() {
+	clock = clock2.RealClock{}
+}
+
+func allocateToken(ctx context.Context, p webapi.AsyncPlugin, tCtx core.TaskExecutionContext, state *State, metrics Metrics) (
```
kumare3, line 29: should these be bound to a struct? In that case clock could be a property of the struct. The module var called clock looks a little off IMO (subjective). I am basically vary of too many inits

go/tasks/pluginmachinery/internal/webapi/allocation_token.go
```diff
@@ -0,0 +1,95 @@
...
+		logger.Infof(ctx, "Couldn't allocate token because allocation status is [%v].", allocationStatus.String())
+		startTime := state.AllocationTokenRequestStartTime
+		if startTime.IsZero() {
+			startTime = clock.Now()
+		}
+
+		return &State{
+			AllocationTokenRequestStartTime: startTime,
+			Phase:                           PhaseNotStarted,
+		}, core.PhaseInfoQueued(time.Now(), 0, "Allocation tokens exhausted"), nil
```
kumare3, line 72: Lets have a better message here too? 
`Quota for task has exceeded. Max [%d]. The request is enqueued.`

go/tasks/pluginmachinery/internal/webapi/allocation_token.go
```diff
@@ -0,0 +1,95 @@
...
+	}
+
+	return nil, core.PhaseInfo{}, fmt.Errorf("allocation status undefined [%v]", allocationStatus)
+}
+
+func releaseToken(ctx context.Context, p webapi.AsyncPlugin, tCtx core.TaskExecutionContext, metrics Metrics) error {
+	ns, _, err := p.ResourceRequirements(ctx, tCtx)
+	if err != nil {
+		logger.Errorf(ctx, "Failed to calculate resource requirements for task. Error: %v", err)
+		return err
```
kumare3, line 82: this should never really happen right?

go/tasks/pluginmachinery/internal/webapi/cache.go
```diff
@@ -0,0 +1,178 @@
...
+			ID:     resource.GetID(),
+			Item:   cacheItem,
+			Action: cache.Update,
+		})
+	}
+
+	return resp, nil
+}
+
+// We need some way to translate results we get from Client, into a plugin phase
```
kumare3, line 129: lets define what plugin phase means in this context

go/tasks/pluginmachinery/internal/webapi/core.go
```diff
@@ -0,0 +1,212 @@
...
+
+func (c CorePlugin) Finalize(ctx context.Context, tCtx core.TaskExecutionContext) error {
+	if len(c.p.GetConfig().ResourceQuotas) == 0 {
+		// If there are no defined quotas, there is nothing to cleanup.
+		return nil
+	}
+
+	incomingState := State{}
+	if _, err := tCtx.PluginStateReader().Get(&incomingState); err != nil {
+		logger.Errorf(ctx, "AsyncPlugin %s failed to unmarshal custom state when handling [%s]. Error: %v",
```
kumare3, line 119: probably a metric?

go/tasks/pluginmachinery/internal/webapi/launcher.go
```diff
@@ -0,0 +1,53 @@
...
+		}
+	}
+
+	state.ResourceMeta = rMeta
+	state.Phase = PhaseResourcesCreated
+	cacheItem := CacheItem{
+		State: *state,
+	}
+
+	// The first time we put it in the cache, we know it won't have succeeded so we don't need to look at it
```
kumare3, line 45: is this because this is an async webapi? lets clarify that in the comment

go/tasks/plugins/webapi/athena/plugin.go
```diff
@@ -0,0 +1,212 @@
...
+	timeNow := time.Now()
+	return &core.TaskInfo{
+		OccurredAt: &timeNow,
+		Logs: []*idlCore.TaskLog{
+			{
+				Uri: fmt.Sprintf("https://%v.console.aws.amazon.com/athena/home?force&region=%v#query/history/%v",
+					cfg.Region,
+					cfg.Region,
+					queryID),
+				Name: "Athena Query History",
```
kumare3, line 184: Should this `Athena Query Console` or something?

go/tasks/plugins/webapi/athena/utils.go
```diff
@@ -0,0 +1,143 @@
...
+
+func validatePrestoQuery(prestoQuery pluginsIdl.PrestoQuery) error {
+	if len(prestoQuery.Statement) == 0 {
+		return errors.Errorf(errors.BadTaskSpecification, "Statement is a required field.")
+	}
+
+	return nil
+}
+
+func extractQueryInfo(ctx context.Context, tCtx webapi.TaskExecutionContextReader) (QueryInfo, error) {
```
kumare3, line 81: do we want to do it this way, of is this just to get going with athena and we will add a proper plugin?

go/tasks/pluginmachinery/webapi/example/plugin.go
```diff
@@ -0,0 +1,221 @@
+package example
```
kumare3, line 1: do we want to keep the admin example now that we have Athena?

kumare3:
This looks awesome, some minor nits!

go/tasks/aws/config.go
EngHabu: This is nonsense :-D I don't know when did this slip in.. it existed before this PR... it's not used/referenced anywhere... I meant to delete it...

go/tasks/pluginmachinery/internal/webapi/allocation_token.go
EngHabu: We do not currently expose the capacities... I think the whole resource management need a revisit...

go/tasks/pluginmachinery/internal/webapi/allocation_token.go
EngHabu: It's a call to the plugin... so maybe?

go/tasks/plugins/webapi/athena/utils.go
EngHabu: Depending on how we land on the SQL Abstraction, we might have a more generic Sql Plugin with engine (and maybe engine-specific properties)... or continue to use Presto here... Athena is based on Presto and should be able to execute presto queries just fine...

kumare3: ok, btw where are we landing on that? I would recommend looking at getdbt for inspiration

kumare3:
LGTM

kumare3 (approved):

anandswaminathan (approved):

PR #153, opened 2021-03-02
Inject and Use values from Security Context

anandswaminathan:
This change is related to deprecation on auth from Flyte and injecting SecurityContext.

- Ensured Backward compatibility
- Spark will use the serviceAccount from SecurityContext, if not default to "spark"
- AWS Array array tasks will check the securityContext before defaulting to iam_role.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

@EngHabu @kumare3


go/tasks/plugins/array/awsbatch/job_definition.go
```diff
@@ -50,7 +50,12 @@ func EnsureJobDefinition(ctx context.Context, tCtx pluginCore.TaskExecutionConte
 		return nil, errors.Errorf(pluginErrors.BadTaskSpecification, "Tasktemplate does not contain a container image.")
 	}
 
-	role := awsUtils.GetRole(ctx, cfg.RoleAnnotationKey, tCtx.TaskExecutionMetadata().GetAnnotations())
+	role := awsUtils.GetRole(ctx, cfg.RoleSecurityContextKey, tCtx.TaskExecutionMetadata().GetSecurityContext())
```
kumare3, line 53: should we make this into a protobuf enum - `cfg.RoleSecurityContextKey`

go/tasks/plugins/k8s/container/container.go
```diff
@@ -62,8 +62,15 @@ func (Plugin) BuildResource(ctx context.Context, taskCtx pluginsCore.TaskExecuti
 
 	pod := flytek8s.BuildPodWithSpec(podSpec)
 
-	// We want to Also update the serviceAccount to the serviceaccount of the workflow
-	pod.Spec.ServiceAccountName = taskCtx.TaskExecutionMetadata().GetK8sServiceAccount()
+	serviceAccountName := flytek8s.GetServiceAccountNameFromSecurityContext(taskCtx.TaskExecutionMetadata().GetSecurityContext())
```
kumare3, line 65: so in the previous case, we have a key-value pair, here it seems like an enum.
I think a cleaner interface would be to always make it enum, but for non supported enums, have a string?

kumare3 (commented):
I think you have missed out on sagemaker

go/tasks/plugins/array/awsbatch/job_definition.go
anandswaminathan: I am not inclined for this one. But for serviceAccountName may be? What do you think ?

go/tasks/plugins/k8s/container/container.go
anandswaminathan: @kumare3 The question is how to add enums. In FlyteIDL, any securitycontext stuff would then need to be added everywhere to be plumbed.

go/tasks/plugins/array/awsbatch/job_definition.go
kumare3: Why not for Role - AWS Batch, Sagemaker, Athena etc may all need a role, then how do we make it consistent

anandswaminathan: done

go/tasks/plugins/k8s/container/container.go
anandswaminathan: Chatted offline

anandswaminathan:
PTAL @akhurana001 This is a backward incompatible for SPARK - right fix.

If someone passes the ServiceAccount. Spark plugin has to respect it.

akhurana001:
> PTAL @akhurana001 This is a backward incompatible for SPARK - right fix.
> 
> If someone passes the ServiceAccount. Spark plugin has to respect it.

I am still wrapping my head around why users of the platform need to know about service accounts or K8s concepts.

How are you planning to role this out if this is backward incompatible ? Can we default to `spark` if not set ?

anandswaminathan:
@akhurana001 Let me clarify
- So it's backward compatible for most of folks. Default is still Spark.
- Only affects folks who are using ServiceAccount field in flyteidl, but still expect default spark to be used will be affected.
cc @kumare3

go.sum
```diff
@@ -228,6 +228,8 @@ github.com/fatih/color v1.10.0/go.mod h1:ELkj/draVOlAH/xkhN6mQ50Qd0MPOk5AAr3maGE
 github.com/fatih/structtag v1.2.0/go.mod h1:mBJUNpUnHmRKrKlQQlmCrh5PuhftFbNv8Ys4/aAZl94=
 github.com/flyteorg/flyteidl v0.18.15 h1:sXrlwTRaRjQsXYMNrY/S930SKdKtu4XnpNFEu8I4tn4=
 github.com/flyteorg/flyteidl v0.18.15/go.mod h1:b5Fq4Z8a5b0mF6pEwTd48ufvikUGVkWSjZiMT0ZtqKI=
+github.com/flyteorg/flyteidl v0.18.16 h1:84IHIuhWvmJ6j/bAPOlA3AT8yd063joTjIoKcUKCD1I=
```
EngHabu, line 231: `go mod tidy` plz

go/tasks/plugins/k8s/sagemaker/builtin_training.go
```diff
@@ -116,8 +116,14 @@ func (m awsSagemakerPlugin) buildResourceForTrainingJob(
 
 	inputModeString := strings.Title(strings.ToLower(sagemakerTrainingJob.GetAlgorithmSpecification().GetInputMode().String()))
 
-	role := awsUtils.GetRole(ctx, cfg.RoleAnnotationKey, taskCtx.TaskExecutionMetadata().GetAnnotations())
-	if role == "" {
+	role := awsUtils.GetRoleFromSecurityContext(taskCtx.TaskExecutionMetadata().GetSecurityContext())
+
+	// Continue this for backward compatibility
+	if len(role) == 0 {
```
EngHabu, line 122: We should probably encapsulate the backward compat code in one place... so we can deprecate it easier and avoid new plugins having to know to do that...

go/tasks/plugins/k8s/sagemaker/builtin_training.go
```diff
@@ -116,8 +116,14 @@ func (m awsSagemakerPlugin) buildResourceForTrainingJob(
 
 	inputModeString := strings.Title(strings.ToLower(sagemakerTrainingJob.GetAlgorithmSpecification().GetInputMode().String()))
 
-	role := awsUtils.GetRole(ctx, cfg.RoleAnnotationKey, taskCtx.TaskExecutionMetadata().GetAnnotations())
-	if role == "" {
+	role := awsUtils.GetRoleFromSecurityContext(taskCtx.TaskExecutionMetadata().GetSecurityContext())
+
```
akhurana001, line 120: nit: helper method since looks like this is being repetitive ?

akhurana001:
LGTM , one minor comment. Also can we test this if not already :)

go/tasks/plugins/k8s/sagemaker/builtin_training.go
anandswaminathan: Sure

go/tasks/plugins/k8s/sagemaker/builtin_training.go
anandswaminathan: Cool.

EngHabu (approved):

EngHabu (approved):

PR #158, opened 2021-03-06
Add pluginmachinery k8s KubeClient

regadas:
## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

### DIFF
--- a/go/tasks/pluginmachinery/k8s/client.go
+++ b/go/tasks/pluginmachinery/k8s/client.go
@@ -56,36 +56,36 @@ func (c fallbackClientReader) List(ctx context.Context, list client.ObjectList, 
 }
 
 type fallbackClientBuilder struct {
- 	uncached []client.Object
- }
+	uncached []client.Object
+}
 
- func (f *fallbackClientBuilder) WithUncached(objs ...client.Object) cluster.ClientBuilder {
- 	f.uncached = append(f.uncached, objs...)
- 	return f
- }
+func (f *fallbackClientBuilder) WithUncached(objs ...client.Object) cluster.ClientBuilder {
+	f.uncached = append(f.uncached, objs...)
+	return f
+}
 
- func (f fallbackClientBuilder) Build(cache cache.Cache, config *rest.Config, options client.Options) (client.Client, error) {
- 	c, err := client.New(config, options)
- 	if err != nil {
- 		return nil, err
- 	}
+func (f fallbackClientBuilder) Build(cache cache.Cache, config *rest.Config, options client.Options) (client.Client, error) {
+	c, err := client.New(config, options)
+	if err != nil {
+		return nil, err
+	}
 
- 	return client.NewDelegatingClient(client.NewDelegatingClientInput{
- 		Client: c,
- 		CacheReader: fallbackClientReader{
- 			orderedClients: []client.Reader{cache, c},
- 		},
- 		UncachedObjects: f.uncached,
- 		// TODO figure out if this should be true?
- 		// CacheUnstructured: true,
- 	})
- }
+	return client.NewDelegatingClient(client.NewDelegatingClientInput{
+		Client: c,
+		CacheReader: fallbackClientReader{
+			orderedClients: []client.Reader{cache, c},
+		},
+		UncachedObjects: f.uncached,
+		// TODO figure out if this should be true?
+		// CacheUnstructured: true,
+	})
+}
 
 // Creates a new k8s client that uses the cached client for reads and falls back to making API
 // calls if it failed. Write calls will always go to raw client directly.
 func NewFallbackClientBuilder() cluster.ClientBuilder {
- 	return &fallbackClientBuilder{}
- }
+	return &fallbackClientBuilder{}
+}
 
 type Options struct {
 	MapperProvider func(*rest.Config) (meta.RESTMapper, error)
--- a/go/tasks/pluginmachinery/k8s/client.go
+++ b/go/tasks/pluginmachinery/k8s/client.go
@@ -1,0 +1,133 @@
+package k8s
+
+import (
+	"context"
+
+	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/core"
+
+	"k8s.io/apimachinery/pkg/api/meta"
+	"k8s.io/client-go/rest"
+
+	"sigs.k8s.io/controller-runtime/pkg/cache"
+	"sigs.k8s.io/controller-runtime/pkg/client"
+	"sigs.k8s.io/controller-runtime/pkg/client/apiutil"
+	"sigs.k8s.io/controller-runtime/pkg/cluster"
+)
+
+type kubeClient struct {
+	client client.Client
+	cache  cache.Cache
+}
+
+func (k *kubeClient) GetClient() client.Client {
+	return k.client
+}
+
+func (k *kubeClient) GetCache() cache.Cache {
+	return k.cache
+}
+
+func newKubeClient(c client.Client, cache cache.Cache) core.KubeClient {
+	return &kubeClient{client: c, cache: cache}
+}
+
+type fallbackClientReader struct {
+	orderedClients []client.Reader
+}
+
+func (c fallbackClientReader) Get(ctx context.Context, key client.ObjectKey, out client.Object) (err error) {
+	for _, k8sClient := range c.orderedClients {
+		if err = k8sClient.Get(ctx, key, out); err == nil {
+			return nil
+		}
+	}
+
+	return
+}
+
+func (c fallbackClientReader) List(ctx context.Context, list client.ObjectList, opts ...client.ListOption) (err error) {
+	for _, k8sClient := range c.orderedClients {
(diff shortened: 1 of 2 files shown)


go/tasks/pluginmachinery/k8s/config.go
```diff
@@ -0,0 +1,58 @@
...
+	"io/ioutil"
+
+	"github.com/pkg/errors"
+	restclient "k8s.io/client-go/rest"
+)
+
+type ClusterConfig struct {
+	Name     string `json:"name" pflag:",Friendly name of the remote cluster"`
+	Endpoint string `json:"endpoint" pflag:", Remote K8s cluster endpoint"`
+	Auth     Auth   `json:"auth" pflag:"-, Auth setting for the cluster"`
```
EngHabu, line 14:
```suggestion
	Auth     Auth   `json:"auth" pflag:", Auth setting for the cluster"`
```

go/tasks/pluginmachinery/k8s/config.go
```diff
@@ -0,0 +1,58 @@
...
+	Enabled  bool   `json:"enabled" pflag:", Boolean flag to enable or disable"`
+}
+
+type Auth struct {
+	Type      string `json:"type" pflag:", Authentication type"`
```
EngHabu, line 19: What are the possible values? or maybe a link to the docs that list the values?
```suggestion
	Type      string `json:"type" pflag:",Authentication type"`
```

go/tasks/pluginmachinery/k8s/config.go
```diff
@@ -0,0 +1,58 @@
...
+	TokenPath string `json:"token-path" pflag:", Token path"`
+	CertPath  string `json:"cert-path" pflag:", Certificate path"`
```
EngHabu, line 21: I know it's not consistent across the codebase, but I'm trying to make new configs at least be camelCase
```suggestion
	TokenPath string `json:"tokenPath" pflag:",Token path"`
	CACertPath  string `json:"caCertPath" pflag:",Certificate Authority path"`
```

go/tasks/pluginmachinery/k8s/config.go
```diff
@@ -0,0 +1,58 @@
...
+func (auth *Auth) GetToken() (string, error) {
+	token, err := ioutil.ReadFile(auth.TokenPath)
+	if err != nil {
+		return "", errors.Wrap(err, "failed to read k8s bearer token from configured path")
+	}
+	return string(token), nil
+}
+
+func NewK8sClusterConfig(host string, auth Auth) (*restclient.Config, error) {
+	tokenString, err := auth.GetToken()
```
EngHabu, line 41: I might be missing it but I don't see us referencing auth.Type...

go/tasks/pluginmachinery/k8s/config.go
```diff
@@ -0,0 +1,58 @@
...
+type ClusterConfig struct {
+	Name     string `json:"name" pflag:",Friendly name of the remote cluster"`
+	Endpoint string `json:"endpoint" pflag:", Remote K8s cluster endpoint"`
+	Auth     Auth   `json:"auth" pflag:"-, Auth setting for the cluster"`
+	Enabled  bool   `json:"enabled" pflag:", Boolean flag to enable or disable"`
+}
+
+type Auth struct {
+	Type      string `json:"type" pflag:", Authentication type"`
+	TokenPath string `json:"token-path" pflag:", Token path"`
```
EngHabu, line 20: How long are these tokens valid for? don't you need a mechanism to renew them? assume Propeller is up and running for 3 months straight will the same token continue to work?

go/tasks/pluginmachinery/k8s/config.go
```diff
@@ -0,0 +1,58 @@
...
+	if err != nil {
+		return nil, errors.New(fmt.Sprintf("Failed to get auth token: %+v", err))
+	}
+
+	caCert, err := auth.GetCA()
+	if err != nil {
+		return nil, errors.New(fmt.Sprintf("Failed to get auth CA: %+v", err))
+	}
+
+	tlsClientConfig := restclient.TLSClientConfig{}
```
EngHabu, line 51: is NOT providing a CA cert a valid config? if so, should you or should you not create a TLS client config? should that be an option `InsecureConnection bool` ?

go/tasks/pluginmachinery/k8s/client.go
```diff
@@ -0,0 +1,122 @@
...
+		client.DelegatingClient{
+			Reader:       &client.DelegatingReader{CacheReader: cache, ClientReader: rawClient},
+			Writer:       rawClient,
+			StatusClient: rawClient,
+		}, rawClient)
+
+	return newKubeClient(fallbackClient, cache), nil
+}
+
+func NewDefaultKubeClient(config ClusterConfig) (core.KubeClient, error) {
```
EngHabu, line 120: nit: docs for the public types/funcs

go/tasks/pluginmachinery/k8s/config.go
regadas: I believe we can. I was keeping it simple since users supplying this config will have the token and ca cert at their disposal. This kinda avoids users making mistakes as insecure connection should only be used for testing.

Should we go ahead and add it anyway?

go/tasks/pluginmachinery/k8s/config.go
EngHabu: aha, make sense... let's remove it then as you said... addition is backward compat, removing/changing type... etc. isn't...

go/tasks/pluginmachinery/k8s/config.go
EngHabu: 👍

go/tasks/pluginmachinery/k8s/config.go
EngHabu: If you are going over the internet here... like exposing ApiServer through LoadBalancer... etc. etc. you will probably not use the self-signed CA cert and use a CloudProvider-provided Cert... will that still work? I imagine it's common to terminate SSL at ingress point in the remote cluster and let traffic be unauthenticated after... or maybe not... 

I'm always wary (as you can tell) when it comes to security, so I really appreciate you going through the exercise to figure out the right way to do this...

Another option is to just expose the ClientConfig in our config object and let people specify all the params they need... not sure how bad that can be...

regadas: Is not clear to me yet what other usages are sensible/common for these cases. That said, one thing we can do is just drop this config and have `NewKubeClient` have the `restclient.Config`.

The downside is that plugin creators might create different remote cluster configs according to their use case.

regadas: >The downside is that plugin creators might create different remote cluster configs according to their use case.

actually, when it comes to the usage of this in plugins it will be up to the user to define the `restclient.Config` for their use case (since they will supply the func that actually creates the KubeClient). So having `restclient.Config` being passed to `NewKubeClient` is perfectly fine because 1) this is a very advanced thing 2) setups can vary a lot.

EngHabu: Fair... for plugin developers, they can add restclient.Config as a config option and it should be parsed correctly so they can just do `NewKubeClient(GetConfig().KubeClient)` or something...

go/tasks/pluginmachinery/k8s/client.go
```diff
@@ -0,0 +1,133 @@
...
+ 		// TODO figure out if this should be true?
+ 		// CacheUnstructured: true,
+ 	})
+ }
+
+// Creates a new k8s client that uses the cached client for reads and falls back to making API
+// calls if it failed. Write calls will always go to raw client directly.
+func NewFallbackClientBuilder() cluster.ClientBuilder {
+ 	return &fallbackClientBuilder{}
+ }
```
regadas, line 88: Taken from https://github.com/flyteorg/flytepropeller/blob/lyft-to-flyteorg/pkg/controller/executors/kube.go as I believe the idea will be to move that code here.

go/tasks/pluginmachinery/k8s/testdata/config.yaml
```diff
@@ -0,0 +1,8 @@
+example:
```
EngHabu, line 1: Do we still need this config?

EngHabu (dismissed):

go/tasks/pluginmachinery/k8s/client.go
```diff
@@ -0,0 +1,133 @@
+package k8s
```
kumare3, line 1: Lets add a comment to the top of the file, that will automatically get generated as a godoc :). I know it did not exist, but been trying to do it everytime i touch a file

kumare3 (approved):
lgtm - one nit. Also you will remove the client from flytepropeller right?

regadas:
@kumare3 yup that's the idea.

kumare3 (approved):
Awesome!

PR #165, opened 2021-03-20
Pass along external resource IDs in task event execution metadata

katrogan:
Signed-off-by: Katrina Rogan <katroganGH@gmail.com>

# TL;DR
Add external resource id info to TaskExecutionMetadata

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/325

## Follow-up issue
_NA_


go/tasks/pluginmachinery/internal/webapi/allocation_token.go
```diff
@@ -23,27 +23,37 @@ func newTokenAllocator(c clock.Clock) tokenAllocator {
 	}
 }
 
+type TokenDetails struct {
```
kumare3, line 26: what is this structure for? 
It seems it is associated with the PhaseInfo, if so, why can we not add it to PhaseInfo?

go/tasks/plugins/webapi/athena/plugin.go
```diff
@@ -184,6 +188,14 @@ func createTaskInfo(queryID string, cfg awsSdk.Config) *core.TaskInfo {
 				Name: "Athena Query Console",
 			},
 		},
+		Metadata: &event.TaskExecutionMetadata{
+			PluginIdentifier: athenaPluginName,
```
kumare3, line 192: I think the PluginIdentifier should be injected by the TaskHandler itself in FlytePropeller. This is error prone

kumare3: if not in propeller then the WebAPI plugin, PluginManager should inject it. But the simplest would be to inject it in the transformer for task/handler.go

go/tasks/pluginmachinery/internal/webapi/allocation_token.go
katrogan: it's specific to the webapi token allocator, wan't sure it was worth broadly making accessible

go/tasks/plugins/webapi/athena/plugin.go
katrogan: done, added to propeller

go/tasks/plugins/hive/execution_state.go
```diff
@@ -143,14 +148,31 @@ func ConstructTaskLog(e ExecutionState) *idlCore.TaskLog {
 	}
 }
 
-func ConstructTaskInfo(e ExecutionState) *core.TaskInfo {
+func ConstructTaskInfo(tCtx core.TaskExecutionContext, e ExecutionState) *core.TaskInfo {
 	logs := make([]*idlCore.TaskLog, 0, 1)
 	t := time.Now()
+
+	metadata := &event.TaskExecutionMetadata{
+		GeneratedName: tCtx.TaskExecutionMetadata().GetTaskExecutionID().GetGeneratedName(),
```
EngHabu, line 156: Following the changes to "if we know the value of the field, we shouldn't ask plugin authors to fill them in...", GeneratedName is something in our control...

And so is ResourcePoolInfo...

katrogan: oh wow! i got thrown off by the allocation token in the webapi plugin, ty for pointing this out re: resource manager

go/tasks/pluginmachinery/internal/webapi/allocation_token.go
katrogan: no longe relevant

katrogan:
PTAL @EngHabu @kumare3

go/tasks/pluginmachinery/core/exec_context.go
```diff
@@ -18,8 +18,11 @@ type TaskReader interface {
 // An interface that is passed to every plugin invocation. It carries all meta and contextual information for the current
 // task execution
 type TaskExecutionContext interface {
-	// Returns a resource manager that can be used to create reservations for limited resources
-	ResourceManager() ResourceManager
+	// During execution time, plugins can call AllocateResource() to register a token to the token pool associated with a resource with the resource manager.
+	AllocateResource(ctx context.Context, namespace ResourceNamespace, allocationToken string, constraintsSpec ResourceConstraintsSpec) (AllocationStatus, error)
```
EngHabu, line 22: Can you elaborate on why do we need this change? it's sort of a break from the rest of the APIs in this interface...

EngHabu (commented):
I like this a whole lot better... so basically out of all the new fields we added to flyteIdl, only ExternalIds are explicitly added by plugins... right?

Just one comment (I have a feeling I need to see the flytepropeller change to understand why...)

go/tasks/pluginmachinery/core/exec_context.go
katrogan: yes these calls are now intercepted (in propeller code) to record allocated resource pool information

EngHabu: so Propeller also provides the implementation of the ResourceManager as well.. can't we intercept those calls there instead?

katrogan: but this interface is tied to a specific task execution context. more convenient since there's no global state that needs to be accounted for in tracing resource allocation to specific task executions

katrogan:
@EngHabu one more time please?

go/tasks/pluginmachinery/core/exec_context.go
```diff
@@ -57,6 +59,9 @@ type TaskExecutionContext interface {
 
 	// Returns a handle to the Task events recorder, which get stored in the Admin.
 	EventsRecorder() EventsRecorder
+
+	// Returns details about any resources allocated during the execution of this task.
+	GetResourcePoolInfo() []*event.ResourcePoolInfo
```
EngHabu, line 64: This is not used by plugins, right? is it needed here?

EngHabu (approved):

go/tasks/pluginmachinery/core/exec_context.go
katrogan: sure, it made things a little cleaner in propeller to be able to just the interface type but updated. ptal @EngHabu

EngHabu (approved):

PR #172, opened 2021-04-13
[Array-Jobs] Support separate namespace + Fix remote logs

akhurana001:
# TL;DR
 Adds supports for the following in K8s Array Jobs:
- Running in a separate namespace from parent job based on a regex config.
- Add Logs/job-status for array-jobs . Similar to current Batch behavior
- Add tests.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR


go/tasks/plugins/array/k8s/config.go
```diff
@@ -113,8 +115,14 @@ type Config struct {
 	RemoteClusterConfig  ClusterConfig     `json:"remoteClusterConfig" pflag:"-,Configuration of remote K8s cluster for array jobs"`
 	NodeSelector         map[string]string `json:"node-selector" pflag:"-,Defines a set of node selector labels to add to the pod."`
 	Tolerations          []v1.Toleration   `json:"tolerations"  pflag:"-,Tolerations to be applied for k8s-array pods"`
+	Namespace            string            `json:"namespace"  pflag:"-,Namespace pattern to spawn array-jobs in. Defaults to parent namespace if not set"`
```
anandswaminathan, line 118: I don't think this is right. This is config for the entire deployment right? Not for workflow

go/tasks/plugins/array/k8s/monitor.go
```diff
@@ -192,11 +194,23 @@ func CheckPodStatus(ctx context.Context, client core.KubeClient, name k8sTypes.N
 	}
 
 	if pod.Status.Phase != v1.PodPending && pod.Status.Phase != v1.PodUnknown {
-		taskLogs, err := logs.GetLogsForContainerInPod(ctx, pod, 0, " (User)")
+		p, err := logs.InitializeLogPlugins(&GetConfig().LogConfig.Config)
```
anandswaminathan, line 197: why initialize for all calle?

go/tasks/plugins/array/k8s/config.go
akhurana001: Yes, this is for deployment and namespace here is a regex like `"a-{{.namespace}}-b"` so platform admin can decide on this if they want to use a separate namespace to run array-jobs. If this is not set, it's ignored and current behavior is maintained.

go/tasks/plugins/array/k8s/monitor.go
akhurana001: Looks like we do the same for spark plugin after the recent logging re-write . I will update this to be less frequent :)

go/tasks/plugins/array/k8s/config.go
```diff
@@ -113,8 +115,14 @@ type Config struct {
...
 	OutputAssembler      workqueue.Config
 	ErrorAssembler       workqueue.Config
+	LogConfig            LogConfig `json:"logs" pflag:",Config for log links for spark applications."`
```
EngHabu, line 121: for spark applications?

go/tasks/plugins/array/k8s/config.go
EngHabu, line 118:
```suggestion
	NamespaceTemplate            string            `json:"namespaceTemplate"  pflag:"-,Namespace pattern to spawn array-jobs in. Defaults to parent namespace if not set"`
```

akhurana001: resolving, as discussed

go/tasks/plugins/array/k8s/config.go
akhurana001: fixed

go/tasks/plugins/array/k8s/config.go
akhurana001: done

akhurana001:
PTAL @EngHabu @anandswaminathan Fixed some of the open issues regarding logs etc. and PR comments.

go/tasks/plugins/array/k8s/transformer.go
```diff
@@ -29,6 +32,22 @@ func (a *arrayTaskContext) InputReader() io.InputReader {
 	return a.arrayInputReader
 }
 
+func GetNamespaceForExecution(tCtx core.TaskExecutionContext) string {
```
EngHabu, line 35: Can you make it take `namespaceTemplate` (or the entire config if needed)? I try to avoid calling GetConfig() outside the initialization code to make it easier to unit test these functions...

```suggestion
func GetNamespaceForExecution(tCtx core.TaskExecutionContext, namespaceTemplate string) string {
```

akhurana001: done

EngHabu (approved):

PR #175, opened 2021-05-12
[K8s Array] Fix ResourceQuota usage + logs for partially running jobs

akhurana001:
# TL;DR
Fixes the following:

- **Log-links for partially running jobs**: If some sub-tasks are running while others are waiting for resources/tokens, the status returned is `PhaseWaitingForResources` which currently doesn't support logs. Added a new method with support for logs. Will deprecate the older one once this is checked in and all references are updated in FlytePropeller.

- **Resource Quotas Issues**:   Re-submitting all sub-jobs on each sync incorrectly increase the quota usage and causes quota issues due to quota being an admission control invoked before etc.d rejects the request as resource already exists. (https://github.com/kubernetes/kubernetes/issues/76787 ). This is reconciled in the background (usually after a few minutes) but is still an issue for batch jobs till the initial sub-jobs succeed, they will keep on getting re-tried/increase quota usage.  Added a get check to fix this.  This is also an issue at other places where we rely on resubmit/de-dupe behavior but a larger issue with Batch jobs where there can be 1000s of pods getting re-submitted. 

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/go/tasks/pluginmachinery/core/phase.go
+++ b/go/tasks/pluginmachinery/core/phase.go
@@ -159,6 +159,12 @@ func PhaseInfoWaitingForResources(t time.Time, version uint32, reason string) Ph
 	return pi
 }
 
+func PhaseInfoWaitingForResourcesInfo(t time.Time, version uint32, reason string, info *TaskInfo) PhaseInfo {
+	pi := phaseInfo(PhaseWaitingForResources, version, nil, info)
+	pi.reason = reason
+	return pi
+}
+
 func PhaseInfoQueued(t time.Time, version uint32, reason string) PhaseInfo {
 	pi := phaseInfo(PhaseQueued, version, nil, &TaskInfo{OccurredAt: &t})
 	pi.reason = reason
--- a/go/tasks/pluginmachinery/core/phase.go
+++ b/go/tasks/pluginmachinery/core/phase.go
@@ -152,13 +152,14 @@ func PhaseInfoNotReady(t time.Time, version uint32, reason string) PhaseInfo {
 	return pi
 }
 
-// Return in the case the plugin is not ready to start
+// Deprecated: Please use PhaseInfoWaitingForResourcesInfo instead
 func PhaseInfoWaitingForResources(t time.Time, version uint32, reason string) PhaseInfo {
 	pi := phaseInfo(PhaseWaitingForResources, version, nil, &TaskInfo{OccurredAt: &t})
 	pi.reason = reason
 	return pi
 }
 
+// Return in the case the plugin is not ready to start
 func PhaseInfoWaitingForResourcesInfo(t time.Time, version uint32, reason string, info *TaskInfo) PhaseInfo {
 	pi := phaseInfo(PhaseWaitingForResources, version, nil, info)
 	pi.reason = reason
--- a/go/tasks/plugins/array/k8s/task.go
+++ b/go/tasks/plugins/array/k8s/task.go
@@ -113,7 +113,7 @@ func (t Task) Launch(ctx context.Context, tCtx core.TaskExecutionContext, kubeCl
 
 	// Check for existing pods to prevent unnecessary Resource-Quota usage: https://github.com/kubernetes/kubernetes/issues/76787
 	existingPod := &corev1.Pod{}
-	err = kubeClient.GetClient().Get(context.Background(), client.ObjectKey{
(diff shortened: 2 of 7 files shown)


go/tasks/pluginmachinery/core/phase.go
```diff
@@ -159,6 +159,12 @@ func PhaseInfoWaitingForResources(t time.Time, version uint32, reason string) Ph
 	return pi
 }
 
+func PhaseInfoWaitingForResourcesInfo(t time.Time, version uint32, reason string, info *TaskInfo) PhaseInfo {
```
EngHabu, line 162: Can you add `// Deprecated: Please use PhaseInfoWaitingForResourcesInfo instead.` to the `PhaseInfoWaitingForResources` func?

go/tasks/plugins/array/k8s/task.go
```diff
@@ -111,32 +111,47 @@ func (t Task) Launch(ctx context.Context, tCtx core.TaskExecutionContext, kubeCl
...
-		if k8serrors.IsForbidden(err) {
-			if strings.Contains(err.Error(), "exceeded quota") {
-				// TODO: Quota errors are retried forever, it would be good to have support for backoff strategy.
-				logger.Infof(ctx, "Failed to launch  job, resource quota exceeded. Err: %v", err)
-				t.State = t.State.SetPhase(arrayCore.PhaseWaitingForResources, 0).SetReason("Not enough resources to launch job")
-			} else {
-				t.State = t.State.SetPhase(arrayCore.PhaseRetryableFailure, 0).SetReason("Failed to launch job.")
+	// Check for existing pods to prevent unnecessary Resource-Quota usage: https://github.com/kubernetes/kubernetes/issues/76787
+	existingPod := &corev1.Pod{}
+	err = kubeClient.GetClient().Get(context.Background(), client.ObjectKey{
```
EngHabu, line 116: This is scary 😮 . It means we will issue two requests for every create. A couple of things:
1. Can you add a metric to count how many times we do a `Get` and find the pod?
2. Can you make it a config setting to control whether to do a Get before create or not? 
3. I think adding progress limit per step will greatly improve the reliability and success rate here... Is that something you are also working on? even if it's a config (e.g. launch 10 at a time) will make sure we will at most retry those 10 not 1000...

go/tasks/pluginmachinery/core/phase.go
akhurana001: Done

go/tasks/plugins/array/k8s/task.go
akhurana001: I moved to use `cache` instead of k8s-client . Believe that will add a bit of delay but should resolve most of the concerns ?  

Also need to start working on 3 . Trying to get this in a good state otherwise and have a couple of use-cases running in prod first.

go/tasks/plugins/array/k8s/task.go
```diff
@@ -111,32 +111,47 @@ func (t Task) Launch(ctx context.Context, tCtx core.TaskExecutionContext, kubeCl
...
-		if k8serrors.IsForbidden(err) {
-			if strings.Contains(err.Error(), "exceeded quota") {
-				// TODO: Quota errors are retried forever, it would be good to have support for backoff strategy.
-				logger.Infof(ctx, "Failed to launch  job, resource quota exceeded. Err: %v", err)
-				t.State = t.State.SetPhase(arrayCore.PhaseWaitingForResources, 0).SetReason("Not enough resources to launch job")
-			} else {
-				t.State = t.State.SetPhase(arrayCore.PhaseRetryableFailure, 0).SetReason("Failed to launch job.")
+	// Check for existing pods to prevent unnecessary Resource-Quota usage: https://github.com/kubernetes/kubernetes/issues/76787
+	existingPod := &corev1.Pod{}
+	err = kubeClient.GetCache().Get(context.Background(), client.ObjectKey{
```
EngHabu, line 116:
```suggestion
	err = kubeClient.GetCache().Get(ctx, client.ObjectKey{
```

EngHabu (approved):

PR #186, opened 2021-07-07
[K8s-Array] Add finalizer

akhurana001:
# TL;DR

- Add finalizer
- Update log-name to be in sync with  actual pod names i.e. pod-${retry}-${index}

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/go/tasks/plugins/array/k8s/monitor.go
+++ b/go/tasks/plugins/array/k8s/monitor.go
@@ -224,7 +224,7 @@ func FetchPodStatusAndLogs(ctx context.Context, client core.KubeClient, name k8s
 			o, err := logPlugin.GetTaskLogs(tasklog.Input{
 				PodName:          pod.Name,
 				Namespace:        pod.Namespace,
-				LogName:          fmt.Sprintf(" #%d-%d", index, retryAttempt),
+				LogName:          fmt.Sprintf(" #%d-%d", retryAttempt, index),
 				PodUnixStartTime: pod.CreationTimestamp.Unix(),
 			})
 
--- a/go/tasks/plugins/array/k8s/task.go
+++ b/go/tasks/plugins/array/k8s/task.go
@@ -260,7 +260,7 @@ func (t Task) Finalize(ctx context.Context, tCtx core.TaskExecutionContext, kube
 
 	err := kubeClient.GetClient().Get(ctx, k8sTypes.NamespacedName{
 		Name:      podName,
-		Namespace: tCtx.TaskExecutionMetadata().GetNamespace(),
+		Namespace: GetNamespaceForExecution(tCtx, t.Config.NamespaceTemplate),
 	}, pod)
 
 	if err != nil {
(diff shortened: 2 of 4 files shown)


anandswaminathan:
+1

anandswaminathan:
PTAL @kumare3 @EngHabu

go/tasks/plugins/array/k8s/monitor.go
```diff
@@ -224,7 +224,7 @@ func FetchPodStatusAndLogs(ctx context.Context, client core.KubeClient, name k8s
 			o, err := logPlugin.GetTaskLogs(tasklog.Input{
 				PodName:          pod.Name,
 				Namespace:        pod.Namespace,
-				LogName:          fmt.Sprintf(" #%d-%d", index, retryAttempt),
+				LogName:          fmt.Sprintf(" #%d-%d", retryAttempt, index),
```
EngHabu, line 227: Don't they get ordered by the name? is this better UX than essentially grouping the task by the index?

go/tasks/plugins/array/k8s/task.go
```diff
@@ -227,8 +251,34 @@ func (t Task) Finalize(ctx context.Context, tCtx core.TaskExecutionContext, kube
...
+	pod := &v1.Pod{
+		TypeMeta: metaV1.TypeMeta{
+			Kind:       PodKind,
+			APIVersion: v1.SchemeGroupVersion.String(),
+		},
+	}
+
+	err := kubeClient.GetClient().Get(ctx, k8sTypes.NamespacedName{
+		Name:      podName,
+		Namespace: tCtx.TaskExecutionMetadata().GetNamespace(),
```
EngHabu, line 263: [line 233 above](https://github.com/flyteorg/flyteplugins/pull/186/files#diff-01ee392f44bbd91e1e5f4afd999327e02f8f1df54fd5b5b576478e903cbcb3c2R233) uses a different piece of code to get the namespace... I think we need that.... and given this, maybe we should move the logic that builds this Pod skeleton into a function and just use it in both places?

akhurana001: Good catch. This is not building a pod, just name + namespace to query for the pod. There is already a common method for namespace, I changed this to use that here as well.

go/tasks/plugins/array/k8s/monitor.go
akhurana001: This is cleaner as this is in-sync with the pod name which can be otherwise confusing if you are looking at k8s-logs or just investigating any issues.

akhurana001:
@EngHabu @kumare3 PTAL

go/tasks/plugins/array/k8s/monitor.go
kumare3: agreed

kumare3 (approved):

PR #187, opened 2021-07-08
Greater consistency in map pod task pod spec construction

katrogan:
Signed-off-by: Katrina Rogan <katroganGH@gmail.com>

# TL;DR
Apply conventional resource and env var overrides in map pod tasks. This is [already done](https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/plugins/k8s/sidecar/sidecar.go#L64,L65) for regular pod tasks.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
User reported bug

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/1223

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/plugins/array/k8s/transformer.go
+++ b/go/tasks/plugins/array/k8s/transformer.go
@@ -94,6 +94,40 @@ func buildPodMapTask(task *idlCore.TaskTemplate, metadata core.TaskExecutionMeta
 	return pod, nil
 }
 
+// Here we customize the k8sPod primary container by templatizing args.
+// The call to ToK8sPodSpec for the task container target
+// case already handles this but we must explicitly do so for K8sPod task targets.
+func modifyMapPodTaskPrimaryContainer(ctx context.Context, tCtx core.TaskExecutionContext, arrTCtx *arrayTaskContext, container *v1.Container) error {
+	var err error
+	container.Args, err = template.Render(ctx, container.Args,
+		template.Parameters{
+			TaskExecMetadata: tCtx.TaskExecutionMetadata(),
+			Inputs:           arrTCtx.arrayInputReader,
+			OutputPath:       tCtx.OutputWriter(),
+			Task:             tCtx.TaskReader(),
+		})
+	if err != nil {
+		return err
+	}
+	container.Command, err = template.Render(ctx, container.Command,
+		template.Parameters{
+			TaskExecMetadata: tCtx.TaskExecutionMetadata(),
+			Inputs:           arrTCtx.arrayInputReader,
+			OutputPath:       tCtx.OutputWriter(),
+			Task:             tCtx.TaskReader(),
+		})
+	if err != nil {
+		return err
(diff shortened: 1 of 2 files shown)


go/tasks/plugins/array/k8s/transformer.go
```diff
@@ -94,6 +94,40 @@ func buildPodMapTask(task *idlCore.TaskTemplate, metadata core.TaskExecutionMeta
 	return pod, nil
 }
 
+// Here we customize the k8sPod primary container by templatizing args.
+// The call to ToK8sPodSpec for the task container target
+// case already handles this but we must explicitly do so for K8sPod task targets.
+func modifyMapPodTaskPrimaryContainer(ctx context.Context, tCtx core.TaskExecutionContext, arrTCtx *arrayTaskContext, container *v1.Container) error {
+	var err error
```
kumare3, line 101: isnt this container modification a simple shared library (flytek8s)

katrogan: it is, but for flyte container definitions, not k8s ones. ideally all this duplicate code[ goes away](https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/plugins/array/k8s/transformer.go#L56) after we introduce array nodes

EngHabu: Wild guess: are you, @kumare3, asking about `consistency` as in `concurrent programming consistency` ? it took me a sec to realize how off base I was :-D

go/tasks/plugins/array/k8s/transformer.go
```diff
@@ -94,6 +94,40 @@ func buildPodMapTask(task *idlCore.TaskTemplate, metadata core.TaskExecutionMeta
...
+		template.Parameters{
+			TaskExecMetadata: tCtx.TaskExecutionMetadata(),
+			Inputs:           arrTCtx.arrayInputReader,
+			OutputPath:       tCtx.OutputWriter(),
+			Task:             tCtx.TaskReader(),
+		})
+	if err != nil {
+		return err
+	}
+	resources := flytek8s.ApplyResourceOverrides(ctx, container.Resources)
```
EngHabu, line 122: Do we apply that only on the primary container? or should it apply to all containers? I don't know what we do in regular pod tasks so the same question applies....

katrogan: good q: we only do it for primary containers in pod tasks now. not sure if it should be extended to other container definitions or if a hands-off model is okay

EngHabu (approved):

go/tasks/plugins/array/k8s/transformer.go
kumare3: what?

PR #189, opened 2021-07-09
[WIP]K8s-Array Finalizer

akhurana001:
Cherry-pick https://github.com/flyteorg/flyteplugins/pull/186 and revert https://github.com/flyteorg/flyteplugins/commit/ba8eb6cba064bb4333353733ee7cd1a830bc1c80 which is causing a idl conflict


go/tasks/plugins/array/k8s/task.go
```diff
@@ -46,6 +47,29 @@ const (
 	LaunchReturnState
 )
 
+const finalizer = "array.k8s.finalizers.flyte.k8s.io"
```
kumare3, line 50: can we instead call it 
```suggestion
const finalizer = "finalizers.array.flyte.org"
```

kumare3:
you have to also sign dco

go/tasks/plugins/array/k8s/task.go
```diff
@@ -220,8 +244,34 @@ func (t Task) Finalize(ctx context.Context, tCtx core.TaskExecutionContext, kube
...
+		Namespace: GetNamespaceForExecution(tCtx, t.Config.NamespaceTemplate),
+	}, pod)
+
+	if err != nil {
+		if !k8serrors.IsNotFound(err) {
+			logger.Errorf(ctx, "Error fetching pod [%s] in Finalize [%s]", podName, err)
+			return err
+		}
+	} else {
+		pod = clearFinalizer(pod)
```
kumare3, line 265: do you want to clear the finalizer for the cases when `err != nil`

go/tasks/plugins/array/k8s/task.go
akhurana001: RuntimeExecutionError: failed during plugin execution, caused by: failed to execute handle for plugin [k8s-array]: [SUBMIT_JOB_FAILED] Failed to submit job., caused by: Pod  is invalid: metadata.finalizers[0]: Invalid value: "finalizers.array.flyte.org": name is neither a standard finalizer name nor is it fully qualified

akhurana001: I changed it to "flyte/array"

go/tasks/plugins/array/k8s/task.go
akhurana001: I believe returning an err might be cleaner if we get an error from k8s-api ? except for notfound error which is ignored here.

akhurana001:
Also please review https://github.com/flyteorg/flyteplugins/pull/186/files . This PR was just to test with the branch we use

PR #193, opened 2021-08-03
Support setting ephemeral storage

katrogan:
# TL;DR
Now that k8s supports setting ephemeral storage we should expose that as a configurable setting to end users. At the very least this change allows doing so in pod tasks.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/494

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/pluginmachinery/flytek8s/container_helper.go
+++ b/go/tasks/pluginmachinery/flytek8s/container_helper.go
@@ -73,7 +73,6 @@ func ApplyResourceOverrides(ctx context.Context, resources v1.ResourceRequiremen
 	// TODO: Make configurable. 1/15/2019 Flyte Cluster doesn't support setting storage requests/limits.
 	// https://github.com/kubernetes/enhancements/issues/362
 	delete(resources.Requests, v1.ResourceStorage)
-
 	delete(resources.Limits, v1.ResourceStorage)
 
 	// Override GPU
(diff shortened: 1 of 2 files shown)


katrogan:
cc @jeevb

jeevb:
Awesome! 🙏

I imagine this will require changes in flytekit (and flyteidl)?

EngHabu (commented):
cool! a couple of questions:
* A quick google search shows me this [result](https://kubernetes.io/docs/concepts/configuration/manage-resources-containers/#local-ephemeral-storage) which suggests it's still in beta? but then it says 1.10... so I'm confused... mind double checking...
* If it only went to stable recently, should we have a config here that enables this behavior? otherwise it goes back to the old behavior... and maybe the default value is the new behavior?

katrogan:
@jeevb yup, working on those changes in parallel

@EngHabu hah yeah we went through the same confusion, see https://flyte-org.slack.com/archives/CREL4QVAQ/p1624469230251200?thread_ts=1624468996.250400&cid=CREL4QVAQ

it should now be supported on 1.19+

EngHabu (approved):

PR #194, opened 2021-08-10
Fix resource requirements for sidecar/pod tasks

jeevb:
# TL;DR
Fixes task node resource overrides via `.with_overrides` in flytekit not applying to resource requirements in sidecar/pod tasks. Also fixes k8s plugin default resource requests not applying to sidecar/pod tasks.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added
 - [ ] Code documentation added
 - [ ] Any pending items have an associated Issue

## Complete description
Merges task node resource requirements with resource requirements of primary container in sidecar/pod task, after applying defaults where necessary.

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/1323

PR #197, opened 2021-08-12
Apply overrides before applying defaults to sidecar resource requirements

jeevb:
# TL;DR
Apply overrides before applying defaults to sidecar resource requirements. Related to: https://github.com/flyteorg/flyteplugins/pull/194.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Refactored code to apply resource requirements overrides from node spec first before applying default resource requirements. Found a bug in `ApplyResourceOverrides` as a result, and fixed the unit tests to properly catch the issue.

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/go/tasks/plugins/k8s/sidecar/sidecar.go
+++ b/go/tasks/plugins/k8s/sidecar/sidecar.go
@@ -39,11 +39,11 @@ func validateAndFinalizePod(
 	for index, container := range pod.Spec.Containers {
 		if container.Name == primaryContainerName {
 			hasPrimaryContainer = true
-			container.Resources = *flytek8s.ApplyResourceOverrides(ctx, container.Resources)
 			if taskCtx.TaskExecutionMetadata().GetOverrides() != nil && taskCtx.TaskExecutionMetadata().GetOverrides().GetResources() != nil {
 				resOverrides := taskCtx.TaskExecutionMetadata().GetOverrides().GetResources()
 				flytek8s.MergeResources(*resOverrides, &container.Resources)
 			}
+			container.Resources = *flytek8s.ApplyResourceOverrides(ctx, container.Resources)
 		}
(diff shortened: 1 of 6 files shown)


go/tasks/plugins/k8s/sidecar/sidecar.go
```diff
@@ -39,11 +39,11 @@ func validateAndFinalizePod(
 	for index, container := range pod.Spec.Containers {
 		if container.Name == primaryContainerName {
 			hasPrimaryContainer = true
-			container.Resources = *flytek8s.ApplyResourceOverrides(ctx, container.Resources)
 			if taskCtx.TaskExecutionMetadata().GetOverrides() != nil && taskCtx.TaskExecutionMetadata().GetOverrides().GetResources() != nil {
 				resOverrides := taskCtx.TaskExecutionMetadata().GetOverrides().GetResources()
 				flytek8s.MergeResources(*resOverrides, &container.Resources)
 			}
+			container.Resources = *flytek8s.ApplyResourceOverrides(ctx, container.Resources)
```
katrogan, line 46: primary container resources should never be nil, but just in case should we check before dereferencing?

katrogan (approved):

go/tasks/plugins/k8s/sidecar/sidecar.go
katrogan: just kidding, ApplyResourceOverrides handles initializing resources always

jeevb: Resolved offline.

PR #202, opened 2021-08-23
Add Snowflake backend plugin

pingsutw:

# TL;DR
Implements WebAPI plugin for Snowflake.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Use Snowflake Rest API instead of Golang API to implement the backend plugin since Snowflake Golang API doesn't have a function that cancels a query or gets query status by query ID.

And Snowflake Rest API only support client use `Oauth` or `JSON Web Token (JWT)` to authenticate to the server.
https://docs.snowflake.com/en/developer-guide/sql-api/guide.html#authenticating-to-the-server

If clients want to use `username` and `password` to authenticate to the server, they can use `SQLAlchemyTask`.
For example: 
```python
task_config=SQLAlchemyConfig(
        uri='snowflake://{user}:{password}@{account}/'.format(
            user='haytham',
            password='<your_password>',
            account='hoa61102',
        ),
```

Refer:
[Snowflake SQL API Reference](https://docs.snowflake.com/en/developer-guide/sql-api/reference.html#snowflake-sql-api-reference)
[Snowflake SQL API Developer Guide](https://docs.snowflake.com/en/developer-guide/sql-api/guide.html#authenticating-to-the-server)
[Snowflake Golang API](https://github.com/snowflakedb/gosnowflake)

## TODO
- [ ] Add example in Flytesnack
- [ ] Update Flyteidl
- [ ] Add Snowflake tasks in Flytekit

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/plugins/webapi/snowflake/plugin.go
+++ b/go/tasks/plugins/webapi/snowflake/plugin.go
@@ -251,7 +251,7 @@ func createTaskInfo(queryID string, account string) *core.TaskInfo {
 		OccurredAt: &timeNow,
 		Logs: []*flyteIdlCore.TaskLog{
 			{
-				Uri: fmt.Sprintf("https://%v.snowflakecomputing.com/console#/monitoring/queries/detail?queryID=%v",
+				Uri: fmt.Sprintf("https://%v.snowflakecomputing.com/console#/monitoring/queries/detail?queryId=%v",
 					account,
 					queryID),
 				Name: "Snowflake Console",
--- a/go/tasks/plugins/webapi/snowflake/plugin.go
+++ b/go/tasks/plugins/webapi/snowflake/plugin.go
@@ -1,0 +1,286 @@
+package snowflake
+
+import (
+	"bytes"
+	"context"
+	"encoding/gob"
+	"encoding/json"
+	"fmt"
+	flyteIdlCore "github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
+	pluginErrors "github.com/flyteorg/flyteplugins/go/tasks/errors"
+	pluginsCore "github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/core"
+	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/core/template"
+	pluginUtils "github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/utils"
+	"github.com/flyteorg/flytestdlib/errors"
+	"github.com/flyteorg/flytestdlib/logger"
+	"io/ioutil"
+	"net/http"
+	"os"
+	"strconv"
+	"time"
+
+	"github.com/flyteorg/flytestdlib/promutils"
+
+	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery"
+	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/core"
+	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/webapi"
+)
+
+const (
+	ErrUser   errors.ErrorCode = "User"
+	ErrSystem errors.ErrorCode = "System"
+)
+
+type Plugin struct {
+	metricScope    promutils.Scope
+	cfg            *Config
+	snowflakeToken string
+}
+
+type ResourceWrapper struct {
+	Status  string
+	Message string
+}
+
+type ResourceMetaWrapper struct {
+	QueryID string
+	Account string
+}
+
+func (p Plugin) GetConfig() webapi.PluginConfig {
+	return GetConfig().WebAPI
+}
+
+type QueryInfo struct {
+	Account   string
+	Warehouse string
+	Schema    string
+	Database  string
+	Statement string
+}
+
+// TODO: Add QueryJobConfig in Flyteidl
+type QueryJobConfig struct {
+	Account   string `json:"account"`
+	Warehouse string `json:"warehouse"`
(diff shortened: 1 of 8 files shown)


go/tasks/plugins/webapi/snowflake/plugin.go
```diff
@@ -0,0 +1,286 @@
...
+					queryId),
+				Name: "Snowflake Console",
+			},
+		},
+	}
+}
+
+func getSnowflakeToken() string {
+	return os.Getenv("SNOWFLAKE_TOKEN")
+}
```
pingsutw, line 265: Don't know how to properly get the token. 
1. Get environment variable from k8s?
2. Get from the secret in k8s?
3. Users specify the token in Flytekit?
4. or Others

Any thought?

pingsutw:
Going to add more tests to improve the test coverage

go/tasks/plugins/webapi/snowflake/plugin.go
wild-endeavor: Users should probably not be the ones to specify the token, though perhaps they should be able to specify an override.  we can do that in the future if it becomes necessary.

The plugin should probably fetch the k8s secrets from the setup context https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/pluginmachinery/core/setup_context.go#L22

@kumare3 @EngHabu could you point kevin to the docs to all the secrets stuff?

kumare3: https://docs.flyte.org/projects/cookbook/en/latest/auto/core/containerization/use_secrets.html?highlight=secret

kumare3: So the backend secrets are accessed using the
[SecretManager](https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/pluginmachinery/core/secret_manager.go), which is given in the context [here](https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/pluginmachinery/core/exec_context.go#L32)

For backend this is simply accessing File based secrets using [this](https://github.com/flyteorg/flytepropeller/blob/master/pkg/controller/nodes/task/handler.go#L777)

Here is the code
https://github.com/flyteorg/flytepropeller/blob/c59db1be84897f8465dd14413c65088bce600646/pkg/controller/nodes/task/secretmanager/secrets.go#L25

So you can inject the secret into flytepropeller env, or use k8s secrets to mount a secret into the pod. Preferred to mount it into the pod

kumare3: The secret itself should be added during deployment and it should be done as part of the helm / kustomize manifest

kumare3: cc @EngHabu, we did not merge the paths for the webhook and propeller, i thought we were going to?

pingsutw: Thx! I updated it.

go/tasks/plugins/webapi/snowflake/plugin.go
```diff
@@ -0,0 +1,294 @@
...
+
+func (p Plugin) Delete(ctx context.Context, taskCtx webapi.DeleteContext) error {
+	exec := taskCtx.ResourceMeta().(*ResourceMetaWrapper)
+	req, err := buildRequest(post, QueryInfo{}, p.cfg.snowflakeEndpoint,
+		exec.Account, exec.Token, exec.QueryID, true)
+	if err != nil {
+		return err
+	}
+	resp, err := p.client.Do(req)
```
wild-endeavor, line 180: If this returns a non-200, is it an err @EngHabu ?  Or is it a response with no error but a non-200?  do we need to check for it?

go/tasks/plugins/webapi/snowflake/plugin.go
```diff
@@ -0,0 +1,294 @@
...
+	return "default", p.cfg.ResourceConstraints, nil
+}
+
+func (p Plugin) Create(ctx context.Context, taskCtx webapi.TaskExecutionContextReader) (webapi.ResourceMeta,
+	webapi.Resource, error) {
+	task, err := taskCtx.TaskReader().Read(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+	fmt.Print("kevin = ", p.cfg.TokenKey)
```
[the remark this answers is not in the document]
pingsutw: Yes, sorry forgot to remove it.

pingsutw:
Snowflake Demo: 

https://user-images.githubusercontent.com/37936015/131316960-ccdb8e7e-f6dc-4a10-8950-61eb11d29ed6.mp4


Snowflake workflow example: https://gist.github.com/pingsutw/dcda212f5e342e20d974cfe027206e57

pingsutw:
Already Updated Flyteidl version.

pingsutw:
Athena tests and others failed due to IDL change in https://github.com/flyteorg/flyteidl/commit/8d83de8d1b401e91a4b46fb52f9f5f41f597b4bc.
I will file a new ticket, and work on it.

wild-endeavor:
there's already a PR https://github.com/flyteorg/flyteplugins/pull/198 cc @mayitbeegh it just needs to be merged.

pingsutw:
I see, thanks.

wild-endeavor (approved):

PR #203, opened 2021-08-23
add composable PluginCleanupPolicy (flyteorg/flyte#1345)

clairemcginty:
## _Read then delete this section_

_- Make sure to use a concise title for the pull-request._

_- Use #patch, #minor or #major in the pull-request title to bump the corresponding version. Otherwise, the patch version
will be bumped. [More details](https://github.com/marketplace/actions/github-tag-bump)_

# TL;DR
Adds an optional interface that k8s plugins can implement, `PluginCleanupPolicy`, as an alternative to flytepropeller simply deleting the resource.

cc @regadas , @EngHabu 

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
https://github.com/flyteorg/flyte/issues/1345 Adds an optional interface that k8s plugins can implement, `PluginCleanupPolicy`, as an alternative to flytepropeller simply deleting the resource. The idea is that it would be used in flytepropeller [as in this branch](https://github.com/flyteorg/flytepropeller/compare/master...clairemcginty:use_cleanup_policy?expand=1), where it checks if the plugin implements `PluginCleanupPolicy` and if so substitutes that action. For our use case we just need the `Abort` action override; I know flytepropeller does some cleanup actions in its finalizers also, but we don't need to override that.

Note: I'm not sure about what unit tests to add here. My plan was to add specific unit tests to flytepropeller to check that the cleanup policy is being selected when applicable. I've also tested end to end (the proposed changes to flyteplugins, flytepropeller, and my custom plugin) with a local setup on Minikube.

## Tracking Issue

## Follow-up issue
_NA_
OR


go/tasks/pluginmachinery/k8s/mocks/plugin_cleanup_policy.go
```diff
@@ -0,0 +1,30 @@
+// Code generated by mockery v1.0.1. DO NOT EDIT.
```
clairemcginty, line 1: I matched the version of Mockery used by the other interfaces in this repo, but it's definitely pretty out of date

kumare3: ya, so the way to generate the code is, 
`make download_tooling`, `make generate`

clairemcginty: thank you! 🙏 I've applied the changes.

regadas (approved):
LGTM!

go/tasks/pluginmachinery/k8s/plugin.go
```diff
@@ -91,3 +91,8 @@ type Plugin interface {
 	// Properties desired by the plugin
 	GetProperties() PluginProperties
```
regadas, line 92: I was thinking and wdyt of having`PluginCleanupPolicy` defined in `PluginProperties`?

clairemcginty: that sounds reasonable to me! @EngHabu wdyt?

EngHabu: I would rather keep properties to a safe-to-serialize struct (so that it can be read from config...) we can add another optional (`GetOverrides() PluginOverrides`) and that would be where we add all of these functions people can choose to implement or leave empty...

go/tasks/pluginmachinery/k8s/plugin.go
```diff
@@ -91,3 +91,8 @@ type Plugin interface {
...
 }
+
+// Defines functions to clean up the resource object after the task has been completed or cancelled.
+type PluginCleanupPolicy interface {
```
EngHabu, line 96:
```suggestion
type PluginAbortOverride interface {
```

go/tasks/pluginmachinery/k8s/plugin.go
```diff
@@ -91,3 +91,8 @@ type Plugin interface {
...
+	OnAbort(ctx context.Context, kubeClient client.Client, resource client.Object) error
```
EngHabu, line 97: I would rather keep this signature consistent with Finalize/Abort... we will likely want to send additional things besides kubeClient (potentially the task name/project/domain...) so it's a good pattern to send a `Context` object instead...

clairemcginty: so just have the signature be `OnAbort(ctx context.Context, tCtx pluginsCore.TaskExecutionContext)` ?

EngHabu: I was thinking something like this... (with docs and perhaps better names :-D)
```
type PluginAbortOverride interface {
    PreAbort(ctx context.Context, tCtx pluginsCore.TaskExecutionContext, resource client.Object) (behavior AbortBehavior, err error)
}

type AbortBehavior struct {
    Resource client.Object
    DeleteResource    bool
    Update                  UpdateObjectOperation
    Patch                     PatchObjectOperation
}

type PatchObjectOperation struct {
    Patch     client.Patch
    Options []client.PatchOption
}

type UpdateObjectOperation struct {
    Options []client.UpdateOption
}
```

go/tasks/pluginmachinery/k8s/plugin.go
```diff
@@ -51,6 +51,31 @@ type PluginProperties struct {
...
+}
+
+type PluginAbortOverride interface {
+	OnAbort(ctx context.Context, tCtx pluginsCore.TaskExecutionContext, resource client.Object) (behavior AbortBehavior, err error)
+}
+
+type AbortBehavior struct {
+	// Optional override to the default k8s Resource being acted on.
+	Resource       client.Object
+	DeleteResource bool
```
clairemcginty, line 65: an additional thought is to add either custom constructors here, or a check in flytepropeller, to make sure that, for example, users don't set both `Delete` and `Update`, or other invalid configurations.

go/tasks/pluginmachinery/k8s/plugin.go
```diff
@@ -51,6 +51,31 @@ type PluginProperties struct {
 	// override that behavior unless the resource that gets created for this plugin does not consume resources (cluster's
 	// cpu/memory... etc. or external resources) once the plugin's Plugin.GetTaskPhase() returns a terminal phase.
 	DisableDeleteResourceOnFinalize bool
+	// Optionally specifies an override for the default OnAbort behavior (deleting the underlying resource).
+	AbortOverride PluginAbortOverride
```
EngHabu, line 55: Let's delete this... As discussed on the thread, I want to keep this struct serializable... from yaml -> go...
```suggestion
```

go/tasks/pluginmachinery/k8s/plugin.go
```diff
@@ -51,6 +51,31 @@ type PluginProperties struct {
...
+}
+
+type PluginAbortOverride interface {
+	OnAbort(ctx context.Context, tCtx pluginsCore.TaskExecutionContext, resource client.Object) (behavior AbortBehavior, err error)
```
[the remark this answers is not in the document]
EngHabu: error here will mean some catastrophic thing has occurred (e.g. the object belong to an unexpected type when you try to cast it)... I think, particularly when designing public APIs, it's a good thing to have from the get go...

go/tasks/pluginmachinery/k8s/plugin.go
EngHabu: Excellent point! I like the custom constructors pattern yes... 
e.g. https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/pluginmachinery/core/phase.go#L149-L220

clairemcginty: ok! I added some custom constructors in that style. It's a lot of LOC, not sure if I should move them into a separate file at this point?

go/tasks/pluginmachinery/k8s/plugin.go
```diff
@@ -91,3 +91,68 @@ type Plugin interface {
 	// Properties desired by the plugin
 	GetProperties() PluginProperties
 }
+
+type PluginAbortOverride interface {
```
kumare3, line 95: lets add some comments on this explaining when should you implement this interface and why it exists

go/tasks/pluginmachinery/k8s/plugin.go
```diff
@@ -91,3 +91,68 @@ type Plugin interface {
...
+
+func AbortBehaviorDeleteDefaultResource() AbortBehavior {
+	return AbortBehaviorDelete(nil)
+}
+
+func AbortBehaviorDelete(resource client.Object) AbortBehavior {
+	return AbortBehavior {
+		Resource: resource,
+		DeleteResource: true,
+		Update: nil,
```
kumare3, line 154: avoid setting zero values
pointers/refs have zero value of nil and bools have zero value of false. I appreciate that the zero behaviour is the default in all cases :), lets drop them and make them clean.

go/tasks/pluginmachinery/k8s/plugin.go
kumare3: overall this patch looks good to me, just needs a  comments and explanations to help new plugin authors understand the idea

go/tasks/pluginmachinery/k8s/plugin.go
```diff
@@ -91,3 +91,80 @@ type Plugin interface {
...
+		DeleteOnErr: deleteOnErr,
+	}
+}
+
+// AbortBehavior that updates the default resource
+func AbortBehaviorUpdateDefaultResource(updateOperation UpdateResourceOperation, deleteOnErr bool) AbortBehavior {
+	return AbortBehavior {
+		Update: &updateOperation,
+		DeleteOnErr: deleteOnErr,
+	}
```
EngHabu, line 145:
```suggestion
	return AbortBehaviorUpdate(updateOperation, deleteOnErr, nil)
```

go/tasks/pluginmachinery/k8s/plugin.go
```diff
@@ -91,3 +91,80 @@ type Plugin interface {
...
+type UpdateResourceOperation struct {
+	Options []client.UpdateOption
+}
+
+// AbortBehavior that patches the default resource
+func AbortBehaviorPatchDefaultResource(patchOperation PatchResourceOperation, deleteOnErr bool) AbortBehavior {
+	return AbortBehavior {
+		Patch: &patchOperation,
+		DeleteOnErr: deleteOnErr,
+	}
```
EngHabu, line 128:
```suggestion
	return AbortBehaviorPatch(patchOperation, deleteOnErr, nil)
```

go/tasks/pluginmachinery/k8s/plugin.go
```diff
@@ -91,3 +91,80 @@ type Plugin interface {
...
+		Update: &updateOperation,
+		DeleteOnErr: deleteOnErr,
+	}
+}
+
+// AbortBehavior that deletes the default resource
+func AbortBehaviorDeleteDefaultResource() AbortBehavior {
+	return AbortBehavior {
+		DeleteResource: true,
+	}
```
EngHabu, line 161:
```suggestion
	return AbortBehaviorDelete(nil)
```

EngHabu (commented):
Looks great! thank you for adapting the design. just a couple of minor comments and we can merge it today!

EngHabu:
Just to speed up the next review, there seems to be some lint errors and DCO isn't happy... mind fixing those too?

EngHabu (approved):

PR #208, opened 2021-09-10
Separate out resource overrides from static resource definitions

katrogan:
# TL;DR
The introduction of https://github.com/flyteorg/flytepropeller/pull/310 means that platform resource defaults are now applied as task resource defaults in the absence of task decorator resource specifications. For container tasks, this is fine. For pod tasks this results in the pod spec primary container definition resource spec being overwritten by the platform defaults. 

Tested with Flytesandbox.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
User-reported bug.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1450

## Follow-up issue
_NA_


go/tasks/pluginmachinery/flytek8s/container_helper.go
```diff
@@ -22,21 +22,32 @@ const resourceGPU = "gpu"
...
+		Requests: make(v1.ResourceList),
+	}
+	for key, val := range resources.Limits {
+		result.Limits[key] = val
+	}
+	for key, val := range resources.Requests {
+		result.Requests[key] = val
+	}
+	if resources.Limits == nil {
+		result.Limits = preferred.Limits
```
kumare3, line 37: preferred here seems to be inverted?
it seems to get applied only when resources are nil. 
so should they be called `default` instead

PR #209, opened 2021-09-13
InvalidImageName errors are not reported correctly

pingsutw:

# TL;DR
The task will keep running if we use an invalid image name 

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
The task will immediately fail if we use an invalid image name.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/306

## Follow-up issue
_NA_

```
 "containerStatuses": [
            {
                "image": "TEST/pingsutw/myapp:latest",
                "imageID": "",
                "lastState": {},
                "name": "ie9qnpo16b-n0-0",
                "ready": false,
                "restartCount": 0,
                "started": false,
                "state": {
                    "waiting": {
                        "message": "Failed to apply default image tag \"TEST/pingsutw/myapp:latest\": couldn't parse image reference \"TEST/pingsutw/myapp:latest\": invalid reference format: repository name must be lowercase",
                        "reason": "InvalidImageName"
                    }
                }
            }

```
## Before
image

## After
image

### DIFF
--- a/go/tasks/pluginmachinery/flytek8s/pod_helper.go
+++ b/go/tasks/pluginmachinery/flytek8s/pod_helper.go
@@ -253,7 +253,7 @@ func DemystifyPending(status v1.PodStatus) (pluginsCore.PhaseInfo, error) {
 									&pluginsCore.TaskInfo{OccurredAt: &t},
 								), nil
 
-							case "CreateContainerConfigError":
+							case "CreateContainerConfigError", "InvalidImageName":
 								t := c.LastTransitionTime.Time
 								return pluginsCore.PhaseInfoFailure(finalReason, finalMessage, &pluginsCore.TaskInfo{
 									OccurredAt: &t,
@@ -264,6 +264,7 @@ func DemystifyPending(status v1.PodStatus) (pluginsCore.PhaseInfo, error) {
 								return pluginsCore.PhaseInfoRetryableFailure(finalReason, finalMessage, &pluginsCore.TaskInfo{
 									OccurredAt: &t,
 								}), nil
+
 							default:
 								// Since we are not checking for all error states, we may end up perpetually
 								// in the queued state returned at the bottom of this function, until the Pod is reaped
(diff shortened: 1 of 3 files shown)


go/tasks/pluginmachinery/flytek8s/pod_helper.go
```diff
@@ -253,7 +253,7 @@ func DemystifyPending(status v1.PodStatus) (pluginsCore.PhaseInfo, error) {
 									&pluginsCore.TaskInfo{OccurredAt: &t},
 								), nil
 
-							case "CreateContainerConfigError":
+							case "CreateContainerConfigError", "InvalidImageName":
```
pmahindrakar-oss, line 256: Can we add a unit test to specifically test for this condition of the InvalidImageName and as per the desciption the task would keep running , So can we have the test simulate and after the fix the test succeeds with desired outcome

pingsutw: sure

pingsutw: Hi @pmahindrakar-oss 
How do I simulate this situation? We don't have integration tests in Flyteplugin.
I already have a unit test for it below to make sure that the task phase will be `PhasePermanentFailure`. is it enough?

pmahindrakar-oss: The unit test can just send a `status v1.PodStatus` object populated with `status.ContainerStatuses` which have 
`"CreateContainerConfigError", "InvalidImageName":` one of these values and the expectation would be to return 
`pluginsCore.PhaseInfoFailure(finalReason, finalMessage, &pluginsCore.TaskInfo{
									OccurredAt: &t,...`

I think that should be good enough and in the previous case before your fix it should be returning
`return pluginsCore.PhaseInfoSystemRetryableFailure(finalReason, finalMessage, &pluginsCore.TaskInfo{
									OccurredAt: &t,
								}), nil`

pmahindrakar-oss (approved):

katrogan (approved):

PR #211, opened 2021-09-18
Add util functions to convert Kubernetes DNS-1123 incompatible names to compatible names

mayitbeegh:
# TL;DR
Add util functions to convert incompatible names for Kubernetes pod to compatible names. This enables flytepropeller to fix incompatible pod names. See https://github.com/flyteorg/flytepropeller/pull/325 for full context.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1172

## Follow-up issue
_NA_
OR


go/tasks/pluginmachinery/utils/dns.go
```diff
@@ -0,0 +1,25 @@
+package utils
+
+import (
+	"regexp"
+	"strings"
+)
+
+var dns1123MaxLen = 63
```
pmahindrakar-oss, line 8: There is already available as DNS1123LabelMaxLength
which is  part api machinery validation function and can be utilized instead of this

go/tasks/pluginmachinery/utils/dns.go
```diff
@@ -0,0 +1,25 @@
...
+var dns1123InvalidRegex = regexp.MustCompile("[^-a-z0-9]")
+var camelCaseRegex = regexp.MustCompile("([a-z0-9])([A-Z])")
+
+func ConvertToDNS1123CompatibleString(name string) string {
+	name = ConvertCamelCaseToKebabCase(name) // best effort to preserve readability for Java class name
+	name = strings.ToLower(name)
+	name = dns1123InvalidRegex.ReplaceAllString(name, "")
+	name = strings.Trim(name, ".-")
+	if len(name) > dns1123MaxLen {
+		name = name[:dns1123MaxLen]
```
EngHabu, line 18: For uniqueness purposes, do you want to cut it even shorter and hash the rest?
Check here for examples: https://github.com/flyteorg/flytepropeller/blob/master/pkg/utils/encoder.go

go/tasks/pluginmachinery/utils/dns.go
mayitbeegh: Fixed

go/tasks/pluginmachinery/utils/dns.go
mayitbeegh: Good idea. `encoder.go` is in flytepropeller, so I suppose I can move everything to flytepropeller. I did it here in flyteplugins because validation also happens in plugin. So there's a chance that the pod name conversion helper is also needed in flyteplugins.

mayitbeegh: Code is moved to here https://github.com/flyteorg/flytepropeller/pull/325

EngHabu: Apologies I didn't mean to suggest moving the code there... just pointing you at examples that sort of overlap with what you are trying to do...

I'm ok actually keeping that code here.. (I would go as far as saying I prefer the code be here...)

mayitbeegh: They're moved back (among other changes)

go/tasks/pluginmachinery/utils/dns.go
```diff
@@ -0,0 +1,36 @@
...
+	"k8s.io/apimachinery/pkg/util/validation"
+)
+
+var dns1123InvalidRegex = regexp.MustCompile("[^-a-z0-9]")
+var camelCaseRegex = regexp.MustCompile("([a-z0-9])([A-Z])")
+
+const maxUniqueIDLength = 20
+
+// ConvertToDNS1123CompatibleString converts a string that doesn't conform to the definition of a label in DNS (RFC 1123) to a string that conforms.
+func ConvertToDNS1123CompatibleString(name string) string {
```
EngHabu, line 17: Is this strictly backward compatible? I'm worried about upgrade scenarios, if I've workflows in flight (pun intended) that launch a pod with name X, things will start failing if we start looking for pods under a different name X'...

EngHabu (commented):
Looks good! just a quick question

go/tasks/pluginmachinery/utils/dns.go
mayitbeegh: Yes it will break things if the pod name changes mid-flight. The assumption here is that if a name violates 1123, it will never start properly to begin with (and if it conforms to 1123, nothing changes) ... though, it's still a scary thought that it might break something. Would you recommend some test scenarios that I can run through?

EngHabu: One potential change is to run the validation here to make sure that if it would have been accepted, we don't try to touch it... can you triple check that DNS1123 is what they validate pod names against?

```suggestion
func ConvertToDNS1123CompatibleString(name string) string {

    if validation.IsDNS1123(name) {
        return name
    }
```

mayitbeegh: No official literature on this subject matter. https://kubernetes.io/docs/concepts/overview/working-with-objects/names/ vaguely mentioned the four types of commonly used name constraints for resources: "DNS Subdomain Names", "RFC 1123 Label Names", "RFC 1035 Label Names", and "Path Segment Names". The closest I could find (Google) was this page from Oracle https://docs.oracle.com/en/operating-systems/olcne/1.1/orchestration/kubectl-run.html that claims `Deployment, pod and service names conform to a requirement to match a DNS-1123 label. These must consist of lower case alphanumeric characters or -, and must start and end with an alphanumeric character. The regular expression that is used to validate names is '[a-z0-9]([-a-z0-9]*[a-z0-9])?'. If you use a name for your deployment that does not validate, an error is returned.`. This statement is either stale or inaccurate imo. In the current implementation of K8s, the pod name is validated against "DNS Subdomain Names" as shown here https://github.com/kubernetes/kubernetes/blob/4451138bfa784942e59dc092fc8d707e73eaa284/pkg/apis/core/validation/validation.go#L3969 and here https://github.com/kubernetes/kubernetes/blob/4451138bfa784942e59dc092fc8d707e73eaa284/pkg/apis/core/validation/validation.go#L229. The error message "a lowercase RFC 1123 subdomain must consist of lower case alphanumeric characters, '-' or '.', and must start and end with an alphanumeric character" matches the error message in https://github.com/flyteorg/flyte/issues/1172 . 
To conclude: no written statement on pod name validation, but evidences point to "DNS Subdomain Names" which is 253 characters max.

EngHabu (approved):
Thank you for doing the due diligence to keep it backward compatible!

PR #213, opened 2021-09-23
Cache Serialize API

hamersaw:
# TL;DR
Adding constructs to support reservations over cached objects. This is usefully to mitigate duplicate work over long running, frequent tasks. Specifically we added:
- PhaseWaitingForCache: to indicate that a task is currently waiting for the cache to be populated
- API extensions to datacatalog client for handling cache reservations

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
RFC - https://github.com/flyteorg/flyte/pull/1461

## Follow-up issue
_NA_


kumare3 (commented):
are there changes to be made for array tasks?

EngHabu:
That's a great question, @kumare3. I vote for no and call this out in the docs of this feature... that the way it works for array jobs is on the entire input of the array job as opposed to individual elements... the reason is that the array job code is complex enough that I don't think we should introduce additional complexity with this... until we move to array nodes...

EngHabu (approved):

go/tasks/pluginmachinery/catalog/client.go
```diff
@@ -73,10 +75,55 @@ func NewCatalogEntry(outputs io.OutputReader, status Status) Entry {
 	return Entry{outputs: outputs, status: status}
 }
 
+// Indicates the Reservation in Catalog
+type ReservationEntry struct {
```
EngHabu, line 79: Maybe add a few unit tests for this, admittedly trivial, file?

While we are at it, can we add docs on public types/funcs/methods?

go/tasks/pluginmachinery/catalog/client.go
```diff
@@ -73,10 +75,55 @@ func NewCatalogEntry(outputs io.OutputReader, status Status) Entry {
 	return Entry{outputs: outputs, status: status}
 }
 
+// Indicates the Reservation in Catalog
```
EngHabu, line 78: nit: golang-style ... it seems

```suggestion
// ReservationEntry indicates the Reservation in Catalog
```

go/tasks/pluginmachinery/catalog/client.go
hamersaw: done

go/tasks/pluginmachinery/catalog/client.go
hamersaw: done

EngHabu (approved):

PR #216, opened 2021-10-04
add support for batchScheduler in Spark Tasks

migueltol22:
Signed-off-by: Miguel <mtoledo@lyft.com>

# TL;DR
Spark has support for batchScheduler such as Volcano however even if it enabled in the spark operator it cannot currently be used because we don't set it as part of the sparkApp.

Here we allow a user to specify `spark.batchScheduler` as part of their config to leverage the batchScheduler option. This can also be enabled for all spark tasks as part of the default spark config an admin sets in the plugins config

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/go/tasks/plugins/k8s/spark/spark.go
+++ b/go/tasks/plugins/k8s/spark/spark.go
@@ -210,6 +210,10 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
 		},
 	}
 
+	if val, ok := sparkConfig["spark.batchScheduler"]; ok {
+		j.Spec.BatchScheduler = &val
+	}
+
 	if sparkJob.MainApplicationFile != "" {
 		j.Spec.MainApplicationFile = &sparkJob.MainApplicationFile
 	}


go/tasks/plugins/k8s/spark/spark_test.go
```diff
@@ -47,6 +47,7 @@ var (
 		"spark.flyte.feature1.enabled":    "true",
 		"spark.flyteorg.feature2.enabled": "true",
 		"spark.flyteorg.feature3.enabled": "true",
+		"spark.batchScheduler":            "volcano",
```
kumare3, line 50: why add it to spark-conf? Can we add it to just regular config? And I would prefer the config to say
scheduler: `default` - as default
and users can then suplement it as `scheduler`: `volcano`

kumare3: cc @akhurana001

migueltol22: Reason for spark conf is that for our use case we don't want to enable it for everyone, we'd like to try it out with a specific workflow. Adding it to regular config will set it for all unless we move to a config for each project/workflow. And I have no preference on whether `batchScheduler` or `scheduler`. Spark uses batch scheduler so figured I'd keep the same naming convention

PR #221, opened 2021-11-08
#minor Pluginmachinery support for Intra-task checkpointing

kumare3:
# TL;DR
Whenever Output file paths are created, Flytepropeller can now provision a destination directory for Checkpoints. The directory is just a sandbox and the usage of this is entirely dependendent on the client - e.g. FlyteKit. 

The protocol is as follows, 
 - FlytePropeller will replace any occurrence of `(?i){{\s*[\.$]CheckpointOutputPrefix\s*}}` with the new sandbox directory / prefix in remote Blob store.
 - Flytepropeller will replace any occurrence of `(?i){{\s*[\.$]PrevCheckpointPrefix\s*}}` with either the previous sandbox for checkpointing or an `""` empty string for attempts >= 1. 
 - The client is responsible to handle an empty `""` string in previous, but potentially even in current. Since this is a backend change, the client should also handle missing replacement (the existence of the original template string itself), as missing paths and continue gracefully
 - The client can create one file (replace it) or create multiple and the decision is to the client itself. The checkpoint is stored in the users preferred path and Flyte has no access to it

This change does not work without a corresponding Flytekit change

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
This PR implements a per execution Checkpointing system, that allows executions to recover from past unsuccessful executions, when retrying. The checkpointing and recovering is completely decided by the client system. 
Use cases for this could be
- Use to checkpoint training and recover in case of retries. Especially beneficial when using spot machines
- Use of checkpoint in spark jobs
- Any other custom checkpointing system that can reduce the penalty in retries and would help users in using Pre-emptible / spot instances much more easily

## Tracking Issue

https://github.com/flyteorg/flyte/issues/634

## Follow-up issue
_NA_


go/tasks/pluginmachinery/ioutils/paths.go
```diff
@@ -9,14 +9,35 @@ import (
...
+	ErrorsSuffix      = "error.pb"
+	IndexLookupSuffix = "indexlookup.pb"
+	// CheckpointPrefix specifies the common prefix that can be used as a directory where all the checkpoint information
+	// will be stored. This directory is under the raw output-prefix path
+	CheckpointPrefix = "_flytecheckpoints"
 )
 
+// ConstructCheckpointPath returns a checkpoint path under the given `base` / rawOutputPrefix, following the conventions of
+// the store
+func ConstructCheckpointPath(store storage.ReferenceConstructor, rawOutputPrefix storage.DataReference) storage.DataReference {
```
hamersaw, line 34: Do we need to add a unit test for this?

hamersaw (approved):

go/tasks/pluginmachinery/ioutils/paths.go
kumare3: i thought i did, if not, let me

kumare3: done

go/tasks/pluginmachinery/io/iface.go
```diff
@@ -9,12 +9,12 @@ import (
 
 //go:generate mockery -all -case=underscore
 
-// If using Files for IO with tasks, then the input will be written to this path
+// InputFilePaths If using Files for IO with tasks, then the input will be written to this path
```
EngHabu, line 12:
```suggestion
// InputFilePaths contains the different ways available for downstream systems to retrieve inputs. If using Files for IO with tasks, then the input will be written to this path
```

go/tasks/pluginmachinery/io/iface.go
```diff
@@ -26,54 +26,64 @@ type InputReader interface {
 	Get(ctx context.Context) (*core.LiteralMap, error)
 }
 
-// OutputReader
+// OutputReader provides an abstracted OutputReader interface. The plugins are responsible to provide
+// the implementations for the interface. Some helper implementations can be found in ioutils
 type OutputReader interface {
-	// Returns true if an error was detected when reading the output and false if no error was detected
+	// IsError Returns true if an error was detected when reading the output and false if no error was detected
```
EngHabu, line 32:
```suggestion
	// IsError returns true if an error was detected when reading the output and false if no error was detected
```

go/tasks/pluginmachinery/io/iface.go
```diff
@@ -26,54 +26,64 @@ type InputReader interface {
...
 	IsError(ctx context.Context) (bool, error)
-	// Returns the error as type ExecutionError
+	// ReadError Returns the error as type ExecutionError
```
EngHabu, line 34:
```suggestion
	// ReadError returns the error as type ExecutionError
```

go/tasks/pluginmachinery/io/iface.go
```diff
@@ -26,54 +26,64 @@ type InputReader interface {
...
 	Exists(ctx context.Context) (bool, error)
-	// Returns the output -> *core.LiteralMap (nil if void), *ExecutionError if user error when reading the output and error to indicate system problems
+	// Read Returns the output -> *core.LiteralMap (nil if void), *ExecutionError if user error when reading the output and error to indicate system problems
 	Read(ctx context.Context) (*core.LiteralMap, *ExecutionError, error)
 }
 
+// CheckpointPaths provides the paths / keys to input Checkpoints directory and an output checkpoints directory.
+type CheckpointPaths interface {
+	GetPreviousCheckpointsPrefix() storage.DataReference
+	GetCheckpointPrefix() storage.DataReference
```
EngHabu, line 48: Docs?

go/tasks/pluginmachinery/ioutils/paths.go
```diff
@@ -9,14 +9,35 @@ import (
...
+	if rawOutputPrefix == "" {
```
EngHabu, line 35: nit:
```suggestion
	if len(rawOutputPrefix) == 0 {
```

go/tasks/plugins/array/catalog.go
```diff
@@ -434,7 +434,8 @@ func ConstructOutputWriter(ctx context.Context, dataStore *storage.DataStore, ou
 		return nil, err
 	}
 
-	p := ioutils.NewRemoteFileOutputPaths(ctx, dataStore, dataReference, ioutils.NewRawOutputPaths(ctx, outputSandbox))
+	// TODO when we fix https://github.com/flyteorg/flyte/issues/1276 we should make so that the checkpoint paths are computed correctly
```
EngHabu, line 437:
```suggestion
	// TODO when we fix https://github.com/flyteorg/flyte/issues/1276 we should make sure that the checkpoint paths are computed correctly
```

go/tasks/plugins/k8s/container/container.go
```diff
@@ -58,7 +58,7 @@ func (Plugin) GetTaskPhase(ctx context.Context, pluginContext k8s.PluginContext,
 	return pluginsCore.PhaseInfoRunning(pluginsCore.DefaultPhaseVersion, &info), nil
 }
 
-// Creates a new Pod that will Exit on completion. The pods have no retries by design
+// BuildResource Creates a new Pod that will Exit on completion. The pods have no retries by design
```
EngHabu, line 61:
```suggestion
// BuildResource creates a new Pod that will Exit on completion. The pods have no retries by design
```

EngHabu (commented):
nits

EngHabu (approved):

EngHabu (dismissed):

EngHabu (approved):

PR #223, opened 2021-11-30
#minor Global Pod Security Context and Security context support for all pods

kumare3:
# TL;DR
This PR introduces 2 new configuration
1. `k8s.default-pod-security-context`
2. `k8s.default-security-context`
which can be configured per propeller instances and will apply to all pods created by that instance. If CRDs are created then, the CRD should support passing these parameters

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Added 2 new flags. Added support for pod/container creation and spark job creation.
Follow-up update the spark dependency to add support for pod and container security context (currently only pod security context will be supported).

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/1854

## Follow-up issue
_NA_


kumare3:
cc @akumor

go/tasks/pluginmachinery/flytek8s/config/config.go
```diff
@@ -128,9 +130,17 @@ type K8sPluginConfig struct {
...
 
 	// The name of the GPU resource to use when the task resource requests GPUs.
-	GpuResourceName v1.ResourceName `json:"gpu-resource-name" pflag:",The name of the GPU resource to use when the task resource requests GPUs."`
+	GpuResourceName v1.ResourceName `json:"gpu-resource-name" pflag:"-,The name of the GPU resource to use when the task resource requests GPUs."`
+
+	// DefaultPodSecurityContext provides a default pod security context that should be applied for every pod that is launched by FlytePropeller
+	DefaultPodSecurityContext *v1.PodSecurityContext `json:"default-pod-security-context" pflag:"-,Optionally specify any default pod security context that should be applied to every Pod launched by FlytePropeller."`
+
+	// DefaultSecurityContext provides a default container security context that should be applied for every container launched by FlytePropeller
+	DefaultSecurityContext *v1.SecurityContext `json:"default-security-context" pflag:"-,Optionally specify a default security context that should be applied to every container launched by FlytePropeller."`
```
akhurana001, line 139: do we really need both if this is anyways defaulted at platform level right now ? The PodSecurityContext will anyways be applied to all containers in every pod created by FlytePropeller ?

kumare3: is that really true?

kumare3: So i guess having both has one benefit, you can simply add ContainerSecurityContext and that will be applied only to your container not to the injected containers. While podSecuritycontext will be applied to everything.
If you specify both, then the resultant will be a combination. WDYT?

go/tasks/pluginmachinery/flytek8s/container_helper.go
```diff
@@ -220,6 +220,9 @@ func ToK8sContainer(ctx context.Context, taskContainer *core.Container, iFace *c
 	if err := AddCoPilotToContainer(ctx, config.GetK8sPluginConfig().CoPilot, container, iFace, taskContainer.DataConfig); err != nil {
 		return nil, err
 	}
+	if container.SecurityContext == nil && config.GetK8sPluginConfig().DefaultSecurityContext != nil {
```
EngHabu, line 223: Should we do this for all containers that we find in PodTask plugin? not just the main container?

go/tasks/pluginmachinery/flytek8s/config/config.go
akhurana001: I guess that makes sense. Having both will probably cause some confusion/raise questions so let's  document the overlap in behavior here.

go/tasks/plugins/k8s/spark/spark.go
```diff
@@ -99,20 +99,22 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
...
 		SparkPodSpec: sparkOp.SparkPodSpec{
-			Annotations: annotations,
-			Labels:      labels,
-			EnvVars:     sparkEnvVars,
-			Image:       &container.Image,
+			Annotations:      annotations,
+			Labels:           labels,
+			EnvVars:          sparkEnvVars,
+			Image:            &container.Image,
+			SecurityContenxt: config.GetK8sPluginConfig().DefaultPodSecurityContext.DeepCopy(),
```
akhurana001, line 106: Let's use the corresponding `podSecurityContext` and `securityContext` in SparkApplication based on the conf: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/master/docs/user-guide.md#using-security-context

go/tasks/pluginmachinery/flytek8s/container_helper.go
kumare3: I don't think so, as we allow this to be added for other containers? Seems wrong

EngHabu (approved):

wild-endeavor (approved):

PR #225, opened 2021-12-07
Merge k8s container and sidecar plugins

hamersaw:
# TL;DR
Merge the container and sidecar plugins into a single pod plugin.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?
 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
In addition to providing similar functionality, the container and sidecar plugins share a large amount of code. This is more prominent in retrieving task phases, where they are almost identical. This PR merges these two plugins in a single, pod plugin. We provide an abstraction for podBuilders, with unique container and sidecar implementations, which define how the pod should be constructed. This functionality should ease development of additional pod building paradigms.

A brief discussion of a few implementation decisions to fast track discussion:
- non-existent primary container name: I found that when we try to set the primary container name to the specified container on the container plugin we get an "output not found" error, even though the output.pb file exists in minio. In debugging, the primary container completes and before the copilot container can finish (and write the output) the task is marked as successful and moved on. It seems messy to set the primary container to the copilot container for all container plugin types. Instead, we elected to fallback to setting the status to "running" if the primary container annotation does not exist. This is identical to the existing container plugin implementation so there are no issues. Additionally, it seems there may be additional podBuilders that could use this funtionality.
- plugin default task types and system default: The existing container plugin is set as the system default (for which there can only be one) and it must be so because flytekit seems to set the task type to "raw-container" instead of "container" as it is set in the container plugin. Therefore, every call to the container plugin actually falls back to the default plugin instead of the container plugin lookup. Transitioning a deployment away from the container and sidecar plugins will require addition of the "pod" plugin and removal of the "container" and "sidecar" plugins in FlytePropeller configuration to work correctly. Otherwise, there will be issues with the system default plugin and default task type plugins. Only after deployments are transitioned can we remove the separate container and sidecar plugins from the flyteplugins repo. I'm not sure about the easiest way to advertise this.

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/1837

## Follow-up issue
_NA_


EngHabu (commented):
Awesome! Do you mind adding 
```
// deprecated: Please use PodPlugin instead
```
On top of the Container and Sidecar plugins?

Alternatively, you can duplicate [these lines](https://github.com/flyteorg/flyteplugins/pull/225/files#diff-c4e3100eaecdfaf3780665ab2b39e4603908afc4046d893d74249389f81e21f1R132-R143) using the older plugin names but using the new PodPlugin as the pluginEntry... this way you can go ahead and delete the other plugins in the same PR, have one code path to worry about and not require migration...

EngHabu (approved):
fantastic

PR #228, opened 2021-12-17
Add Raw AWS Batch Task

pingsutw:

# TL;DR
Run regular tasks on AWS batch service
check example in flyteorg/flytekit#782

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1792
https://github.com/flyteorg/flytekit/pull/782

## Follow-up issue
_NA_


go/tasks/plugins/array/awsbatch/client.go
```diff
@@ -68,22 +73,25 @@ func (b client) GetAccountID() string {
 }
 
 // Registers a new job definition. There is no deduping on AWS side (even for the same name).
-func (b *client) RegisterJobDefinition(ctx context.Context, name, image, role string) (arn definition2.JobDefinitionArn, err error) {
+func (b *client) RegisterJobDefinition(ctx context.Context, name, image, role string, structObj *structpb.Struct) (arn definition2.JobDefinitionArn, err error) {
 	logger.Infof(ctx, "Registering job definition with name [%v], image [%v], role [%v]", name, image, role)
+	jobDefinition := batch.RegisterJobDefinitionInput{}
+	if structObj != nil {
+		err = pluginUtils.UnmarshalStructToObj(structObj, &jobDefinition)
```
kumare3, line 80: what happens if structObj is nil? We will get null pointer exception on line 85?

kumare3: we should check if nil, then return an error?

kumare3:
mostly looks good to me, but, looks a little more complicated because of the array case

go/tasks/plugins/array/awsbatch/client.go
pingsutw: We create a `jobDefinition` object at line 78, so `jobDefinition` won't be nil at line 85. correct me if I'm wrong.

btw, check the test [here](https://github.com/flyteorg/flyteplugins/pull/228/files#diff-39aec2e3bb82fc78898943f508e5b70c528166f279c8a931f1cc37e20e6a48f0R92-R94)

go/tasks/plugins/array/awsbatch/definition/job_def_cache.go
```diff
@@ -27,8 +29,9 @@ type CacheKey interface {
 }
 
 type cacheKey struct {
-	role  string
-	image string
+	role      string
+	image     string
+	structObj *structpb.Struct
```
hamersaw, line 34: What is the reason to store the structObj in the cacheKey? I don't think I see it being used anywhere.

go/tasks/plugins/array/catalog.go
```diff
@@ -21,6 +23,8 @@ import (
 	idlCore "github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
 )
 
+const awsBatchTaskType = "aws-batch"
```
hamersaw, line 26: Is there anyway we can define this once and use it everywhere? seems like defining it here and in go/tasks/plugins/array/awsbatch/executor.go makes it more difficult to track.

go/tasks/plugins/array/awsbatch/client.go
```diff
@@ -9,6 +9,11 @@ import (
 	"context"
 	"fmt"
 
+	"github.com/pkg/errors"
```
hamersaw, line 12: nit, rearrange imports to be alphabetical and grouped by repository

```suggestion
	a "github.com/aws/aws-sdk-go/aws"
	"github.com/aws/aws-sdk-go/aws/request"
	"github.com/aws/aws-sdk-go/service/batch"

	"github.com/flyteorg/flyteplugins/go/tasks/aws"
	pluginUtils "github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/utils"
	definition2 "github.com/flyteorg/flyteplugins/go/tasks/plugins/array/awsbatch/definition"

	"github.com/flyteorg/flytestdlib/utils"
	"github.com/flyteorg/flytestdlib/logger"

	structpb "github.com/golang/protobuf/ptypes/struct"
	
	"github.com/pkg/errors"
```

go/tasks/plugins/array/outputs.go
```diff
@@ -185,10 +196,11 @@ func AssembleFinalOutputs(ctx context.Context, assemblyQueue OutputAssembler, tC
...
 		err = assemblyQueue.Queue(ctx, workItemID, &outputAssembleItem{
-			varNames:    varNames,
-			finalPhases: finalPhases,
-			outputPaths: tCtx.OutputWriter(),
-			dataStore:   tCtx.DataStore(),
+			varNames:       varNames,
+			finalPhases:    finalPhases,
+			outputPaths:    tCtx.OutputWriter(),
+			dataStore:      tCtx.DataStore(),
+			isAwsSingleJob: taskTemplate.Type == awsBatchTaskType,
```
hamersaw, line 203: I don't fully understand this, probably because I haven't fully followed this feature. The one-off batch task as defined in the [flytekit pr](https://github.com/flyteorg/flytekit/pull/782) will execute a single job using the AWS batch API and that's why were setting isAwsSingleJob here? we are just reusing the awsbatch array implementation because it already works?

hamersaw (commented):
Few questions regarding implementation and a couple nits. I think it looks great!

go/tasks/plugins/array/awsbatch/definition/job_def_cache.go
pingsutw: sorry, my bad. 
I just updated it, and we will use `role`, `image`, `jobDefinitionInput` as cache key

go/tasks/plugins/array/catalog.go
pingsutw: fixed it, thanks.

go/tasks/plugins/array/outputs.go
pingsutw: yeah, exactly.
we already support running map tasks (array job) on AWS batch, but we can't run regular tasks before.
There are only a few differences:

1. Arrayjob will aggregate output.pb in different directories, but the single job won't. check [here](https://github.com/flyteorg/flyteplugins/pull/228/files#diff-0edeed4ecb7475faa7b319bd5ae0fc578872fb3acd0a20cd0f6cdffcfb914ddeR79-R87)
2. We only need one input reader for a single job instead of a list of readers. check [here](https://github.com/flyteorg/flyteplugins/blob/892ba2cc7401448b9610bfbe5e89507bd5a5cbdd/go/tasks/plugins/array/catalog.go#L65-L97)

pingsutw: will update docs in flyte/flytesnack

go/tasks/plugins/array/awsbatch/definition/job_def_cache.go
kumare3: ohh that is a good catch, it would have failed the cache lookup

go/tasks/plugins/array/awsbatch/definition/job_def_cache.go
```diff
@@ -27,12 +29,13 @@ type CacheKey interface {
 }
 
 type cacheKey struct {
-	role  string
-	image string
+	role               string
+	image              string
+	jobDefinitionInput batch.RegisterJobDefinitionInput
```
kumare3, line 34: hmm this is a process side cache, the cache key could be very large. Is there a simpler way?

go/tasks/plugins/array/awsbatch/definition/job_def_cache.go
```diff
@@ -27,12 +29,13 @@ type CacheKey interface {
...
 }
 
 func (k cacheKey) String() string {
-	return fmt.Sprintf("%v-%v", k.image, k.role)
+	return fmt.Sprintf("%v-%v-%v", k.image, k.role, k.jobDefinitionInput.String())
```
kumare3, line 38: I think we should hash it? maybe fnv, otherwise this will be large memory overhead.

kumare3: cc @EngHabu do we need the entire definition? It looks like name should be enough [here](https://docs.aws.amazon.com/batch/latest/userguide/create-job-definition.html)

pingsutw: but the different tasks have the same job definition name now. The name is the same as the image tag.
If image tag is `pingsutw/flyte-app:b22d5edf6b1777a2bcfeef2465e6a7a6`, the job definition will be `flyte-app`

pingsutw: btw, should we replace freecache with bigcache. It seems like bigcache is faster than freecache. check this [benchmark](https://github.com/allegro/bigcache#benchmarks)
If so, I can create PR for this.

go/tasks/plugins/array/awsbatch/transformer.go
```diff
@@ -85,14 +87,22 @@ func FlyteTaskToBatchInput(ctx context.Context, tCtx pluginCore.TaskExecutionCon
...
 		JobName:            refStr(tCtx.TaskExecutionMetadata().GetTaskExecutionID().GetGeneratedName()),
 		JobDefinition:      refStr(jobDefinition),
 		JobQueue:           refStr(jobConfig.DynamicTaskQueue),
 		RetryStrategy:      toRetryStrategy(ctx, toBackoffLimit(taskTemplate.Metadata), cfg.MinRetries, cfg.MaxRetries),
 		ContainerOverrides: toContainerOverrides(ctx, append(cmd, args...), &resources, envVars),
 		Timeout:            toTimeout(taskTemplate.Metadata.GetTimeout(), cfg.DefaultTimeOut.Duration),
-	}, nil
+	}
+	if taskTemplate.GetCustom() != nil {
+		err = utils.UnmarshalStructToObj(taskTemplate.GetCustom(), &submitJobInput)
```
EngHabu, line 99: This is a UX question: should we unmarshal first then override with the things above?
```
@task(config=AWSBatch(jobName="blah"))
def task():
  ...
```

Should this use that jobName as opposed to the generated name? I don't think so... 

Then this:
```
@task(config=AWSBatch(retries=aws.RetryStrategy(count=3)),
            retries=2)
def task():
  ...
```

Which one should win here? not sure... we maybe should fail?

I want to say let's unmarshal first the override with the properties we computed... we can add code that says things like: "If retry is set in both, log a warning"

pingsutw: The SubmitJobInput properties we computed in the propeller are not included in the [flytekit aws batch config](https://github.com/flyteorg/flytekit/pull/782/files#diff-6126197e8a9fe3df591b719012cd0ef2c32cb87fc027d9d68d5aac7e9c5835abR21-R27)  except `RetryStrategy` and `Timeout`.  
Do we still need to unmarshal first? I think it's fine to override these two properties by flytekit.
or Do we need to also remove RetryStrategy and Timeout in flytekit

EngHabu: aha... I would say let's remove them from flytekit then... and keep the common properties consistent with the other flyte tasks...

pingsutw: done

go/tasks/plugins/array/awsbatch/job_definition.go
```diff
@@ -16,6 +16,9 @@ import (
 	"github.com/flyteorg/flyteplugins/go/tasks/plugins/array/awsbatch/definition"
 )
 
+const defaultComputeEngine = "EC2"
+const platformCapabilities = "platformCapabilities"
```
EngHabu, line 20: nit: let's try to avoid [shadowing variables](https://yourbasic.org/golang/gotcha-shadowing-variables/) in general. It leads to confused unhappy maintainers...
Maybe name it something like:
```suggestion
const platformCapabilitiesConfigKey = "platformCapabilities"
```

go/tasks/plugins/array/awsbatch/transformer.go
EngHabu: Let's revert ordering here to make this protocol strict... even if we, in some future, added these properties to flytekit, they should get overriden here... or maybe somebody produced the task outside of flytekit... or maybe flytekit-java doesn't do it correctly... etc. etc... let's keep this protocol tight

EngHabu (commented):
Thanks @pingsutw almost there... I've a couple more comments!

EngHabu (approved):
Thank you @pingsutw

PR #231, opened 2022-01-03
Track RetryAttempt and Phase of ExternalResources

hamersaw:
# TL;DR
Populating RetryAttempt and Phase on ExternalResources to track array subtasks.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added
 - [x] Code documentation added
 - [x] Any pending items have an associated Issue

## Complete description
Created a new [ExternalResources](https://github.com/flyteorg/flyteplugins/blob/de5b4773187ced71b133e822bccc3fc3c8a85f16/go/tasks/pluginmachinery/core/phase.go#L70) struct which contains metadata on, you may have guessed, external resources which replaces the Metadata protobuf stored in [TaskInfo](https://github.com/flyteorg/flyteplugins/blob/973a51ae8cc6a06e2f78007fe1ffda71f7355fad/go/tasks/pluginmachinery/core/phase.go#L81). This decision was made for two reasons (1) the phase maintained within flyteplugins [for each subtask](https://github.com/flyteorg/flyteplugins/blob/973a51ae8cc6a06e2f78007fe1ffda71f7355fad/go/tasks/plugins/array/k8s/monitor.go#L87) is a flyteplugins core.Phase rather than the TaskExecution.Phase from flyteidl (which is reported in the TaskExecutionEvent) and (2) in flytepropeller all the Metadata fields except external resources [are being overwritten](https://github.com/flyteorg/flytepropeller/blob/51c85b7b842e51489b158fb7dcddaefd04f98dec/pkg/controller/nodes/task/transformer.go#L105), plus we need to convert between the flyteplugin and flyteidl phases.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/1838

## Follow-up issue
_NA_

PR #236, opened 2022-01-28
Retry map task subtasks

hamersaw:
# TL;DR
Currently, k8s array tasks (map tasks) retry the entire collection of subtasks when one of them fails. This PR enables retries over individual subtasks.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
A previous [PR](https://github.com/flyteorg/flyteplugins/pull/231) added tracking of retry attempts for each subtask. As the current implementation retries all subtasks if one fails, this wasn't particularly useful because they were all the same. With this PR we enable subtasks to be retried individually, using this retry attempts array to inform unique pod names and logs for each subtask retry.

When a subtask fails, the current implementation reports the failure (as retryable) to the parent map task. Now we attempt to retry by incrementing the retry attempt value and transitioning the subtask to the "Undefined" phase as if it had never been executed. This is recognized during the next evaluation and another attempt is executed.

We use the max attempts defined on the parent map task to determine the number of times a subtask may be executed. As we track subtask retries internally (rather than within the parent map task), if a subtask exceeds the maximum number of retries we report a permanent failure to the parent map task. This ensure that the entire collection of subtasks is not retried again (with each subtask potentially retried multiple more times). Other retryable failures reported to the map task will still fallthrough.

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/1276

## Follow-up issue
- https://github.com/flyteorg/flyte/issues/1533: This solution does not fix any issues with interruptible nodes within k8s array tasks. Will have to be addressed separately. Discerning what is an interrupted failure is very difficult, the current approach handles all "system" failures as interrupted and enables a different configurable number of retries. We might need to track the failure type as well?

### DIFF
--- a/go/tasks/plugins/array/k8s/launcher.go
+++ b/go/tasks/plugins/array/k8s/launcher.go
@@ -3,6 +3,7 @@ package k8s
 import (
 	"context"
 	"fmt"
+	"strconv"
 
 	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/utils"
 
@@ -33,7 +34,16 @@ var arrayJobEnvVars = []corev1.EnvVar{
 	},
 }
 
-func formatSubTaskName(_ context.Context, parentName, indexStr, retryAttemptStr string) (subTaskName string) {
+func formatSubTaskName(_ context.Context, parentName string, index int, retryAttempt uint64) (subTaskName string) {
+	indexStr := strconv.Itoa(index)
+
+	// If the retryAttempt is 0 we do not include it in the pod name. The gives us backwards
+	// compatibility in the ability to dynamically transition running map tasks to use subtask retries.
+	if retryAttempt == 0 {
+		return utils.ConvertToDNS1123SubdomainCompatibleString(fmt.Sprintf("%v-%v", parentName, indexStr))
+	}
+
+	retryAttemptStr := strconv.FormatUint(retryAttempt, 10)
 	return utils.ConvertToDNS1123SubdomainCompatibleString(fmt.Sprintf("%v-%v-%v", parentName, indexStr, retryAttemptStr))
 }
 
--- a/go/tasks/plugins/array/k8s/launcher.go
+++ b/go/tasks/plugins/array/k8s/launcher.go
@@ -33,8 +33,8 @@ var arrayJobEnvVars = []corev1.EnvVar{
 	},
 }
 
-func formatSubTaskName(_ context.Context, parentName, suffix string) (subTaskName string) {
-	return utils.ConvertToDNS1123SubdomainCompatibleString(fmt.Sprintf("%v-%v", parentName, suffix))
+func formatSubTaskName(_ context.Context, parentName, indexStr, retryAttemptStr string) (subTaskName string) {
+	return utils.ConvertToDNS1123SubdomainCompatibleString(fmt.Sprintf("%v-%v-%v", parentName, indexStr, retryAttemptStr))
 }
 
 func ApplyPodPolicies(_ context.Context, cfg *Config, pod *corev1.Pod) *corev1.Pod {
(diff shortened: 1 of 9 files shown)


go/tasks/plugins/array/k8s/launcher.go
```diff
@@ -33,8 +33,8 @@ var arrayJobEnvVars = []corev1.EnvVar{
 	},
 }
 
-func formatSubTaskName(_ context.Context, parentName, suffix string) (subTaskName string) {
-	return utils.ConvertToDNS1123SubdomainCompatibleString(fmt.Sprintf("%v-%v", parentName, suffix))
+func formatSubTaskName(_ context.Context, parentName, indexStr, retryAttemptStr string) (subTaskName string) {
```
EngHabu, line 36: This name is used as the Pod name, right? Won't this break backward compatibility for in-flight (pun intended) tasks? we might need to use [State Version](https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/pluginmachinery/core/state.go#L15) to differentiate between "before" and "after" upgrade...

EngHabu: nit: Looks like we always have the retry attempt as a uint and we convert it to a string, should we move the logic to do the uint -> str here instead?

EngHabu (commented):
I think it looks good! let's just make it backward compatible and do some testing in our demo environment...

go/tasks/plugins/array/k8s/launcher.go
hamersaw: @EngHabu in diving through backwards-compatibility for in-flight (love the pun!) tasks I think this is challenging. The difficulty stems from the transition from reporting and tracking retryable failures in the task execution context to the subtask retryAttempts array. Since the pod name is contingent on the retry and these may be incremented independently there are many situations which cause pod name overlap (same name for different retry attempts) if we are not careful.

I've convinced myself there is no way to transition a running map_task (ie. where at least one pod is currently running) to this proposed approach (ie. subtask retryAttempts array). Therefore, any solution requires a collection of if statements checking the state_version to allow currently running map tasks to continue using podnames based solely on the task execution context retry attempt and allowing retryable failures to fall-through to propeller which can fail the entire map task. I think this introduces significant complexity to the codebase, where this logic is already complex enough.

Is there a simple solution that I'm overthinking?

hamersaw: nvm, figured it out.

go/tasks/plugins/array/k8s/launcher.go
```diff
@@ -33,8 +34,17 @@ var arrayJobEnvVars = []corev1.EnvVar{
...
-func formatSubTaskName(_ context.Context, parentName, suffix string) (subTaskName string) {
-	return utils.ConvertToDNS1123SubdomainCompatibleString(fmt.Sprintf("%v-%v", parentName, suffix))
+func formatSubTaskName(_ context.Context, parentName string, index int, retryAttempt uint64) (subTaskName string) {
+	indexStr := strconv.Itoa(index)
+
+	// If the retryAttempt is 0 we do not include it in the pod name. The gives us backwards
+	// compatibility in the ability to dynamically transition running map tasks to use subtask retries.
+	if retryAttempt == 0 {
+		return utils.ConvertToDNS1123SubdomainCompatibleString(fmt.Sprintf("%v-%v", parentName, indexStr))
+	}
```
EngHabu, line 44: Cool!

EngHabu (dismissed):
Looking great!

EngHabu:
Can we add some more unit tests to up the coverage for this patch?

hamersaw:
> Can we add some more unit tests to up the coverage for this patch?

Done.

EngHabu (approved):

PR #237, opened 2022-01-30
Add support for handling of node interruption events in Kubernetes plugin

jeevb:
# TL;DR
This PR adds detection of pod failures due to node interruption, initially on GKE >= 1.20. This is required for `interruptible-failure-threshold` to be effective.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Refactored `ConvertPodFailureToError` to `DemystifyFailure` to be consistent with other methods, and return `PhaseInfoSystemRetryableFailure` on pod failures due to interruption.

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/boilerplate/flyte/golang_support_tools/tools.go
+++ b/boilerplate/flyte/golang_support_tools/tools.go
@@ -1,3 +1,4 @@
+//go:build tools
 // +build tools
 
 package tools
@@ -6,6 +7,6 @@ import (
 	_ "github.com/alvaroloes/enumer"
 	_ "github.com/flyteorg/flytestdlib/cli/pflags"
 	_ "github.com/golangci/golangci-lint/cmd/golangci-lint"
-	_ "github.com/vektra/mockery/cmd/mockery"
 	_ "github.com/pseudomuto/protoc-gen-doc/cmd/protoc-gen-doc"
+	_ "github.com/vektra/mockery/cmd/mockery"
 )
--- a/go/tasks/pluginmachinery/flytek8s/pod_helper.go
+++ b/go/tasks/pluginmachinery/flytek8s/pod_helper.go
@@ -347,10 +347,11 @@ func DeterminePrimaryContainerPhase(primaryContainerName string, statuses []v1.C
 		fmt.Sprintf("Primary container [%s] not found in pod's container statuses", primaryContainerName), info)
 }
 
-// ConvertPodFailureToError retruns a legible error message and code from a failed v1.PodStatus field
-func ConvertPodFailureToError(status v1.PodStatus) (code, message string) {
-	code = "UnknownError"
-	message = "Pod failed. No message received from kubernetes."
+// DemystifyFailure resolves the various Kubernetes pod failure modes to determine
+// the most appropriate course of action
+func DemystifyFailure(status v1.PodStatus, info pluginsCore.TaskInfo) (pluginsCore.PhaseInfo, error) {
+	code := "UnknownError"
+	message := "Pod failed. No message received from kubernetes."
 	if len(status.Reason) > 0 {
 		code = status.Reason
 	}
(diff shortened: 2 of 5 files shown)


jeevb:
Note that this PR did not change the way `Interrupted` detection via container `SIGKILL` is handled, but perhaps it should. Can someone comment on this? Would we not expect a SIGINT first, and a SIGKILL as a last resort? If so, this might be brittle?

boilerplate/flyte/golang_support_tools/tools.go
```diff
@@ -1,3 +1,4 @@
+//go:build tools
```
jeevb, line 1: Should this be dropped? It looks like it was auto-generated.

EngHabu: This is ok...

EngHabu (approved):
Thank you!

kumare3:
> Note that this PR did not change the way `Interrupted` detection via container `SIGKILL` is handled, but perhaps it should. Can someone comment on this? Would we not expect a SIGINT first, and a SIGKILL as a last resort? If so, this might be brittle?

@jeevb is this in flytekit?

jeevb:
> > Note that this PR did not change the way `Interrupted` detection via container `SIGKILL` is handled, but perhaps it should. Can someone comment on this? Would we not expect a SIGINT first, and a SIGKILL as a last resort? If so, this might be brittle?
> 
> @jeevb is this in flytekit?

I am not aware of specific handling in Flytekit. In the absence of a handler for the initial SIGINT, it does make sense to receive a SIGKILL after the termination grace period. Is that the intended behavior? If so, perhaps we should catch and handle this as system retryable failures as well.

hamersaw:
> > Note that this PR did not change the way `Interrupted` detection via container `SIGKILL` is handled, but perhaps it should. Can someone comment on this? Would we not expect a SIGINT first, and a SIGKILL as a last resort? If so, this might be brittle?
> 
> @jeevb is this in flytekit?

@kumare3 (cc @jeevb) I think this refers to setting the return code to "Interrupted" when a container is terminated with SIGKILL [here](https://github.com/flyteorg/flyteplugins/blob/0637c345b54ec62277fbd6e47d5b04cbc7ee5929/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L373).

hamersaw:
> > > Note that this PR did not change the way `Interrupted` detection via container `SIGKILL` is handled, but perhaps it should. Can someone comment on this? Would we not expect a SIGINT first, and a SIGKILL as a last resort? If so, this might be brittle?
> > 
> > 
> > @jeevb is this in flytekit?
> 
> I am not aware of specific handling in Flytekit. In the absence of a handler for the initial SIGINT, it does make sense to receive a SIGKILL after the termination grace period. Is that the intended behavior? If so, perhaps we should catch and handle this as system retryable failures as well.
> 
> It does feel a bit brittle to handle this via container exit codes. How can we be sure that sidecar containers don’t exit with SIGKILL during events unrelated to interruptions? My intuition is that this would still be flagged as “Interrupted” by the sidecar plugin.

All good points. Currently, identifying interrupted pods is obviously "best effort". I think it should be noted that the interruptible threshold in propeller is [handled over all system failures](https://github.com/flyteorg/flytepropeller/blob/a3c6e91f19c19601a957b29891437112868845de/pkg/controller/nodes/node_exec_context.go#L213-L217) rather than those flagged as only interrupted (perhaps another change that should be made here?). This solution is again, quite brittle.

Without an implicit flag identifying a pod as interrupted we are forced to accept either false positives or false negatives. Perhaps less common than terminating a container with SIGKILL, a k8s deployment could set the pod status to "Shutdown" without it being interrupted. In this context, I think it makes sense to risk misidentifying a failure as interrupted (so that it counts towards the interrupted threshold) rather than not (where it counts towards the retry count). Thoughts?

jeevb:
> All good points. Currently, identifying interrupted pods is obviously "best effort". I think it should be noted that the interruptible threshold in propeller is [handled over all system failures](https://github.com/flyteorg/flytepropeller/blob/a3c6e91f19c19601a957b29891437112868845de/pkg/controller/nodes/node_exec_context.go#L213-L217) rather than those flagged as only interrupted (perhaps another change that should be made here?). This solution is again, quite brittle.
> 
> Without an implicit flag identifying a pod as interrupted we are forced to accept either false positives or false negatives. Perhaps less common than terminating a container with SIGKILL, a k8s deployment could set the pod status to "Shutdown" without it being interrupted. In this context, I think it makes sense to risk misidentifying a failure as interrupted (so that it counts towards the interrupted threshold) rather than not (where it counts towards the retry count). Thoughts?

Agree with all the points, @hamersaw! I find that handling interruptions is not a perfect science. For example, in the implementation we were using before Flyte, we kept track of the scheduled time of every node. We rescheduled interruptible workloads on non-interruptible nodes when any of the following conditions were met:
1. >= N interruptions (configurable)
2. Any system failure (not just interruptions) after (current time - first scheduled time) exceeds N minutes (configurable)
Again, just a "best effort" way of pushing workloads forward. Perhaps, we can aspire to better interruption handling overall in Flyte as well.

With respect to detecting interruption from pod status, my intuition is that we may not have a proper solution until there is a standard set somewhere. I can understand why Kubernetes may not want to worry about setting this standard, but perhaps it should?

hamersaw (approved):

PR #239, opened 2022-02-02
Track parallelism in k8s array plugin

hamersaw:
# TL;DR
Currently the concurrency / parallelism parameter on the k8s array plugin is non-functional. This means concurrency controls on map tasks do not work as expected. This PR enables tracking and validating parallelism constraints during k8s array plugin evaluations by tracking non-terminal subtask phases.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/2121

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/plugins/array/k8s/monitor.go
+++ b/go/tasks/plugins/array/k8s/monitor.go
@@ -192,13 +192,13 @@ func LaunchAndCheckSubTasksState(ctx context.Context, tCtx core.TaskExecutionCon
 		// validate map task parallelism
 		newSubtaskPhase := core.Phases[newArrayStatus.Detailed.GetItem(childIdx)]
 		if !newSubtaskPhase.IsTerminal() || newSubtaskPhase == core.PhaseRetryableFailure {
-			currentParallelism += 1
+			currentParallelism++
 		}
 
 		if maxParallelism != 0 && currentParallelism >= maxParallelism {
 			// If max parallelism has been achieved we need to fill the subtask phase summary with
-			// the remaining subtasks so the overall map task phase can be accurately identified. 
-			for i := childIdx+1; i<len(currentState.GetArrayStatus().Detailed.GetItems()); i++ {
+			// the remaining subtasks so the overall map task phase can be accurately identified.
+			for i := childIdx + 1; i < len(currentState.GetArrayStatus().Detailed.GetItems()); i++ {
 				childSubtaskPhase := core.Phases[newArrayStatus.Detailed.GetItem(i)]
 				newArrayStatus.Summary.Inc(childSubtaskPhase)
 			}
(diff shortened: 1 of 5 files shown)


go/tasks/plugins/array/k8s/monitor.go
```diff
@@ -175,18 +188,27 @@ func LaunchAndCheckSubTasksState(ctx context.Context, tCtx core.TaskExecutionCon
...
-	taskTemplate, err := tCtx.TaskReader().Read(ctx)
-	if err != nil {
-		return currentState, logLinks, subTaskIDs, err
-	} else if taskTemplate == nil {
-		return currentState, logLinks, subTaskIDs, fmt.Errorf("required value not set, taskTemplate is nil")
+		if maxParallelism != 0 && currentParallelism >= maxParallelism {
+			// If max parallelism has been achieved we need to fill the subtask phase summary with
+			// the remaining subtasks so the overall map task phase can be accurately identified.
+			for i := childIdx + 1; i < len(currentState.GetArrayStatus().Detailed.GetItems()); i++ {
+				childSubtaskPhase := core.Phases[newArrayStatus.Detailed.GetItem(i)]
```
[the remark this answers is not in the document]
hamersaw: it will happen in future rounds as you described. as implemented, this will just stop processing subtasks when we reach the max parallelism. however, without populating the summary array [calculating the map task phase](https://github.com/flyteorg/flyteplugins/blob/14ed6a8d2848fd200ac67f6bc35766ce48377b62/go/tasks/plugins/array/core/state.go#L256-L304) has unintended consequences. for example, failing because there are not enough tasks to achieve the minSuccesses.

go/tasks/plugins/array/k8s/monitor.go
```diff
@@ -91,6 +88,22 @@ func LaunchAndCheckSubTasksState(ctx context.Context, tCtx core.TaskExecutionCon
 		return currentState, logLinks, subTaskIDs, err
 	}
 
+	// identify max parallelism
+	taskTemplate, err := tCtx.TaskReader().Read(ctx)
```
EngHabu, line 92: So this function gets called every time we check for task status... I would try to avoid rereading the task template over and over again... any chance we can do away with that (in a backward compatible way too)?

EngHabu (commented):
Looks great! I've one question about reading TaskTemplates...

go/tasks/plugins/array/k8s/monitor.go
kumare3: @EngHabu so the task reader is cached. You are worried about a bad pattern?

EngHabu: aha... I've not noticed that this block existed but just moved around... It looks like it was a remnant of some refactoring this went through...

I was think of moving it to task state but then we will have to write code for migration and that's not ideal either... 🤷🏽

EngHabu (approved):
👍🏽

PR #244, opened 2022-02-16
Refactor/map task k8s pod plugin

hamersaw:
# TL;DR
Refactor of the k8s array plugin to reuse the pod plugin to manage subtask k8s resources reducing code duplication.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
TODO

## Tracking Issue
_NA_

## Follow-up issue
_NA_


go/tasks/plugins/k8s/pod/plugin.go
```diff
@@ -75,7 +78,16 @@ func (p plugin) BuildResource(ctx context.Context, taskCtx pluginsCore.TaskExecu
 	return pod, nil
 }
 
-func (plugin) GetTaskPhase(ctx context.Context, pluginContext k8s.PluginContext, r client.Object) (pluginsCore.PhaseInfo, error) {
+func (p plugin) GetTaskPhase(ctx context.Context, pluginContext k8s.PluginContext, r client.Object) (pluginsCore.PhaseInfo, error) {
+	logPlugin, err := logs.InitializeLogPlugins(logs.GetLogConfig())
```
EngHabu, line 82: We should probably do this once when we create the plugin and just keep a pointer to the logPlugin...

go/tasks/plugins/array/k8s/subtask.go
```diff
@@ -0,0 +1,346 @@
...
+	if err != nil && !k8serrors.IsAlreadyExists(err) {
+		if k8serrors.IsForbidden(err) {
+			if strings.Contains(err.Error(), "exceeded quota") {
+				logger.Warnf(ctx, "Failed to launch job, resource quota exceeded and the operation is not guarded by back-off. err: %v", err)
+				return pluginsCore.PhaseInfoWaitingForResourcesInfo(time.Now(), pluginsCore.DefaultPhaseVersion, fmt.Sprintf("Exceeded resourcequota: %s", err.Error()), nil), nil
+			}
+			return pluginsCore.PhaseInfoRetryableFailure("RuntimeFailure", err.Error(), nil), nil
+		} else if k8serrors.IsBadRequest(err) || k8serrors.IsInvalid(err) {
+			logger.Errorf(ctx, "Badly formatted resource for plugin [%s], err %s", executorName, err)
+			// return pluginsCore.DoTransition(pluginsCore.PhaseInfoFailure("BadTaskFormat", err.Error(), nil)), nil
```
EngHabu, line 201: Was this commented out in the original code as well?

go/tasks/plugins/array/k8s/subtask.go
```diff
@@ -0,0 +1,346 @@
...
+
+		addMetadata(stCtx, cfg, config.GetK8sPluginConfig(), pod)
+		nsName = k8stypes.NamespacedName{Namespace: pod.GetNamespace(), Name: pod.GetName()}
+	}
+
+	// In InjectFinalizer is on, it means we may have added the finalizers when we launched this resource. Attempt to
+	// clear them to allow the object to be deleted/garbage collected. If InjectFinalizer was turned on (through config)
+	// after the resource was created, we will not find any finalizers to clear and the object may have already been
+	// deleted at this point. Therefore, account for these cases and do not consider them errors.
+	if k8sPluginCfg.InjectFinalizer {
```
EngHabu, line 239: I _think_ this issue existed before... if we every enable the InjectFinalizer flag, then launch some workflows, then disable the flag, the plugin won't even attempt to clear the finalizers and these workflows will be forever stuck... am I reading this correctly?
I think we may have to either store in the task plugin state the fact we injected the finalizer or change this code to always read the resource and clear the finalizer IFF our finalizer exist...
Either way, if this is an existing issue, mind filing a github issue about it so we don't forget? if it's a new issue in this code, which way do you think we should go?

go/tasks/plugins/k8s/pod/plugin.go
```diff
@@ -19,13 +20,15 @@ import (
...
 	podTaskType         = "pod"
-	primaryContainerKey = "primary_container_name"
+	PrimaryContainerKey = "primary_container_name"
 )
 
 var (
-	defaultPodBuilder = containerPodBuilder{}
-	podBuilders       = map[string]podBuilder{
-		sidecarTaskType: sidecarPodBuilder{},
+	DefaultPodPlugin = plugin{
```
EngHabu, line 27: Is there a reason this is a global variable not just initialized in the Array Plugin when it's created?

EngHabu (commented):
This is big! and github compare doesn't make it any easier to spot the code that just moved vs the new code... I tried to read through and left some comments... let's make sure we test a bunch in demo env before we merge...

hamersaw:
@EngHabu leaving a comment to make future tracking through this PR easier. Currently there is a significant amount of code duplication between the k8s array plugin, the pod plugin, and the flytepropeller plugin manager.

This PR aims to remove the duplication between the k8s array plugin and the pod plugin by leveraging the pod plugin internally for creating resources and checking the subtask phase. There was very little difference (basically application of k8s array plugin configuration) between the two implementations, just a mess to sort through.

I think there needs to be another round of refactoring (though not urgent) because this code is now structured so that the subtask.go file is basically a 1 to 1 duplicate of the flytepropeller plugin manager. This PR is already large enough, and I did not want to add further complexity of inter-repo refactoring into it. It should be fairly straight forward to address this in the future so that a k8s task is executed identically whether it's part of a map task or just a single instance. Therefore, any changes to this portion (ex. InjectFinalizer comment above) need to be addressed in both locations, or maintained during the aforementioned refactoring.

go/tasks/plugins/array/k8s/subtask.go
hamersaw: I think you are correct. I will test to make sure.

Similar to ^^^ this came straight from the flytepropeller plugin manager [here](https://github.com/flyteorg/flytepropeller/blob/163824c650fbeaedc55a4df99a69825e8ff37306/pkg/controller/nodes/task/k8s/plugin_manager.go#L426). Left a [comment below](https://github.com/flyteorg/flyteplugins/pull/244#issuecomment-1064153003) on one option as to where this fix may be addressed. Happy to fix it anywhere.

EngHabu (dismissed):

EngHabu:
Agreed! Let's resolve conflicts and merge

EngHabu (approved):

PR #247, opened 2022-02-22
Support default PodTemplate for k8s plugin

hamersaw:
# TL;DR
 This PR introduces a new PodTemplateStore struct which is used to track PodTemplates. Therefore, the k8s plugin can use a default PodTemplate (per namespace) and all k8s Pods created by Flyte will be built on top of it, providing a default configuration.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
The PodTemplateStore is a basic mapping of namespace to *PodTemplate instances. We create a global DefaultPodTemplateStore variable with methods to populate by an external system (ex. FlytePropeller informer) and that can be accessed globally (ex. k8s plugin).

When building k8s Pods the k8s plugin attempts to read the PodTemplate for the Pod namespace, if this doesn't exist we fall-back to the default namespace. For example, all Pods created in the "flytesnacks-development" namespace will look for the PodTemplate in that namespace, if it doesn't exist it will use the PodTemplate defined the the default namespace, which is set by default to the namespace of the FlytePropeller pod (in most cases "flyte"). The PodTemplate is used as a base for all images, meaning any parameters set on the PodSpec (ex. provided by the Pod plugin) will override those set in the default PodTemplate. 

Currently, k8s requires PodTemplates to have at least one Container. In this implementation, needlessly adding noop containers seems unnecessary. Therefore, the containers field in any default PodTemplate, while required to be set, is ignored completely when building a Pod. This functionally is open for discussion and will likely be revisited. 

We have chosen to forego supporting configuration hot-updates. The configuration options for this feature are a PodTemplate name and the informer resync duration. Support of hot-updates requires restarting the k8s InformerFactory which populates the PodTemplateStore. Tracking and handling errors through this process is very difficult. 

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2069

## Follow-up issue
_NA_


go/tasks/pluginmachinery/flytek8s/pod_template_store.go
```diff
@@ -0,0 +1,84 @@
...
+)
+
+var DefaultPodTemplateStore PodTemplateStore = NewPodTemplateStore()
+
+// PodTemplateStore maintains a thread-safe mapping of active PodTemplates with their associated
+// namespaces.
+type PodTemplateStore struct {
+	defaultNamespace      string
+	mutex                 sync.Mutex
+	namespacePodTemplates map[string]*v1.PodTemplate
```
EngHabu, line 17: I think using `sync.Map` may offer better (lower) contentions...
https://pkg.go.dev/sync#Map

hamersaw: Much better, didn't know about the thread-safe libs. Going to have to take a deeper look.

EngHabu (approved):

PR #254, opened 2022-03-29
Use monotonically increasing phase version on map tasks

hamersaw:
# TL;DR
Currently the phase version on map task subtasks may not be monotonically increasing in the event of subtask failures. This means that FlyteAdmin will not process events and therefore can miss updates on subtask phase changes, etc. This PR increments the states phase version when subtasks phases are updates and then uses this value to ensure that all events are properly reported and processed by FlyteAdmin.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2288

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/plugins/array/awsbatch/monitor.go
+++ b/go/tasks/plugins/array/awsbatch/monitor.go
@@ -63,6 +63,11 @@ func CheckSubTasksState(ctx context.Context, taskMeta core.TaskExecutionMetadata
 		Detailed: arrayCore.NewPhasesCompactArray(uint(currentState.GetExecutionArraySize())),
 	}
 
+	currentSubTaskPhaseHash, err := currentState.GetArrayStatus().HashCode()
+	if err != nil {
+		return currentState, err
+	}
+
 	queued := 0
 	for childIdx, subJob := range job.SubJobs {
 		actualPhase := subJob.Status.Phase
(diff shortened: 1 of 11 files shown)


go/tasks/plugins/array/awsbatch/monitor.go
```diff
@@ -132,16 +137,20 @@ func CheckSubTasksState(ctx context.Context, taskMeta core.TaskExecutionMetadata
 		errorMsg := msg.Summary(cfg.MaxErrorStringLength)
 		parentState = parentState.SetReason(errorMsg)
 	}
+	_, version := currentState.GetPhase()
 	if phase == arrayCore.PhaseCheckingSubTaskExecutions {
-		newPhaseVersion := uint32(0)
-		// For now, the only changes to PhaseVersion and PreviousSummary occur for running array jobs.
-		for phase, count := range parentState.GetArrayStatus().Summary {
-			newPhaseVersion += uint32(phase) * uint32(count)
+		newSubTaskPhaseHash, err := parentState.GetArrayStatus().HashCode()
```
EngHabu, line 142: Does that take into account the retries? or just the latest attempt?

hamersaw: So if there is a subtask failure the phase for that subtask will change to RetryableFailure within this iteration and thus the newSubTaskPhaseHash != currentSubTaskPhaseHash. On the next iteration we will start the next retry pod (which updates the subtask phase to Queued or Running (forget which) and there will be another phaseVersion increase. This should be the desired behavior because the overall map task phase will be reported as Running and the phaseVersion increments will ensure that flyteAdmin processes them.

I think that this answers your question. We are just checking if the phase changes and for retries we make sure to log a retryable failure before transitioning to running again.

EngHabu: ah I missed the fact that we don't use the hash directly, we just bump the version... I was worried that the hash can circle back and be a) a previously seen value (which will happen but it's ok given the logic here) or b) be less than the current, which is not ok but doesn't matter given you only compare to previous...

EngHabu (approved):

PR #271, opened 2022-06-14
Pass K8sPluginConfig to spark driver and executor pods #patch

fg91:
# TL;DR

Currently, when running Spark tasks, some of the k8s plugin config values configured in the helm values (such as for instance `DefaultTolerations`, `NodeSelector`, `HostNetwork`, `SchedulerName`, ...) are not carried over to the SparkApplication and, thus, not to the driver and executor pods.

In my specific case this is limiting because we run Flyte itself and the Spark operator on cheap nodes while giving workflows the ability to start high-powered nodes via default tolerations.

This PR fixes this issue.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

Comparing the custom [SparkPodSpec](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/9aa84a418803693c13338430bbe662e402861639/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go#L440) with Flyte’s [K8sPluginConfig](https://github.com/fg91/flyteplugins/blob/2c83e24004c09329150b96e24a66f76c4fc660e1/go/tasks/pluginmachinery/flytek8s/config/config.go#L67) shows that the following configurations are already carried over to the `SparkPodSpec` of the driver and the executor:

* `DefaultAnnotations`
* `DefaultLabels`
* `InterruptibleTolerations`
* `DefaultNodeSelector` and `InterruptibleNodeSelector`
* `DefaultPodSecurityContext` ([called` SecurityContext` in `SparkPodSpec`](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/94775cd89ca0158e869fac39e1b097d1bf56a7e8/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go#L497))
* `DefaultPodDNSConfig` ([called `DNSConfig` in `SparkPodSpec`](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/94775cd89ca0158e869fac39e1b097d1bf56a7e8/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go#L519))

This PR adds logic to carry over the following configurations:

* `DefaultTolerations`
* `SchedulerName`
* `DefaultNodeSelector`
* `EnableHostNetworkingPod`
* `DefaultEnvVarsFromEnv`
* `DefaultAffinity`

## Follow-up issue

The Spark operator passes the tolerations from the `SparkApplication` along to the pods only if the operator itself is installed with the `--set webhook.enable=true` value to activate Mutating Admission Webhooks. I feel I should document this somewhere. Should I make a PR to note this [here](https://docs.flyte.org/projects/cookbook/en/stable/auto/integrations/kubernetes/k8s_spark/index.html) or would you recommend another place?

### DIFF
--- a/go/tasks/plugins/k8s/spark/spark.go
+++ b/go/tasks/plugins/k8s/spark/spark.go
@@ -104,6 +104,7 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
 	}
 	driverSpec := sparkOp.DriverSpec{
 		SparkPodSpec: sparkOp.SparkPodSpec{
+			Affinity:         config.GetK8sPluginConfig().DefaultAffinity,
 			Annotations:      annotations,
 			Labels:           labels,
 			EnvVars:          sparkEnvVars,
@@ -120,6 +121,7 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
 
 	executorSpec := sparkOp.ExecutorSpec{
 		SparkPodSpec: sparkOp.SparkPodSpec{
+			Affinity:         config.GetK8sPluginConfig().DefaultAffinity,
 			Annotations:      annotations,
 			Labels:           labels,
 			Image:            &container.Image,
--- a/go/tasks/plugins/k8s/spark/spark.go
+++ b/go/tasks/plugins/k8s/spark/spark.go
@@ -240,7 +240,8 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
 		j.Spec.MainClass = &sparkJob.MainClass
 	}
 
-	// Add Tolerations/NodeSelector to only Executor pods.
+	// Add Interruptible Tolerations/NodeSelector to only Executor pods.
+	// The Interruptible NodeSelector takes precedence over the DefaultNodeSelector
 	if taskCtx.TaskExecutionMetadata().IsInterruptible() {
 		j.Spec.Executor.Tolerations = append(j.Spec.Executor.Tolerations, config.GetK8sPluginConfig().InterruptibleTolerations...)
 		j.Spec.Executor.NodeSelector = config.GetK8sPluginConfig().InterruptibleNodeSelector
--- a/go/tasks/plugins/k8s/spark/spark.go
+++ b/go/tasks/plugins/k8s/spark/spark.go
@@ -105,6 +105,7 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
 			Image:            &container.Image,
 			SecurityContenxt: config.GetK8sPluginConfig().DefaultPodSecurityContext.DeepCopy(),
 			DNSConfig:        config.GetK8sPluginConfig().DefaultPodDNSConfig.DeepCopy(),
+			Tolerations:      config.GetK8sPluginConfig().DefaultTolerations,
 		},
 		ServiceAccount: &serviceAccountName,
 	}
@@ -117,6 +118,7 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
 			EnvVars:          sparkEnvVars,
 			SecurityContenxt: config.GetK8sPluginConfig().DefaultPodSecurityContext.DeepCopy(),
 			DNSConfig:        config.GetK8sPluginConfig().DefaultPodDNSConfig.DeepCopy(),
+			Tolerations:      config.GetK8sPluginConfig().DefaultTolerations,
 		},
 	}
 
@@ -227,9 +229,10 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
 
 	// Add Tolerations/NodeSelector to only Executor pods.
 	if taskCtx.TaskExecutionMetadata().IsInterruptible() {
-		j.Spec.Executor.Tolerations = config.GetK8sPluginConfig().InterruptibleTolerations
+		j.Spec.Executor.Tolerations = append(j.Spec.Executor.Tolerations, config.GetK8sPluginConfig().InterruptibleTolerations...)
 		j.Spec.Executor.NodeSelector = config.GetK8sPluginConfig().InterruptibleNodeSelector
 	}
+
 	return j, nil
 }
 
--- a/go/tasks/plugins/k8s/spark/spark.go
+++ b/go/tasks/plugins/k8s/spark/spark.go
@@ -108,6 +108,7 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
 			Tolerations:      config.GetK8sPluginConfig().DefaultTolerations,
 			SchedulerName:    &config.GetK8sPluginConfig().SchedulerName,
 			NodeSelector:     config.GetK8sPluginConfig().DefaultNodeSelector,
+			HostNetwork:      config.GetK8sPluginConfig().EnableHostNetworkingPod,
 		},
 		ServiceAccount: &serviceAccountName,
 	}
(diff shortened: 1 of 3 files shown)


kumare3:
cc @hamersaw

hamersaw:
Filed an [issue](https://github.com/flyteorg/flyte/issues/2628) for this so we can better track. Hopefully will improve visibility.

hamersaw:
@fg91 I think we discussed scoping this PR down to just applying the existing configuration? We can push upgrading the k8s-on-spark-operator once we have out-of-core plugins implemented (on the roadmap for 1.3 release). Do you have any bandwidth to make these changes? Otherwise we may be able to contribute as well.

fg91:
@hamersaw I removed the commits in which I tried to upgrade K8s and Spark-on-K8s. Then I compared the custom [SparkPodSpec](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/9aa84a418803693c13338430bbe662e402861639/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go#L440) from the currently pinned version with Flyte’s [K8sPluginConfig](https://github.com/fg91/flyteplugins/blob/2c83e24004c09329150b96e24a66f76c4fc660e1/go/tasks/pluginmachinery/flytek8s/config/config.go#L67) again and additionally carried over `EnableHostNetworkingPod`, `DefaultEnvVarsFromEnv`, and `DefaultAffinity`. (I adapted the PR description.)

I'm interested in your opinion how you would treat:

* `DefaultCPURequest`
* `DefaultMemoryRequest`
* `GpuResourceName`
* `ResourceTolerations`

In principle one could map them to `SparkPodSpec.Cores`, `SparkPodSpec.Memory`, ... however I'm not convinced of this because of the following reason:

In a Spark task in flyte one typically configures the resources for the ephemeral spark cluster this way ([source](https://docs.flyte.org/projects/cookbook/en/stable/auto/integrations/kubernetes/k8s_spark/pyspark_pi.html#spark-in-flytekit)):

```python
@task(
    task_config=Spark(
        # this configuration is applied to the spark cluster
        spark_conf={
            "spark.driver.memory": "1000M",
            "spark.driver.cores": "1",
        }
    ),
)
def hello_spark(partitions: int) -> float:
```

Considering `DefaultCPURequest` in addition to `spark_conf` feels like doing the following which according to my understanding doesn't make sense:

```python
@task(
    task_config=Spark(
        # this configuration is applied to the spark cluster
        spark_conf={
            "spark.driver.memory": "1000M",
            "spark.driver.cores": "1",
        }
    ),
    requests=Resources(                  # <-- new
        mem="1G",
    ),
)
def hello_spark(partitions: int) -> float:
```

If you agree, I'd propose to not carry the above mentioned values over and mark this PR ready for review.

go/tasks/plugins/k8s/spark/spark_test.go
```diff
@@ -353,7 +353,47 @@ func TestBuildResourceSpark(t *testing.T) {
...
+		"x/default": "true",
+	}
+	interruptibleNodeSelector := map[string]string{
+		"x/interruptible": "true",
+	}
+
+	defaultPodHostNetwork := true
+
+	defaultEnvVars := make(map[string]string)
+	defaultEnvVars["foo"] = "bar"
```
fg91, line 371: @hamersaw do you know how `defaultEnvVars` are carried over? I added the test but did not carry them over.

fg91:
@hamersaw can you please validate that the behaviour I documented in [this](https://github.com/flyteorg/flyteplugins/pull/271/commits/ebdd8ac121b981d6d7c006d87e0a493dcfa0cdb6) commit is the desired one?

go/tasks/plugins/k8s/spark/spark.go
```diff
@@ -90,6 +90,11 @@ func (sparkResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsCo
 	for _, envVar := range envVars {
 		sparkEnvVars[envVar.Name] = envVar.Value
 	}
+
+	for k, v := range config.GetK8sPluginConfig().DefaultEnvVarsFromEnv {
+		sparkEnvVars[k] = v
+	}
```
hamersaw, line 96: I think the `DefaultEnvVarsFromEnv` are already carried over in the `DecorateEnvVars` [function above](https://github.com/flyteorg/flyteplugins/blob/eed7326340039dbdfcfae75ae6911219ff49371d/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go#L125-L128).

go/tasks/plugins/k8s/spark/spark_test.go
hamersaw: tested locally and yes - the `DecorateEnvVars` bring over the defaultEnvVars.

go/tasks/plugins/k8s/spark/spark.go
fg91: Good catch, thanks! I had misunderstood `DefaultEnvVarsFromEnv`. I removed the logic I had added to carry them over again and corrected the test I had added in [this](https://github.com/flyteorg/flyteplugins/pull/271/commits/63fd1820e310c9dca3aff9b3cc257293fb291773) commit.

fg91:
> It looks like in other k8s resources setting the task to `interruptible`[ appends the interruptible tolerations and node selectors](https://github.com/fg91/flyteplugins/blob/2c83e24004c09329150b96e24a66f76c4fc660e1/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go#L154-L156) rather than overriding them. I think it makes sense to mimic this functionality here, thoughts?

1. `tolerations`: I agree and this is also currently done [here](https://github.com/flyteorg/flyteplugins/blob/d843ed8863263d49a197c95b34aa38effafce17b/go/tasks/plugins/k8s/spark/spark.go#L242).
2. `nodeSelector`: The link you posted doesn't tackle node selectors and I think appending wouldn't work:

    The [docs say](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#nodeselector):

    > Kubernetes only schedules the Pod onto nodes that have each of the labels you specify.

    When trying to schedule this pod
    
    ```yaml
    apiVersion: v1
    kind: Pod
    metadata:
      name: nginx
      labels:
        env: test
    spec:
      containers:
      - name: nginx
        image: nginx
        imagePullPolicy: IfNotPresent
      nodeSelector:
        foo: bar
        foofoo: barbar
    ```

    on a node that has only been labeled `kubectl label nodes k3d-sandbox-server-0 foo=bar`, it remains pending:
    
    ```console
    Warning  FailedScheduling  10s   default-scheduler  0/1 nodes are available: 1 node(s) didn't match Pod's node affinity/selector.
    ```
    
    I think in order to append, one [would have to use affinities](https://kubernetes.io/docs/concepts/scheduling-eviction/assign-pod-node/#affinity-and-anti-affinity):
    
    > The affinity/anti-affinity language is more expressive. nodeSelector only selects nodes with all the specified labels. Affinity/anti-affinity gives you more control over the selection logic.

    Here in [`K8sPluginConfig`](https://github.com/flyteorg/flyteplugins/blob/eed7326340039dbdfcfae75ae6911219ff49371d/go/tasks/pluginmachinery/flytek8s/config/config.go#L108) it says that `InterruptibleNodeSelector` is deprecated anyways and that one should use `InterruptibleNodeSelectorRequirement` instead.

    Neither `InterruptibleNodeSelectorRequirement` nor `NonInterruptibleNodeSelectorRequirement` are currently handled for Spark tasks. I added functionality to carry over the default affinity and I'm now considering whether the correct behaviour would be to add the `InterruptibleNodeSelectorRequirement` or `NonInterruptibleNodeSelectorRequirement` to the `nodeSelectorRequirements` in the default affinity.  This would be basically the same as done [here](https://github.com/flyteorg/flyteplugins/blob/eed7326340039dbdfcfae75ae6911219ff49371d/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L30) with the only disadvantage that `SparkApplication` doesn't use `v1.PodSpec` but a custom `SparkPodSpec` (see [here](https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/94775cd89ca0158e869fac39e1b097d1bf56a7e8/pkg/apis/sparkoperator.k8s.io/v1beta2/types.go#L440)).

    Did I miss something in my interpretation of `nodeSelector` and do you agree that this should be handling of the `NodeSelectorRequirements`?

hamersaw (approved):

PR #275, opened 2022-07-07
Add dask plugin #patch

bstadlbauer:
# TL;DR
This PR adds a backend [`dask`](https://www.dask.org/) plugin using the [`dask-kubernetes`](https://kubernetes.dask.org/en/latest/operator.html) operator to manage the `dask` cluster lifecycle.

## Type
 - [ ] Bug Fix
 - [ ] Feature
 - [x] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
The plugin works similar to the already existing `spark` plugin. It uses the `DaskJob` ([docs](https://kubernetes.dask.org/en/latest/operator_resources.html#daskjob)) Custom Resource as a client pod which connects to the spun up cluster. This is similar to how the `SparkApplication` Custom Resource works for the `spark` plugin. 

TODO before this can be merged: 
- [ ] Use correct `flyteidl` after https://github.com/flyteorg/flyteidl/pull/339 is merged and released

## Tracking Issue

https://github.com/flyteorg/flyte/issues/427

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/plugins/k8s/dask/dask.go
+++ b/go/tasks/plugins/k8s/dask/dask.go
@@ -5,10 +5,9 @@ import (
 	"fmt"
 	"time"
 
-	"k8s.io/apimachinery/pkg/util/intstr"
-	"k8s.io/client-go/kubernetes/scheme"
-
+	daskAPI "github.com/bstadlbauer/dask-k8s-operator-go-client/pkg/apis/kubernetes.dask.org/v1"
 	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/plugins"
 	"github.com/flyteorg/flyteplugins/go/tasks/errors"
 	"github.com/flyteorg/flyteplugins/go/tasks/logs"
 	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery"
@@ -19,11 +18,10 @@ import (
 	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/k8s"
 	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/tasklog"
 	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/utils"
-	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
-
-	daskAPI "github.com/bstadlbauer/dask-k8s-operator-go-client/pkg/apis/kubernetes.dask.org/v1"
-	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/plugins"
 	v1 "k8s.io/api/core/v1"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+	"k8s.io/apimachinery/pkg/util/intstr"
+	"k8s.io/client-go/kubernetes/scheme"
 	"sigs.k8s.io/controller-runtime/pkg/client"
 )
 


go/tasks/plugins/k8s/dask/dask.go
```diff
@@ -0,0 +1,300 @@
...
+	}
+
+	jobPodName := job.ObjectMeta.Name + jobRunnerPodPostfix
+	jobPodNamespace := job.ObjectMeta.Namespace
+	pod, err := clientset.CoreV1().Pods(jobPodNamespace).Get(ctx, jobPodName, metav1.GetOptions{})
+	if err != nil {
+		return nil, err
+	}
+	return pod, nil
+}
```
bstadlbauer, line 281: At the moment, the `DaskJob` CR has no notion of a state. So to get the state, I am currently instantiating a k8s `clientset`, query for the job pod, and use that to determine state, similar to how the `pod` plugin does that.\

My questions here would be: 
1. Is this something that is acceptable? From what I've seen, usually only propeller has a notion of `clientset`, other plugins rely on the state within the CR provided by to the plugin by the propeller. If this is not an option, I can try to contribute to the `dask`-operator, potentially we could save some state in there.
2. If yes, should there be configuration which allows authenticating using a local kubeconfig, or should we only support in-cluster mode?

go/tasks/plugins/k8s/dask/dask.go
```diff
@@ -0,0 +1,300 @@
...
+	if len(info.Logs) > 0 {
+		return pluginsCore.PhaseInfoRunning(pluginsCore.DefaultPhaseVersion+1, &info), nil
+	}
+	return pluginsCore.PhaseInfoRunning(pluginsCore.DefaultPhaseVersion, &info), nil
+}
+
+
+func getClientset() (*kubernetes.Clientset, error) {
+	// FIXME: Handle auth
+	config, err := clientcmd.BuildConfigFromFlags("", "/Users/bstadlbauer/.kube/config")
```
bstadlbauer, line 257: Note that this is only for testing, if we keep it, it would have a similar logic to the one in propeller (also supporting in-cluster auth, etc.)

bstadlbauer (commented):
Hi @hamersaw @eapolinario @pingsutw!

This is a first draft of the `dask` backend plugin. I mostly wanted to share progress with you here and would have one small question around it.
It already runs successfully, but the code is still lacking:
- Configuration options (currently everything is hardcoded) -> A PR for the Python API should come soon in `flytekit`, this would probably be the best way to discuss the interface? 
- Any sort of testing

go/tasks/plugins/k8s/dask/dask.go
kumare3: ideally plugins should not create a new k8s config. This context is better kept within propeller and injected in.

Is it that you want to get the reference to the driver pod? We can do that in a different way

hamersaw:
This is awesome! Thanks for working on it, I see it's been a bit of effort in different repos as well. One question, in doing some reading on the dask operator [here](https://kubernetes.dask.org/en/latest/operator.html) it sounds like once the DaskJob CRD is created the control is handed over the the Job Runner Pod (hope terminology is correct). As much so that "Once the job Pod runs to completion the cluster is removed automatically". This is why you are forced to retrieve the JobPod to gather the status of the job, because, although a bit unusual for CRDs, the status is not tracked in the top level. Does aborting the JobPod shutdown the cluster as well? If so, FlytePropeller uses the BuildIdentityResource function for checking status, aborting, and finalizing the task - maybe we could set this to the JobPod rather than the DaskJob?

bstadlbauer:
Hi @hamersaw!

Sorry for the slow progress on this, but last time I've tried to continue I ran into weird issues with my Flyte dev setup on my private machine (`flytectl` would segfault upon registering). I have found a workaround now and have had a closer look at your proposal.
I've tried switching over the IdentityResource from being the `DaskJob` CRD to being the Job Runner Pod, however, I saw two problems with this approach: 
1. The resource's name is set to the node name in propeller when adding the objects metadata ([here](https://github.com/bstadlbauer/flytepropeller/blob/9cf5368793e5d60113cb819a6e496a7b6accd619/pkg/controller/nodes/task/k8s/plugin_manager.go#L258)). However, the `dask` operator will be appending `"-runner"` to the that name ([here](https://github.com/dask/dask-kubernetes/blob/5718f1f7eaf1f5220dda6cc0dc4e7162a54751a3/dask_kubernetes/operator/operator.py#L100)) when creating the Job Runner Pod (which is not configurable at the moment).
2. When the plugin first creates a DaskJob CRD in `BuildResouce()` but then uses the Job Runner Pod as Identity Resource, we could potentially get into the state where the CRD is created, but the dask operator has not created the pod yet, thus leading to an error in propeller as it won't be able to find the Identity Resource, right? 

I might be missing something here though? Otherwise, I would propose that I try to add state to the CRD on the `dask` side, so that we can then do something similar to the other k8s plugins?

go/tasks/plugins/k8s/dask/dask.go
```diff
@@ -0,0 +1,433 @@
...
+		quantity, err := resource.ParseQuantity(resourceEntry.GetValue())
+		if err != nil {
+			return nil, errors.Errorf(errors.BadTaskSpecification, fmt.Sprintf("Could not parse %s as k8s resource", resourceEntry.GetValue()))
+		}
+		resourceList[resourceName] = quantity
+	}
+	return resourceList, nil
+}
+
+func convertProtobufResourcesToK8sResources(pb_resources *core.Resources) (*v1.ResourceRequirements, error) {
```
hamersaw, line 124: probably a good idea to stick with camel-case per go standards - `pb_resources` -> `pbResources`

hamersaw (commented):
The only big thing I notice is figuring out resources. I don't like that there is parsing (ie. protobuf to k8s) specific code here. It looks like this functionality has been [implemented in FlytePropeller](https://github.com/flyteorg/flytepropeller/blob/01ec2147cc830ec8271f3423e88d82cc23a68609/pkg/utils/k8s.go#L81), which unfortunately cannot be included as a dependency here.

IIUC this uses the task-defined resources as defaults for everything and then allows overriding through the configuration. I think this a sane way of doing it - but should certainly be documented. I think the two options are:
(1) refactor flytepropeller to move the code to flyteplugins
(2) use the k8s version resources as part of the configuration. i think we would have to unmarshal them similar to how we do [podspecs in the pod plugin](https://github.com/flyteorg/flyteplugins/blob/85c512edc5d960f783187fa1c7c490067b608d02/go/tasks/plugins/k8s/pod/sidecar.go#L66-L70). honestly, i do not know everything about what this would look like in flytekit, but presumably we could then use the k8s Resource objects are part of the flytekit definition as well.

I'm not sure what is a better route here. Thoughts?

go/tasks/plugins/k8s/dask/dask.go
bstadlbauer: @hamersaw Thanks for seeing this, the Python developer coming through! I ran `golangcci-lint` now - so should be all fixed 👍

go/tasks/plugins/k8s/dask/dask.go
```diff
@@ -0,0 +1,371 @@
...
+	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/flytek8s/config"
+	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/k8s"
+	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/tasklog"
+	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/utils"
+	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
+
+	daskAPI "github.com/bstadlbauer/dask-k8s-operator-go-client/pkg/apis/kubernetes.dask.org/v1"
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/plugins"
+	v1 "k8s.io/api/core/v1"
+	"sigs.k8s.io/controller-runtime/pkg/client"
```
hamersaw, line 27:
```suggestion
	daskAPI "github.com/bstadlbauer/dask-k8s-operator-go-client/pkg/apis/kubernetes.dask.org/v1"

	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/plugins"
	"github.com/flyteorg/flyteplugins/go/tasks/errors"
	"github.com/flyteorg/flyteplugins/go/tasks/logs"
	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery"
	pluginsCore "github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/core"
	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/core/template"
	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/flytek8s"
	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/flytek8s/config"
	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/k8s"
	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/tasklog"
	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/utils"

	v1 "k8s.io/api/core/v1"
	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
	"k8s.io/apimachinery/pkg/util/intstr"
	"k8s.io/client-go/kubernetes/scheme"

	"sigs.k8s.io/controller-runtime/pkg/client"
```

hamersaw: making our go imports alphabetically ordered is my biggest nit review comment.

hamersaw (changes requested):
Once we get the FlyteIDL `1.4.0` -> `1.3.x` fixed lets go ahead and merge!

bstadlbauer:
@hamersaw I've reordered the imports and changed the flytekit version to 1.3.2. I've also removed the `#minor` from the PR title (in case that's what triggers the release)

hamersaw (approved):

PR #279, opened 2022-07-28
Ray job plugin

pingsutw:
# TL;DR

- [Design Doc](https://docs.google.com/document/d/1-695lxz8a-GFz4cFamGkF1NhspKMkDGIDdkY9EDLMz8/edit#)
- [Slide](https://docs.google.com/presentation/d/1d9qyswQppPUlEpSVoXdXPeAV2oEWlaSKLpvPtCv8QMw/edit#slide=id.g10d1fdf7a50_0_14)

1. Add ray plugin to create RayJob custom resource
2. Upgrade the kubeflow operator since the version of sigs.k8s.io/controller-runtime in Kuberay is incompatible with the one in tf-operator/pytorch-operator

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [x] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2641

## Follow-up issue
image

image

image


pingsutw:
The version of [sigs.k8s.io/controller-runtime](https://github.com/ray-project/kuberay/blob/76f96bf5a05651611f9a8734c34c2a7f23f54b54/ray-operator/go.mod#L21) in Kuberay is different with the one (v0.8.2) in tf-operator/pytorch-operator. 
If we only upgrade controller-runtime, we'll run into some compatibility issues in kubeflow plugins since we use very old version of the Kubeflow operator. Therefore, we have to upgrade the kf-operator as well.

go/tasks/plugins/k8s/ray/ray.go
```diff
@@ -0,0 +1,353 @@
...
+		}
+
+		rayClusterSpec.WorkerGroupSpecs = append(rayClusterSpec.WorkerGroupSpecs, workerNodeSpec)
+	}
+
+	serviceAccountName := flytek8s.GetServiceAccountNameFromTaskExecutionMetadata(taskCtx.TaskExecutionMetadata())
+
+	rayClusterSpec.HeadGroupSpec.Template.Spec.ServiceAccountName = serviceAccountName
+	for index := range rayClusterSpec.WorkerGroupSpecs {
+		rayClusterSpec.WorkerGroupSpecs[index].Template.Spec.ServiceAccountName = serviceAccountName
```
wild-endeavor, line 114: is there ever a use-case where different worker groups have different service accounts? @daikeshi

go/tasks/plugins/k8s/ray/ray.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	if logPlugin == nil {
+		return nil, nil
+	}
+
+	// TODO: Retrieve the name of head pod from rayJob.status, and add it to task logs
+	// RayJob CRD does not include the name of the worker or head pod for now
+
+	// TODO: Add ray Dashboard URI to task logs
```
wild-endeavor, line 315: can we make an issue for this?

wild-endeavor: @daikeshi in addition to the dashboard, do you need additional logs/metrics coming from ray?

wild-endeavor (approved):

go/tasks/plugins/k8s/ray/ray.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	headReplicas := int32(1)
+	headNodeRayStartParams := make(map[string]string)
+	if rayJob.RayCluster.HeadGroupSpec != nil && rayJob.RayCluster.HeadGroupSpec.RayStartParams != nil {
+		headNodeRayStartParams = rayJob.RayCluster.HeadGroupSpec.RayStartParams
+	}
+	headNodeRayStartParams["include-dashboard"] = "true"
+	headNodeRayStartParams["node-ip-address"] = "$MY_POD_IP"
+	headNodeRayStartParams["dashboard-host"] = "0.0.0.0"
```
hamersaw, line 66: This means that "include-dashboard" is always true because it overrides the RayCluster parameters set by the user right? Should this be user-configurable?

go/tasks/plugins/k8s/ray/ray.go
```diff
@@ -0,0 +1,353 @@
...
+	}
+
+	return &rayJobObject, nil
+}
+
+func buildHeadPodTemplate(podSpec *v1.PodSpec) v1.PodTemplateSpec {
+	// Some configs are copy from  https://github.com/ray-project/kuberay/blob/b72e6bdcd9b8c77a9dc6b5da8560910f3a0c3ffd/apiserver/pkg/util/cluster.go#L97
+	// They should always be the same, so we could hard code here.
+	primaryContainer := &v1.Container{Name: "ray-head", Image: podSpec.Containers[0].Image}
+	primaryContainer.Resources = podSpec.Containers[0].Resources
```
hamersaw, line 141: I think we need to be VERY careful here. I'm not sure how containers are created here, but it is worrisome to use indices. Is there a scenario where multiple containers are defined and we update the wrong one?

go/tasks/plugins/k8s/ray/ray.go
```diff
@@ -0,0 +1,353 @@
...
+		},
+		{
+			Name:          "dashboard",
+			ContainerPort: 8265,
+		},
+	}
+
+	podTemplateSpec := v1.PodTemplateSpec{
+		Spec: v1.PodSpec{
+			Containers: []v1.Container{*primaryContainer},
```
hamersaw, line 170: If we're only using the containers here why are we not calling `ToK8sContainer` rather than `ToK8sPodSpec` above? Is there a reason we need to generate the full PodSpec?

go/tasks/plugins/k8s/ray/config.go
```diff
@@ -0,0 +1,41 @@
...
+
+	// TTLSecondsAfterFinished is the TTL to clean up RayCluster.
+	// It's only working when ShutdownAfterJobFinishes set to true.
+	TTLSecondsAfterFinished int32 `json:"ttlSecondsAfterFinished,omitempty"`
+
+	// Kubernetes Service Type, valid values are 'ClusterIP', 'NodePort' and 'LoadBalancer'
+	ServiceType string `json:"serviceType,omitempty"`
+
+	// ClusterSelector is used to select running rayclusters by labels
+	ClusterSelector map[string]string `json:"clusterSelector,omitempty"`
```
hamersaw, line 32: I'm not sure I fully understand what this configuration option does. Is this a FlytePropeller level option? Or is it better set on each individual Ray task?

go/tasks/plugins/k8s/ray/ray.go
```diff
@@ -0,0 +1,353 @@
...
+	switch rayJob.Status.JobStatus {
+	case rayv1alpha1.JobStatusPending:
+		return pluginsCore.PhaseInfoNotReady(time.Now(), pluginsCore.DefaultPhaseVersion, "job is pending"), nil
+	case rayv1alpha1.JobStatusFailed:
+		reason := fmt.Sprintf("Failed to create Ray job: %s", rayJob.Name)
+		return pluginsCore.PhaseInfoFailure(errors.TaskFailedWithError, reason, info), nil
+	case rayv1alpha1.JobStatusSucceeded:
+		return pluginsCore.PhaseInfoSuccess(info), nil
+	}
+	return pluginsCore.PhaseInfoQueued(time.Now(), pluginsCore.DefaultPhaseVersion, "JobCreated"), nil
```
hamersaw, line 337: Is the default case "Running" rather than "Queued"? Does the plugin ever report a "Running" phase?

hamersaw (commented):
I think it looks great! A few comments, otherwise my only concern is applying the default k8s plugin options. It looks like the head and worker templates are built using only a single container from the PodSpec Flyte generates. This means things like tolerations, labels, etc that are defined at FlytePropeller to be set on all k8s resources are never set. I know the Spark plugin is being updated to recognize those options because users expected them to be applied, not sure if that is the case here.

go/tasks/plugins/k8s/ray/ray.go
pingsutw: Here. https://github.com/flyteorg/flyte/issues/2748

go/tasks/plugins/k8s/ray/ray.go
pingsutw: Updated it. yes, it should be user-configurable. If users don't set this value in flytekit, will use the default value in propeller configMap.

go/tasks/plugins/k8s/ray/ray.go
pingsutw: sorry, I didn't notice that we have `ToK8sContainer`. Updated it and used ToK8sContainer instead.

go/tasks/plugins/k8s/ray/ray.go
pingsutw: Used `ToK8sContainer` instead now.

go/tasks/plugins/k8s/ray/ray.go
pingsutw: sorry, I missed it. Added running phase to it

go/tasks/plugins/k8s/ray/config.go
pingsutw: [ClusterSelector](https://github.com/ray-project/kuberay/blob/0c98cf2797e6e0e96fbc9889509f622373f7d89f/ray-operator/apis/ray/v1alpha1/rayjob_types.go#L52-L53) is used by ray-operator. If the ClusterSelector is set, ray operator will submit the ray job to that cluster.
I assume users only have one cluster, so all the tasks will run on that cluster. Thus, I add it to the propeller config.
We could also add `ClusterSelector` to the flytekit config in the future if we have two tasks and these tasks run on different clusters. Then, the value in the flytekit config will override the one in propeller config.

hamersaw (dismissed):

hamersaw (approved):

PR #282, opened 2022-08-19
Fixing cache lookup on large fanout map tasks

hamersaw:
# TL;DR
Map task cache lookups use background processes to perform datacatalog requests. This allows parallel execution for faster retrievals. In the case of large fanout map tasks the means that cache lookups may require multiple propeller rounds to complete. This means that the map task state will stay as "PhaseStart" and the data structures used to track subtask caching and phases are not initialized on the first round. With the current implementation, we attempt to use the cache lookup results to indicate cache status of subtasks on the PhaseStart round which can cause a panic in the scenario where the cache lookup has not yet completed and the subtask data structures have not yet been initialized. In this PR we wait to construct external resources (to notify the UI of subtask cache status and phase) until the map task cache lookup completes and we transition out of "PhaseStart".

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/2786
fixes https://github.com/flyteorg/flyte/issues/2739

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/plugins/array/awsbatch/executor.go
+++ b/go/tasks/plugins/array/awsbatch/executor.go
@@ -108,8 +108,9 @@ func (e Executor) Handle(ctx context.Context, tCtx core.TaskExecutionContext) (c
 	// Always attempt to augment phase with task logs.
 	var logLinks []*idlCore.TaskLog
 	var externalResources []*core.ExternalResource
-	switch p {
-	case arrayCore.PhaseStart:
+
+	np, _ := pluginState.GetPhase()
+	if p == arrayCore.PhaseStart && np != arrayCore.PhaseStart {
 		externalResources, err = arrayCore.InitializeExternalResources(ctx, tCtx, pluginState.State,
 			func(tCtx core.TaskExecutionContext, childIndex int) string {
 				// subTaskIDs for the the aws_batch are generated based on the job ID, therefore
@@ -117,7 +118,7 @@ func (e Executor) Handle(ctx context.Context, tCtx core.TaskExecutionContext) (c
 				return ""
 			},
 		)
-	default:
+	} else {
 		logLinks, externalResources, err = GetTaskLinks(ctx, tCtx.TaskExecutionMetadata(), e.jobStore, pluginState)
 	}
 
--- a/go/tasks/plugins/array/awsbatch/executor.go
+++ b/go/tasks/plugins/array/awsbatch/executor.go
@@ -68,7 +68,7 @@ func (e Executor) Handle(ctx context.Context, tCtx core.TaskExecutionContext) (c
 
 	switch p {
 	case arrayCore.PhaseStart:
-		pluginState.State, err = array.DetermineDiscoverability(ctx, tCtx, pluginState.State)
+		pluginState.State, err = array.DetermineDiscoverability(ctx, tCtx, pluginConfig.MaxArrayJobSize, pluginState.State)
 
 	case arrayCore.PhasePreLaunch:
 		pluginState, err = EnsureJobDefinition(ctx, tCtx, pluginConfig, e.jobStore.Client, e.jobDefinitionCache, pluginState)
--- a/go/tasks/plugins/array/awsbatch/executor.go
+++ b/go/tasks/plugins/array/awsbatch/executor.go
@@ -109,8 +109,7 @@ func (e Executor) Handle(ctx context.Context, tCtx core.TaskExecutionContext) (c
 	var logLinks []*idlCore.TaskLog
 	var externalResources []*core.ExternalResource
 
-	np, _ := pluginState.GetPhase()
-	if p == arrayCore.PhaseStart && np != arrayCore.PhaseStart {
+	if p == arrayCore.PhasePreLaunch {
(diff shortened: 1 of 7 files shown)


go/tasks/plugins/array/catalog.go
```diff
@@ -121,6 +121,13 @@ func DetermineDiscoverability(ctx context.Context, tCtx core.TaskExecutionContex
 	}
 
 	// Otherwise, run the data catalog steps - create and submit work items to the catalog processor,
+	cfg := catalog.GetConfig()
+	if int(arrayJobSize) > cfg.WriterWorkqueueConfig.IndexCacheMaxItems || int(arrayJobSize) > cfg.ReaderWorkqueueConfig.IndexCacheMaxItems {
```
EngHabu, line 125: Do we not have a check on the array job size at an earlier point? Is it not sufficient? Should we update it to be so we fail as early as possible

go/tasks/plugins/array/awsbatch/executor.go
```diff
@@ -108,16 +108,17 @@ func (e Executor) Handle(ctx context.Context, tCtx core.TaskExecutionContext) (c
 	// Always attempt to augment phase with task logs.
 	var logLinks []*idlCore.TaskLog
 	var externalResources []*core.ExternalResource
-	switch p {
-	case arrayCore.PhaseStart:
+
+	np, _ := pluginState.GetPhase()
+	if p == arrayCore.PhaseStart && np != arrayCore.PhaseStart {
```
CalvinLeather, line 113: This is added logic to hande the new scenario where we stay in PhaseStart for multiple cycles right? i.e., this only runs InitializeExternalResources if something above moves us out of PhaseStart?

go/tasks/plugins/array/k8s/executor.go
```diff
@@ -91,12 +91,19 @@ func (e Executor) Handle(ctx context.Context, tCtx core.TaskExecutionContext) (c
...
-				subTaskExecutionID := NewSubTaskExecutionID(tCtx.TaskExecutionMetadata().GetTaskExecutionID(), childIndex, 0)
-				return subTaskExecutionID.GetGeneratedName()
-			},
-		)
+		// if the next state has not yet transitioned from PhaseStart it means that the background
+		// cache lookup for map task subtasks has not yet been completed. this is necessary for
+		// the InitializeExternalResources operation, which is used to notify admin of subtask
+		// including cache status. a transition from PhaseStart indicates cache lookup completion
+		// and we may InitializeExternalResources.
+		if np, _ := nextState.GetPhase(); np != arrayCore.PhaseStart {
```
CalvinLeather, line 99: In go/tasks/plugins/array/awsbatch/executor.go, you checked explicitly for the transition from PhaseStart to "something that isn't phaseStart"... I"ll look around a bit more, but is "the previous state was PhaseStart" guaranteed  by virtue of execution reaching this line? Otherwise, is there a situation where this gets called, but the previous state wasn't Phase Start? Does that cause problems?

CalvinLeather: Okay cool, the switch statement ensures the we were previously in PhaseStart, so the logic here replicates that in go/tasks/plugins/array/awsbatch/executor.go (i.e., ensuring a state transition from PhaseStart to something that isn't PhseStart)

hamersaw: Correct.

go/tasks/plugins/array/awsbatch/executor.go
hamersaw: Correct, the progression of map task plugin state will always start with PhaseStart and transition to PhasePreLaunch / PhaseLaunch to progress. The PhaseStart status continues until the cache lookup is completed (performed in the background). So we basically wait to initialize subtask status' (with admin to track in the UI) until the cache lookups are completed. We need the cache lookups to correctly report which subtasks are read from cache and which are executing.

go/tasks/plugins/array/catalog.go
hamersaw: Great point. The [current check ](https://github.com/flyteorg/flyteplugins/blob/cdfc191012c2b2d5e3d956dec031df790b75386e/go/tasks/plugins/array/k8s/management.go#L69-L74) is performed to check if the execution array size (subtasks that are not cache hits) is larger than the max configured array job size. Two things (1) the check added in this PR does actually happen before the existing check (perhaps we should move the existing to perform at the same time as this one?) and (2) the checks have some difference, the existing check is on the number of executed subtasks and this new check pertains only the the cached items. In diving into the code we have a separate configuration for the size of the LRU cache for performing subtask cache lookups. The cache lookup only completes once all of the items are completed (available in the LRU cache). So intuitively, if we have an array size larger than the size of the LRU cache we will be stuck in an infinite loop.

I can move the existing check if that makes more sense? I think it probably does.

EngHabu (approved):

EngHabu (approved):

PR #285, opened 2022-09-27
Added overwrite parameter to catalog client.Put

MorpheusXAUT:
# TL;DR
The catalog client's `Put` now has an additional parameter `overwrite`, indicating whether the artifact should be created (`false`) or updated (`true`).

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [X] Smoke tested
 - [ ] Unit tests added

## Complete description
The catalog client interface's `Put` was previously only used to _create_ artifacts. In order to support the cache eviction RFC (and its first step of overwriting existing results for a single execution), an `overwrite` parameter was added, indicating the data store should replace an existing artifact instead of creating a new one.
As the current implementation is already called `Put` and the HTTP verb `PUT` is expected to replace existing data as well, I've opted for extending the current signature instead of creating a separate new method.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2867

## Follow-up issue
_NA_


hamersaw (approved):

go/tasks/pluginmachinery/catalog/client.go
```diff
@@ -128,7 +128,7 @@ func NewReservationEntry(expiresAt time.Time, heartbeatInterval time.Duration, o
 type Client interface {
 	Get(ctx context.Context, key Key) (Entry, error)
 	GetOrExtendReservation(ctx context.Context, key Key, ownerID string, heartbeatInterval time.Duration) (*datacatalog.Reservation, error)
-	Put(ctx context.Context, key Key, reader io.OutputReader, metadata Metadata) (Status, error)
+	Put(ctx context.Context, key Key, reader io.OutputReader, metadata Metadata, overwrite bool) (Status, error)
```
hamersaw, line 131: Do you think it makes sense to change this to have separate `Put` and `Update` functions rather than adding the overwrite flag? I think it would closer mimic the service API, then we could move the handling of the `overwrite_cache` flag from the datacatalog client to the [pre_post_execution.go file](https://github.com/flyteorg/flytepropeller/pull/489/files#diff-0693141f0fe3dad54187bd2eb73f89fba7e5abc2bac038db3e5ff0fc00cd4cfe). It would be exactly like what you had in the client, ex:
```
if overwrite_cache {
    catalogClient.Update()
    // check if err is AlreadyExists and fallthrough to Put
}

catalogClient.Put()
```

MorpheusXAUT: I was thinking about that as well, but as someone with roots in web development, I would expect a function called `Put` to handle updates as well (instead of just creating resources), so the (existing) naming is not... ideal. Since I didn't want to rename the existing function, a flag to toggle functionality seemed less breaking.

Can change it to `Put` and `Update` including the handling you've mentioned if you'd prefer that, I'm not 100% particular for either way.

hamersaw: @MorpheusXAUT makes plenty of sense. IMO opinion either approach is fine, but if we have a `Put` that upserts I think that functionality should be implemented in the backend. Is that how you would expect it in a web service application?. I don't think implementing the `Update` with `Put` fallback in the client is the right way to do it. I'm leaning towards a separate `Update` function, but if your vehemently opposed I could be talked out of it :).

MorpheusXAUT: @hamersaw Yeah, I would expect the webserver to handle the `PUT` accordingly (as the client wouldn't know whether a resource exists in many cases) - that is, if the server is actually called with a `PUT` (instead of `POST` for creating or `PATCH` for partial updates).

Unfortunately the client interface (in flyteplugins here) and server implementation (in datacatalog) diverge a bit since the client defines a `Put` method (which _could_ create or update), however the server explicitly uses `Create` (and now `Update`), so there's no idempotent operation like the HTTP verb `PUT` available on the server side.
To keep the existing APIs as close to their current implementation as possible, I opted to "override" the existing client `Put` and offloading the decision whether to `Create` or `Update` on the server side to that as I believe that's the least breaking/ambiguous change.

With the current implementation in propeller, splitting it into `Put` and `Update` would not be a problem though since the code is mostly encapsulated in separate functions anyways, so I can relatively easily change that if you'd prefer. I'm not 100% happy about the `Put` naming for "create-only", however I'm not keen on potentially breaking other implementations by renaming it - it's not a hill to die on for me, so I'm fine with changing it as well :grin:

MorpheusXAUT: As discussed briefly on Slack, I've split this into `Put` (remaining basically untouched) and `Update` (providing the new functionality).
The decision on what catalog client method to call is performed in `pre_post_execution.go` now.

hamersaw (approved):

PR #288, opened 2022-10-05
Added -unknown suffix to container name

jcarey03:
Signed-off-by: Jason Carey <jcarey@lyft.com>

# TL;DR
Adding -unknown suffix to container name for Lyft logging agent to detect and send logs to logging pipeline.

## Type
 - [ ] Bug Fix
 - [X] Feature
 - [ ] Plugin

### DIFF
--- a/go/tasks/pluginmachinery/flytek8s/container_helper.go
+++ b/go/tasks/pluginmachinery/flytek8s/container_helper.go
@@ -117,7 +117,7 @@ func ToK8sContainer(ctx context.Context, taskContainer *core.Container, iFace *c
 	}
 	// Make the container name the same as the pod name, unless it violates K8s naming conventions
 	// Container names are subject to the DNS-1123 standard
-	containerName := parameters.TaskExecMetadata.GetTaskExecutionID().GetGeneratedName()
+	containerName := parameters.TaskExecMetadata.GetTaskExecutionID().GetGeneratedName() + "-unknown"
 	if errs := validation.IsDNS1123Label(containerName); len(errs) > 0 {
 		containerName = rand.String(4)
 	}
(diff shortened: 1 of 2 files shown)


hamersaw:
@jcarey03 can you explain a little bit more about the use-case here? I'm a little concerned about modifying the container name on all deployments. How is the "-unknown" detected by the agent? Is there a different field that can be set to identify these containers?

jcarey03:
@hamersaw This PR is only scoped for the `flyteorg:lyft` branch as it pertains to Lyft-specific infrastructure; it will not be pulled into the `master` branch. Our logging infrastructure uses fluentbit as a log collector and forwarder and is configured to monitor containers with a specific suffix. We decided to make the basic change first and test integration. Once that is confirmed, we'll circle back and improve upon the naive hardcoded solution to something more configurable, but again, this is only for Lyft and not general Flyte.

hamersaw:
> @hamersaw This PR is only scoped for the `flyteorg:lyft` branch as it pertains to Lyft-specific infrastructure; it will not be pulled into the `master` branch. With that said, I'm pretty sure there is a much better way to make this feature configurable if that is a desirable feature for the Flyte community.

Oh sorry! did not catch this PR was into the Lyft branch.

> Our logging infrastructure uses fluentbit as a log collector and forwarder and is configured to monitor containers with a specific suffix. We decided to make the basic change first and test integration. Once that is confirmed, we'll circle back and improve upon the naive hardcoded solution to something more configurable, but again, the thinking is that this is only for Lyft and not general Flyte.

So, for what my two cents it worth here, there was recently a community contribution to enable default configuration of containers using the [default PodTemplate](https://docs.flyte.org/en/latest/deployment/cluster_config/general.html#using-default-k8s-podtemplates) work. I know their use-case was for logging in the primary container, but I can't recall what fields they were using to identify the container or link up the logging framework. Not sure if this is worth integrating into for your setup, but may be worth a quick look!

jcarey03:
Thanks for sharing. We'll take a look.

PR #290, opened 2022-10-13
added 'python-task' to registered task types for pod plugin

hamersaw:
# TL;DR
Python tasks are annotated with "python-task" for the task type. In FlytePlugins we do not have a Handler that registers with that task type, so in every case it falls back to the container handler which is registered as the default for everything that does not match an existing handler. However, this results in log messages like `No plugin found for Handler-type [python-task], defaulting to [container]` everytime that propeller evaluates a "python-task" node. This can result in a very large number of unuseful log messages, effectively reducing the utility of logging. Adding "python-task" as a registered type will remove this message from the logs.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
^^^

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/plugins/k8s/pod/container.go
+++ b/go/tasks/plugins/k8s/pod/container.go
@@ -14,6 +14,7 @@ import (
 
 const (
 	ContainerTaskType = "container"
+	PythonTaskType    = "python-task"
 )
 
 type containerPodBuilder struct {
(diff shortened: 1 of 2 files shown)


hamersaw:
cc @nicklofaso

nicklofaso:
Very helpful! Thanks @hamersaw

pingsutw (approved):

PR #291, opened 2022-10-17
The status of the AWS batch job should become failed once the retry limit exceeded

pingsutw:
# TL;DR
The task status is always running if the task failure is retryable. In other words, the propeller will keep waiting no matter how often the task is re-run. Take a look at these [code](https://github.com/flyteorg/flyteplugins/blob/eed7326340039dbdfcfae75ae6911219ff49371d/go/tasks/plugins/array/core/state.go#L271-L275). If all the sub-task failed and all the failure is retryable, the state of the task will never become `PhaseWriteToDiscoveryThenFail`.  (totalRetryableFailures is equal to minSuccesses here)

In this pr, we kept tracking how often the task has retried and counted the total number of sub-tasks that exceeded the retry limit.  Stop the task when `totalRunning+totalSuccesses+totalWaitingForResources+totalRetryableFailures-totalRetryLimitExceeded < minSuccesses` is true

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
image

## Tracking Issue
https://github.com/flyteorg/flyte/issues/2979

## Follow-up issue
_NA_


go/tasks/plugins/array/awsbatch/monitor.go
```diff
@@ -108,6 +109,10 @@ func CheckSubTasksState(ctx context.Context, taskMeta core.TaskExecutionMetadata
 			} else {
 				msg.Collect(childIdx, "Job failed")
 			}
+
+			if subJob.Status.Phase == core.PhaseRetryableFailure && retryLimit == int64(len(subJob.Attempts)) {
+				totalRetryLimitExceeded++
+			}
```
hamersaw, line 115: Does it make sense to transition the subtask to PermanentFailure here (like we do in the [array plugin](https://github.com/flyteorg/flyteplugins/blob/d95b86b4ba9e6fcab18b3ea834047c4248cf314a/go/tasks/plugins/array/k8s/management.go#L251-L255)) rather than tracking the number of retryable failures which exceeded the retry limit?

pingsutw: updated it

hamersaw (approved):

PR #293, opened 2022-10-23
#patch Cloudwatch FluentD

Azanul:
Signed-off-by: Azanul Haque <42029519+Azanul@users.noreply.github.com>

# TL;DR
This PR fixes Log Links that do not work with CloudWatch FluentD out of the box

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [X] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
Replaced the linking cloudwatch url template with correct one.

## Tracking Issue
fixes https://github.com/flyteorg/flyte/issues/2635

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/logs/logging_utils.go
+++ b/go/tasks/logs/logging_utils.go
@@ -97,7 +97,7 @@ func InitializeLogPlugins(cfg *LogConfig) (tasklog.Plugin, error) {
 			logPlugins = append(logPlugins, logPlugin{Name: "Cloudwatch Logs", Plugin: tasklog.NewTemplateLogPlugin([]string{cfg.CloudwatchTemplateURI}, core.TaskLog_JSON)})
 		} else {
 			logPlugins = append(logPlugins, logPlugin{Name: "Cloudwatch Logs", Plugin: tasklog.NewTemplateLogPlugin(
-				[]string{fmt.Sprintf("https://console.aws.amazon.com/cloudwatch/home?region=%s#logEventViewer:group=%s;stream=var.log.containers.{{ .podName }}_{{ .namespace }}_{{ .containerName }}-{{ .containerId }}.log", cfg.CloudwatchRegion, cfg.CloudwatchLogGroup)}, core.TaskLog_JSON)})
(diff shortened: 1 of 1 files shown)


samhita-alla:
@EngHabu, is this the required change, or is there something else that needs to be done?

EngHabu:
Thank you for your help! I went through this exercise recently and what I see is missing at the moment is the hostname. By default, these logstreams include the name of the host they were connected from. There are no template values (e.g. {{.host}}) to use. The fix will involve adding that template like the other ones, populate it from the Pod spec and then use it in the log link.

Azanul:
@EngHabu By host do you mean nodeName ? are you referring to:

> Fluent Bit optimized configuration sends logs to kubernetes-nodeName-application.var.log.containers.kubernetes-podName_kubernetes-namespace_kubernetes-container-name-kubernetes-containerID

EngHabu:
That's correct, the hostname is not provided... that's the work needed

andrusha:
From what I see the pods which Flyte create have local hostname, but what container insights log path expects is the node name, which is not available on the Flyte pod. I imagine that there should be another template parameter which will include node name or --hostname-override could be passed to the Flyte pods on creation to set it to the node hostname instead.

EngHabu:
There seems to be a `pod.Spec.NodeName` field: https://kubernetes.io/docs/reference/kubernetes-api/workload-resources/pod-v1/#scheduling
Is that not enough to populate the URL?
And yes, we will need an additional template parameter that resolves to this value.

ashrielbrian:
@EngHabu @andrusha is there someone working on this? we use fluent-bit for logging as well, and this is breaking our cloudwatch log link.

andrewwdye:
I used [this](https://github.com/aws/eks-charts/blob/master/stable/aws-for-fluent-bit/values.yaml#L114) aws-for-fluent-bit chart recently which uses a default `fluentbit-` log stream prefix and doesn't contain hostname. The change to use `logSteamNameFilter` in this PR may be the safest bet.

andrusha:
@ashrielbrian  I don't plan working on this myself, locally we use Datadog, so the issue was solved for us by adding custom logging URL towards it, which didn't require a hostname

PR #297, opened 2022-11-24
Apply default pod template to PytorchJob pods

fg91:
# TL;DR
Currently, the default pod template is applied to Python tasks but not to the kf-operators tasks like the Pytorch task.

This is a problem since e.g. for the pytorch dataloaders, one often has to increase the shared memory, wich on K8s is done by mounting an [`emptyDir` volume](https://www.sobyte.net/post/2022-04/k8s-pod-shared-memory/#kubernetes-empty-dir). Currently there exists no mechanism to do that for Pytorch tasks.

In this PR, the `v1.PodSpec` of the default pod template is applied to the `PodSpec` of the `PytorchJob`.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Tracking Issue

## Follow-up issue
_NA_
OR

### DIFF
--- a/go/tasks/pluginmachinery/flytek8s/pod_helper.go
+++ b/go/tasks/pluginmachinery/flytek8s/pod_helper.go
@@ -2,6 +2,7 @@ package flytek8s
 
 import (
 	"context"
+	"errors"
 	"fmt"
 	"strings"
 	"time"
@@ -145,6 +146,10 @@ func ToK8sPodSpec(ctx context.Context, tCtx pluginsCore.TaskExecutionContext) (*
 }
 
 func MergePodSpecs(podTemplatePodSpec *v1.PodSpec, podSpec *v1.PodSpec, primaryContainerName string) (*v1.PodSpec, error) {
+	if podTemplatePodSpec == nil || podSpec == nil {
+		return nil, errors.New("podTemplatePodSpec and podSpec cannot be nil.")
+	}
+
 	err := mergo.Merge(podTemplatePodSpec, podSpec, mergo.WithOverride, mergo.WithAppendSlice)
 	if err != nil {
 		return nil, err
--- a/go/tasks/pluginmachinery/flytek8s/pod_helper.go
+++ b/go/tasks/pluginmachinery/flytek8s/pod_helper.go
@@ -150,7 +150,10 @@ func MergePodSpecs(podTemplatePodSpec *v1.PodSpec, podSpec *v1.PodSpec, primaryC
 		return nil, errors.New("podTemplatePodSpec and podSpec cannot be nil.")
 	}
 
-	err := mergo.Merge(podTemplatePodSpec, podSpec, mergo.WithOverride, mergo.WithAppendSlice)
+	var podTemplatePodSpecCopy *v1.PodSpec
+	podTemplatePodSpecCopy = podTemplatePodSpec.DeepCopy()
+
+	err := mergo.Merge(podTemplatePodSpecCopy, podSpec, mergo.WithOverride, mergo.WithAppendSlice)
 	if err != nil {
 		return nil, err
 	}
@@ -158,11 +161,11 @@ func MergePodSpecs(podTemplatePodSpec *v1.PodSpec, podSpec *v1.PodSpec, primaryC
 	// merge template Containers
 	var mergedContainers []v1.Container
 	var defaultContainerTemplate, primaryContainerTemplate *v1.Container
-	for i := 0; i < len(podTemplatePodSpec.Containers); i++ {
-		if podTemplatePodSpec.Containers[i].Name == defaultContainerTemplateName {
-			defaultContainerTemplate = &podTemplatePodSpec.Containers[i]
-		} else if podTemplatePodSpec.Containers[i].Name == primaryContainerTemplateName {
-			primaryContainerTemplate = &podTemplatePodSpec.Containers[i]
+	for i := 0; i < len(podTemplatePodSpecCopy.Containers); i++ {
+		if podTemplatePodSpecCopy.Containers[i].Name == defaultContainerTemplateName {
+			defaultContainerTemplate = &podTemplatePodSpecCopy.Containers[i]
+		} else if podTemplatePodSpecCopy.Containers[i].Name == primaryContainerTemplateName {
+			primaryContainerTemplate = &podTemplatePodSpecCopy.Containers[i]
 		}
 	}
 
@@ -201,9 +204,9 @@ func MergePodSpecs(podTemplatePodSpec *v1.PodSpec, podSpec *v1.PodSpec, primaryC
 	}
 
 	// update Pod fields
-	podTemplatePodSpec.Containers = mergedContainers
+	podTemplatePodSpecCopy.Containers = mergedContainers
 
-	return podTemplatePodSpec, nil
+	return podTemplatePodSpecCopy, nil
 }
 
 func BuildPodWithSpec(podTemplate *v1.PodTemplate, podSpec *v1.PodSpec, primaryContainerName string) (*v1.Pod, error) {
@@ -216,9 +219,7 @@ func BuildPodWithSpec(podTemplate *v1.PodTemplate, podSpec *v1.PodSpec, primaryC
 
 	if podTemplate != nil {
 		// merge template PodSpec
-		basePodSpec := podTemplate.Template.Spec.DeepCopy()
-
-		mergedPodSpec, err := MergePodSpecs(basePodSpec, podSpec, primaryContainerName, defaultContainerTemplateName)
+		mergedPodSpec, err := MergePodSpecs(&podTemplate.Template.Spec, podSpec, primaryContainerName, defaultContainerTemplateName)
 		if err != nil {
 			return nil, err
 		}
--- a/go/tasks/pluginmachinery/flytek8s/pod_helper.go
+++ b/go/tasks/pluginmachinery/flytek8s/pod_helper.go
@@ -144,10 +144,10 @@ func ToK8sPodSpec(ctx context.Context, tCtx pluginsCore.TaskExecutionContext) (*
 	return pod, nil
 }
 
-func MergePodSpecs(podTemplatePodSpec *v1.PodSpec, podSpec *v1.PodSpec, primaryContainerName string) error {
+func MergePodSpecs(podTemplatePodSpec *v1.PodSpec, podSpec *v1.PodSpec, primaryContainerName string) (*v1.PodSpec, error) {
 	err := mergo.Merge(podTemplatePodSpec, podSpec, mergo.WithOverride, mergo.WithAppendSlice)
 	if err != nil {
-		return err
+		return nil, err
 	}
 
 	// merge template Containers
@@ -175,7 +175,7 @@ func MergePodSpecs(podTemplatePodSpec *v1.PodSpec, podSpec *v1.PodSpec, primaryC
 			} else {
 				err := mergo.Merge(mergedContainer, primaryContainerTemplate, mergo.WithOverride, mergo.WithAppendSlice)
 				if err != nil {
-					return err
+					return nil, err
 				}
 			}
 		}
@@ -187,7 +187,7 @@ func MergePodSpecs(podTemplatePodSpec *v1.PodSpec, podSpec *v1.PodSpec, primaryC
 		} else {
 			err := mergo.Merge(mergedContainer, container, mergo.WithOverride, mergo.WithAppendSlice)
 			if err != nil {
-				return err
+				return nil, err
 			}
 
 			mergedContainers = append(mergedContainers, *mergedContainer)
@@ -197,7 +197,8 @@ func MergePodSpecs(podTemplatePodSpec *v1.PodSpec, podSpec *v1.PodSpec, primaryC
 
 	// update Pod fields
 	podTemplatePodSpec.Containers = mergedContainers
-	return nil
+
+	return podTemplatePodSpec, nil
 }
 
 func BuildPodWithSpec(podTemplate *v1.PodTemplate, podSpec *v1.PodSpec, primaryContainerName string) (*v1.Pod, error) {
@@ -212,13 +213,13 @@ func BuildPodWithSpec(podTemplate *v1.PodTemplate, podSpec *v1.PodSpec, primaryC
 		// merge template PodSpec
 		basePodSpec := podTemplate.Template.Spec.DeepCopy()
 
-		err := MergePodSpecs(basePodSpec, podSpec, primaryContainerName)
+		mergedPodSpec, err := MergePodSpecs(basePodSpec, podSpec, primaryContainerName)
 		if err != nil {
 			return nil, err
 		}
 
 		pod.ObjectMeta = podTemplate.Template.ObjectMeta
-		pod.Spec = *basePodSpec
+		pod.Spec = *mergedPodSpec
 
 	} else {
 		pod.Spec = *podSpec
--- a/go/tasks/pluginmachinery/flytek8s/pod_helper.go
+++ b/go/tasks/pluginmachinery/flytek8s/pod_helper.go
@@ -145,7 +145,7 @@ func ToK8sPodSpec(ctx context.Context, tCtx pluginsCore.TaskExecutionContext) (*
 	return pod, nil
 }
 
-func MergePodSpecs(podTemplatePodSpec *v1.PodSpec, podSpec *v1.PodSpec, primaryContainerName string, defaultContainerName string) (*v1.PodSpec, error) {
+func MergePodSpecs(podTemplatePodSpec *v1.PodSpec, podSpec *v1.PodSpec, primaryContainerName string) (*v1.PodSpec, error) {
 	if podTemplatePodSpec == nil || podSpec == nil {
 		return nil, errors.New("podTemplatePodSpec and podSpec cannot be nil")
 	}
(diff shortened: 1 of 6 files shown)


fg91:
@hamersaw How should I test this?

Ideally I would want to duplicate [this test](https://github.com/flyteorg/flyteplugins/blob/2ec7da85527cebccfff8f16e02895b06e9aa7a8a/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch_test.go#L285) but apply a pod template. However, for this I would have to mock/patch `flytek8s.DefaultPodTemplateStore.LoadOrDefault` which is called [here](https://github.com/flyteorg/flyteplugins/blob/2ec7da85527cebccfff8f16e02895b06e9aa7a8a/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go#L69).

[This](https://github.com/flyteorg/flyteplugins/blob/2ec7da85527cebccfff8f16e02895b06e9aa7a8a/go/tasks/plugins/k8s/pod/plugin.go#L51) function which is the equivalent for normal pods is not tested in the case of an existing pod template either from what I can see.

Methods for mocking like mentioned [here](https://stackoverflow.com/a/19168875) wouldn't work without refactoring and unfortunately I don't understand how you employ [mockery](github.com/EngHabu/mockery/cmd/mockery). Do you see a way to mock `flytek8s.DefaultPodTemplateStore.LoadOrDefault` with `mockery` or would you suggest to only add a unit test for the new function `MergePodSpecs`?

hamersaw:
All great questions. TL;DR I think testing the `MergePodSpecs` function is enough here.

> Ideally I would want to duplicate [this test](https://github.com/flyteorg/flyteplugins/blob/2ec7da85527cebccfff8f16e02895b06e9aa7a8a/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch_test.go#L285) but apply a pod template. However, for this I would have to mock/patch `flytek8s.DefaultPodTemplateStore.LoadOrDefault` which is called [here](https://github.com/flyteorg/flyteplugins/blob/2ec7da85527cebccfff8f16e02895b06e9aa7a8a/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go#L69).

You could manually inject a `PodTemplate` into the `DefaultPodTemplateStore` using the `Store` function. To get an idea of what this looks like you can [see here](https://github.com/flyteorg/flyteplugins/blob/85c512edc5d960f783187fa1c7c490067b608d02/go/tasks/pluginmachinery/flytek8s/pod_template_store.go#L50). I would be careful about how this might effect other tests.

> Methods for mocking like mentioned [here](https://stackoverflow.com/a/19168875) wouldn't work without refactoring and unfortunately I don't understand how you employ [mockery](github.com/EngHabu/mockery/cmd/mockery). Do you see a way to mock `flytek8s.DefaultPodTemplateStore.LoadOrDefault` with `mockery` or would you suggest to only add a unit test for the new function `MergePodSpecs`?

So mockery works by generating stubs for golang interfaces. Currently the `PodTemplateStore` is just [defined as a struct](https://github.com/flyteorg/flyteplugins/blob/85c512edc5d960f783187fa1c7c490067b608d02/go/tasks/pluginmachinery/flytek8s/pod_template_store.go#L14). We could easily add an interface abstraction and use mockery to generate mocks. I'm not sure that this is necessary as long as we are sure the merging logic is sound.

go/tasks/pluginmachinery/flytek8s/pod_helper.go
```diff
@@ -144,70 +144,83 @@ func ToK8sPodSpec(ctx context.Context, tCtx pluginsCore.TaskExecutionContext) (*
 	return pod, nil
 }
 
-func BuildPodWithSpec(podTemplate *v1.PodTemplate, podSpec *v1.PodSpec, primaryContainerName string) (*v1.Pod, error) {
-	pod := v1.Pod{
-		TypeMeta: v12.TypeMeta{
-			Kind:       PodKind,
-			APIVersion: v1.SchemeGroupVersion.String(),
-		},
+func MergePodSpecs(podTemplatePodSpec *v1.PodSpec, podSpec *v1.PodSpec, primaryContainerName string) (*v1.PodSpec, error) {
```
hamersaw, line 147: Do we need to add any specific `if podTemplatePodSpec == nil` or `if podSpec == nil` functionality here now that it is a separate function? I don't recall how `mergo.Merge` operates on nil data.

hamersaw (commented):
I think this looks great. A few nits to maybe look into.

Also I see the `PodTemplateSpec` [here](https://github.com/flyteorg/flyteplugins/blob/2ec7da85527cebccfff8f16e02895b06e9aa7a8a/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go#L87) and [here](https://github.com/flyteorg/flyteplugins/blob/2ec7da85527cebccfff8f16e02895b06e9aa7a8a/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go#L94) have an `ObjectMeta` field. In the `BuildPodWithSpec` we [maintain the metadata](https://github.com/flyteorg/flyteplugins/blob/85c512edc5d960f783187fa1c7c490067b608d02/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L209) to bring things like labels and annotations over. Do you know if we set `ObjectMeta` on the `PodTemplateSpec` will be be maintained in the `Pods` launched from the operator?

fg91:
> Also I see the `PodTemplateSpec` [here](https://github.com/flyteorg/flyteplugins/blob/2ec7da85527cebccfff8f16e02895b06e9aa7a8a/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go#L87) and [here](https://github.com/flyteorg/flyteplugins/blob/2ec7da85527cebccfff8f16e02895b06e9aa7a8a/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go#L94) have an `ObjectMeta` field. In the `BuildPodWithSpec` we [maintain the metadata](https://github.com/flyteorg/flyteplugins/blob/85c512edc5d960f783187fa1c7c490067b608d02/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L209) to bring things like labels and annotations over. Do you know if we set `ObjectMeta` on the `PodTemplateSpec` will be be maintained in the `Pods` launched from the operator?

No, the operator doesn't carry them over. I created a `PytorchJob` with the following `ObjectMeta`:

```yaml
apiVersion: "kubeflow.org/v1"
kind: PyTorchJob
metadata:
  name: pytorch-simple
  namespace: test
  labels:
    foo: bar
  annotations:
    afoo: abar
```

It results in Pods that look like this:

```yaml
apiVersion: v1
kind: Pod
metadata:
  creationTimestamp: "2022-11-30T10:07:42Z"
  labels:
    group-name: kubeflow.org
    job-name: pytorch-simple
    job-role: master
    replica-index: "0"
    replica-type: master
    training.kubeflow.org/job-name: pytorch-simple
    training.kubeflow.org/job-role: master
    training.kubeflow.org/operator-name: pytorchjob-controller
    training.kubeflow.org/replica-index: "0"
    training.kubeflow.org/replica-type: master
  name: pytorch-simple-master-0
  namespace: test
  ownerReferences:
  - apiVersion: kubeflow.org/v1
    blockOwnerDeletion: true
    controller: true
    kind: PyTorchJob
    name: pytorch-simple
    uid: 84b4e76d-de63-46a5-a829-c1595b7fd4bd
  resourceVersion: "30970195"
  uid: 7e5232d1-0deb-4013-bbba-cf350232c1db
```

fg91:
@hamersaw can you please take a look at the new changes? I tried to make sensible commits that make the review easy, maybe look at them one by one.


About unit tests:

The existing test `TestBuildPodWithSpec` tests the podSpec merging behaviour but it currently doesn't test that the object metadata is copied over.

* I copied the merging test logic from `TestBuildPodWithSpec` into the new test `TestMergePodSpecs`
* I created a new test `TestBuildPodWithSpec2` that checks that podSpec merging is taking place without testing any details of it. In addition, I now test that the object metadata is copied over.
* I left the `TestBuildPodWithSpec` test untouched even though the entire logic was moved into `TestBuildPodWithSpec` in order to give confidence that the refactoring didn't break anything that the old test would have caught. Maybe we should remove this test now though after this has been shown?

Having added these lines ...

```go
if podTemplate != nil {
    mergedPodSpec, err := flytek8s.MergePodSpecs(
```

... reduced the test coverage though. Do you have a suggestion how we can make the build green without mocking anything here?

hamersaw:
> No, the operator doesn't carry them over. I created a `PytorchJob` with the following `ObjectMeta`:
> 
> ```yaml
> apiVersion: "kubeflow.org/v1"
> kind: PyTorchJob
> metadata:
>   name: pytorch-simple
>   namespace: test
>   labels:
>     foo: bar
>   annotations:
>     afoo: abar
> ```

Oh sorry, I don't think I explained in-depth enough. I was thinking the `ObjectMeta` should be set on the `PodTemplateSpec` [like here](https://github.com/flyteorg/flyteplugins/blob/d5295d284c028be2b6d3c7ba9d6888e616d31551/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go#L76-L78). I haven't been able to see this carried over in the kf pytorch operator, but it looks like the [MPI controller just copies the PodSpec](https://github.com/kubeflow/training-operator/blob/25ba7c296818fa4ce7158f7933ff8e93a0ddbc6e/pkg/controller.v1/mpi/mpijob_controller.go#L964), which should carry over `ObjectMeta` fields. Does this sound right?

fg91:
> IIUC the updated default container names means that we need to create a container in the default PodTemplate called `pytorch` (because that's the [pytorch default container name](https://github.com/kubeflow/training-operator/blob/dce2b2ca97f5f4cae63e62bbadd4e203b2dfaf09/pkg/apis/kubeflow.org/v1/pytorch_types.go#L28)) which will be used as the base for all pytorch containers? So the `default` container will not be used? I'm assuming this is because you want to create a single `PodTemplate` for the entire deployment and use separate default container configuration for regular k8s tasks and pytorch jobs, can you say a little more about this? I kind of like the strict 'default' applies to everything approach, but just want to make sure I understand the change.

No, I'm currently using this pod template with a single container named `default`:

```yaml
apiVersion: v1
kind: PodTemplate
metadata:
  name: flyte-template
  namespace: flyte
template:
  spec:
    containers:
    - image: foo
      imagePullPolicy: Always
      name: default
      volumeMounts:
      - mountPath: /dev/shm
        name: dshm
    volumes:
    - emptyDir:
        medium: Memory
      name: dshm
```

With the current implementation this translates into this pod:

```yaml
apiVersion: v1
kind: Pod
...
spec:
  containers:
  - name: pytorch
    ...
    volumeMounts:
    - mountPath: /dev/shm
      name: dshm
  volumes:
  - emptyDir:
      medium: Memory
    name: dshm
...
```

hamersaw:
> No, I'm currently using this pod template with a single container named `default`:

Ah OK, small nit then. If we are not using the `defaultContainerName` parameter on the `MergePodSpecs` function can we just remove it? After than, lets merge!

fg91:
> Ah OK, small nit then. If we are not using the `defaultContainerName` parameter on the `MergePodSpecs` function can we just remove it? After that, looks great - lets merge!

I just realised that in `MergePodSpecs` I accidentally use the global vars `defaultContainerTemplateName` and `primaryContainerTemplateName` instead of the function args `primaryContainerName` and `defaultContainerName`:

```go
// merge template Containers
	var mergedContainers []v1.Container
	var defaultContainerTemplate, primaryContainerTemplate *v1.Container
	for i := 0; i < len(podTemplatePodSpecCopy.Containers); i++ {
		if podTemplatePodSpecCopy.Containers[i].Name == defaultContainerTemplateName {
			defaultContainerTemplate = &podTemplatePodSpecCopy.Containers[i]
		} else if podTemplatePodSpecCopy.Containers[i].Name == primaryContainerTemplateName {
			primaryContainerTemplate = &podTemplatePodSpecCopy.Containers[i]
		}
	}
```

That is why `defaultContainerName` isn't used.

Let me take another close look at this and also at setting the `ObjectMeta` on the `PodTemplateSpec` before we merge please.

fg91:
I tested the Pytorch Job plugin with the following template:

```yaml
apiVersion: v1
kind: PodTemplate
metadata:
  name: flyte-template
  namespace: flyte
template:
  metadata:
    labels:
      foo: bar                              # <- new
  spec:
    containers:
    - image: foo
      imagePullPolicy: Always
      name: default
      volumeMounts:
      - mountPath: /dev/shm
        name: dshm
    volumes:
    - emptyDir:
        medium: Memory
      name: dshm
```

The template leads to the following pod:

```yaml
apiVersion: v1
kind: Pod
metadata:
  labels:
    foo: bar
    group-name: kubeflow.org
    ....
  name: avjfhs8nsrq4jqm46hzr-feihpdka-0-master-0
  namespace: development
  ...
spec:
  affinity: {}
  containers:
  - args:
    - pyflyte-fast-execute
    ...
    name: pytorch
    ...
    volumeMounts:
    - mountPath: /dev/shm
      name: dshm
    ...
  volumes:
  - emptyDir:
      medium: Memory
    name: dshm
  ...
```

go/tasks/pluginmachinery/flytek8s/pod_helper_test.go
```diff
@@ -1013,6 +1014,144 @@ func TestDeterminePrimaryContainerPhase(t *testing.T) {
...
+	primaryContainer := pod.Spec.Containers[0]
+	assert.Equal(t, podSpec.Containers[0].Name, primaryContainer.Name)
+	assert.Equal(t, primaryContainerTemplate.TerminationMessagePath, primaryContainer.TerminationMessagePath)
+
+	// Test that template object metadata is copied
+	assert.Contains(t, pod.ObjectMeta.Labels, "fooKey")
+	assert.Equal(t, pod.ObjectMeta.Labels["fooKey"], "barVal")
+}
+
 func TestBuildPodWithSpec(t *testing.T) {
```
fg91, line 1155: @hamersaw Shell we keep this test (which I currently left in to demonstrate that the changes don't brick the existing tests) despite its logic having been refactored into `TestMergePodSpecs` and `TestBuildPodWithSpec2`?

go/tasks/plugins/k8s/kfoperators/mpi/mpi.go
```diff
@@ -67,6 +67,21 @@ func (mpiOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx plu
 		return nil, flyteerr.Errorf(flyteerr.BadTaskSpecification, "Unable to create pod spec: [%v]", err.Error())
 	}
 
+	common.OverrideDefaultContainerName(taskCtx, podSpec, kubeflowv1.MPIJobDefaultContainerName)
```
fg91, line 70: @hamersaw Do you think adding this line is incorrect?
As opposed to the pytorch and tfjob plugins, the container name is currently not overridden with `mpi` even though this is the default container name (see [here](https://github.com/kubeflow/training-operator/blob/ec53930b77e6c10adb75b966acb4bda13364ae8e/pkg/apis/kubeflow.org/v1/mpi_types.go#L28) and [here](https://github.com/kubeflow/training-operator/blob/ec53930b77e6c10adb75b966acb4bda13364ae8e/examples/mpi/tensorflow-mnist.yaml#L16)).

go/tasks/plugins/k8s/kfoperators/mpi/mpi.go
```diff
@@ -67,6 +67,21 @@ func (mpiOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx plu
...
+
+	podTemplate := flytek8s.DefaultPodTemplateStore.LoadOrDefault(taskCtx.TaskExecutionMetadata().GetNamespace())
+
+	objectMeta := metav1.ObjectMeta{}
```
fg91, line 74: @hamersaw this is new compared to the last review.

go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
```diff
@@ -68,20 +68,35 @@ func (pytorchOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx
 
 	common.OverrideDefaultContainerName(taskCtx, podSpec, kubeflowv1.PytorchJobDefaultContainerName)
```
fg91, line 69: @hamersaw As discussed I now first override the primary container name with the pytorch job default container name. Next, the podSpecs are merged, specifying the pytorch job default container name as the primary container name to the `MergePodSpecs` function.

fg91, line 69: @hamersaw As discussed I now first override the primary container name with the pytorch job default container name. Next, the podSpecs are merged, passing the pytorch job default container name as the primary container name to the `MergePodSpecs` function.

go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
```diff
@@ -68,20 +68,35 @@ func (pytorchOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx
...
 
+	podTemplate := flytek8s.DefaultPodTemplateStore.LoadOrDefault(taskCtx.TaskExecutionMetadata().GetNamespace())
+
+	objectMeta := metav1.ObjectMeta{}
```
fg91, line 73: This is new compared to the last review.

hamersaw (commented):
Looks great! Clean up tests and get another pair of eyes on it and lets merge!

hamersaw:
cc @bimtauer - this PR adds the default PodTemplate to the MPI task type. IIUC this should address [this issue](https://github.com/flyteorg/flyte/issues/3043) on IAM roles with annotations for the MPI tasks. Could you take a look through this to sanity check MPI updates? It would be VERY helpful!

fg91:
> Looks great! Clean up tests and get another pair of eyes on it and lets merge!

Renamed my test and deleted the old one :+1:

hamersaw (approved):

go/tasks/plugins/k8s/kfoperators/mpi/mpi.go
bimtauer: @hamersaw I will try and have a look by the end of the week!

PR #299, opened 2022-12-02
Add support Databricks plugin

pingsutw:
# TL;DR
Add databricks web api plugin. it will read databricks conf from flytekit, and send a rest job request to dbx platform.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [x] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
_NA_

## Follow-up issue
_NA_


go/tasks/plugins/webapi/databricks/plugin.go
```diff
@@ -0,0 +1,316 @@
...
+	err = json.Unmarshal(decodeBytes, &databricksJob)
+	if err != nil {
+		return nil, nil, fmt.Errorf("failed to unmarshal databricksJob: %v: %v", decodeBytes, err)
+	}
+
+	if err != nil {
+		return nil, nil, err
+	}
+
+	databricksJob["spark_python_task"] = map[string]interface{}{"python_file": p.cfg.EntrypointFile, "parameters": modifiedArgs}
```
tmathur-hbo, line 115: Does this assume, that all task code will be in one file or do you push it into master and workers? Also does this allow for the team to write packages?

go/tasks/plugins/webapi/databricks/plugin.go
```diff
@@ -0,0 +1,316 @@
...
+	return &ResourceMetaWrapper{runID, p.cfg.DatabricksInstance, token},
+		&ResourceWrapper{StatusCode: resp.StatusCode}, nil
+}
+
+func (p Plugin) Get(ctx context.Context, taskCtx webapi.GetContext) (latest webapi.Resource, err error) {
+	exec := taskCtx.ResourceMeta().(*ResourceMetaWrapper)
+	req, err := buildRequest(get, nil, p.cfg.databricksEndpoint,
+		p.cfg.DatabricksInstance, exec.Token, exec.RunID, false)
+	if err != nil {
+		logger.Info(ctx, "Failed to build databricks job request [%v]", err)
```
tmathur-hbo, line 147: Should it be logged as an error instead of info?

go/tasks/plugins/webapi/databricks/plugin.go
```diff
@@ -0,0 +1,316 @@
...
+		return nil, err
+	}
+	defer resp.Body.Close()
+	data, err := buildResponse(resp)
+	if err != nil {
+		return nil, err
+	}
+	logger.Info(ctx, "data data data %v", data)
+	message := fmt.Sprintf("%v", data["state"].(map[string]string)["state_message"])
+	jobID := fmt.Sprintf("%v", data["job_id"])
```
tmathur-hbo, line 162: is there a condition, where the job_id or state are not available?

go/tasks/plugins/webapi/databricks/plugin.go
```diff
@@ -0,0 +1,316 @@
...
+	databricksEndpoint string,
+	databricksInstance string,
+	token string,
+	runID string,
+	isCancel bool,
+) (*http.Request, error) {
+	var databricksURL string
+	// for mocking/testing purposes
+	if databricksEndpoint == "" {
+		databricksURL = "https://" + databricksInstance + ".cloud.databricks.com/api/2.0/jobs/runs"
```
tmathur-hbo, line 231: use fmt sprintf?

go/tasks/plugins/webapi/databricks/config.go
```diff
@@ -0,0 +1,74 @@
...
+type Config struct {
+	// WebAPI defines config for the base WebAPI plugin
+	WebAPI webapi.PluginConfig `json:"webApi" pflag:",Defines config for the base WebAPI plugin."`
+
+	// ResourceConstraints defines resource constraints on how many executions to be created per project/overall at any given time
+	ResourceConstraints core.ResourceConstraintsSpec `json:"resourceConstraints" pflag:"-,Defines resource constraints on how many executions to be created per project/overall at any given time."`
+
+	DefaultCluster string `json:"defaultWarehouse" pflag:",Defines the default warehouse to use when running on Databricks unless overwritten by the task."`
+
+	TokenKey string `json:"databricksTokenKey" pflag:",Name of the key where to find Databricks token in the secret manager."`
```
tmathur-hbo, line 59: In future, may be this tokenkey can be passed by the client

go/tasks/plugins/webapi/databricks/plugin.go
```diff
@@ -0,0 +1,316 @@
...
+
+	logger.Info(ctx, "statusCode statusCode %v", statusCode)
+	logger.Info(ctx, "lifeCycleState lifeCycleState lifeCycleState %v", lifeCycleState)
+
+	taskInfo := createTaskInfo(exec.RunID, jobID, exec.DatabricksInstance)
+	switch statusCode {
+	case http.StatusAccepted:
+		return core.PhaseInfoRunning(pluginsCore.DefaultPhaseVersion, taskInfo), nil
+	case http.StatusOK:
+		//if lifeCycleState == "TERMINATED" {
```
tmathur-hbo, line 207: remove commented code?

go/tasks/plugins/webapi/databricks/plugin.go
```diff
@@ -0,0 +1,316 @@
...
+}
+
+func (p Plugin) Create(ctx context.Context, taskCtx webapi.TaskExecutionContextReader) (webapi.ResourceMeta,
+	webapi.Resource, error) {
+	taskTemplate, err := taskCtx.TaskReader().Read(ctx)
+	if err != nil {
+		return nil, nil, err
+	}
+
+	token, err := taskCtx.SecretManager().Get(ctx, p.cfg.TokenKey)
```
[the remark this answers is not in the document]
pingsutw: currently, we can pass the secret through flyte task as well. https://docs.flyte.org/projects/cookbook/en/stable/auto/core/containerization/use_secrets.html

go/tasks/plugins/webapi/databricks/plugin.go
pingsutw: updated, thanks

go/tasks/plugins/webapi/databricks/plugin.go
pingsutw: updated, thanks

go/tasks/plugins/webapi/databricks/plugin.go
pingsutw: job_id and state are always available

go/tasks/plugins/webapi/databricks/plugin.go
pingsutw: updated, thanks

go/tasks/plugins/webapi/databricks/plugin.go
pingsutw: I wrote a entrypoint file like below. basically, we run this file first when running a spark job. it will execute the command we pass in the spark `parameters` (pyflyte-execute .....).

The reason to do that is because we need to run [pyflyte-fast-execute](https://github.com/flyteorg/flytekit/blob/2340e5dbaaea75f7b0c0418a92ace654b7bcf6d7/flytekit/bin/entrypoint.py) to execute some flytekit code. DBX doesn't allow us to override the default command, which always is `python spark_job.py`

What `pyflyte-fast-execute` does are
1. Download the workflow code (a zip file) to the driver. The code is uploaded when we register the workflow.
2. Download the task input (a proto file) from S3.
3. Deserialize proto to python value
4. Run the spark task
5. Serialize the output to proto
6. Upload the output to S3

```python
import os
import subprocess
import sys


def main():
    args = sys.argv
    print(args[1:])
    p = subprocess.run(args[1:], capture_output=True)
    print("==========================env==============================")
    print("AWS_ACCESS_KEY_ID", os.environ.get("AWS_ACCESS_KEY_ID"))
    print("==========================stdout==============================")
    print(p.stdout.decode('utf-8'))
    print("==========================stderr==============================")
    print(p.stderr.decode('utf-8'))


if __name__ == '__main__':
    main()

```

go/tasks/plugins/webapi/databricks/config.go
```diff
@@ -0,0 +1,74 @@
...
+		ResourceConstraints: core.ResourceConstraintsSpec{
+			ProjectScopeResourceConstraint: &core.ResourceConstraint{
+				Value: 100,
+			},
+			NamespaceScopeResourceConstraint: &core.ResourceConstraint{
+				Value: 50,
+			},
+		},
+		DefaultCluster: "COMPUTE_CLUSTER",
+		TokenKey:       "FLYTE_DATABRICKS_API_TOKEN",
```
tmathur-hbo, line 43: Should these values be defined as constants and referenced here?

go/tasks/plugins/webapi/databricks/integration_test.go
```diff
@@ -0,0 +1,111 @@
...
+
+		assert.Equal(t, true, phase.Phase().IsSuccess())
+	})
+}
+
+func newFakeDatabricksServer() *httptest.Server {
+	runID := "065168461"
+	jobID := "019e7546"
+	return httptest.NewServer(http.HandlerFunc(func(writer http.ResponseWriter, request *http.Request) {
+		if request.URL.Path == "/api/2.0/jobs/runs/submit" && request.Method == "POST" {
```
tmathur-hbo, line 73: would be cleaner if we used str formatting here

tmathur-hbo, line 73: why not use the constants here?

go/tasks/plugins/webapi/databricks/config.go
```diff
@@ -0,0 +1,77 @@
...
+
+	DefaultCluster string `json:"defaultWarehouse" pflag:",Defines the default warehouse to use when running on Databricks unless overwritten by the task."`
+
+	TokenKey string `json:"databricksTokenKey" pflag:",Name of the key where to find Databricks token in the secret manager."`
+
+	DatabricksInstance string `json:"databricksInstance" pflag:",Databricks workspace instance name."`
+
+	EntrypointFile string `json:"entrypointFile" pflag:",A URL of the entrypoint file. DBFS and cloud storage (s3:/, gcs:/, adls:/, etc) locations are supported."`
+	// databricksEndpoint overrides databricks instance endpoint, only for testing
+	databricksEndpoint string
```
wild-endeavor, line 68: what's an example of one?  it's not in the default config.

go/tasks/plugins/webapi/databricks/config.go
```diff
@@ -0,0 +1,77 @@
...
+	// ResourceConstraints defines resource constraints on how many executions to be created per project/overall at any given time
+	ResourceConstraints core.ResourceConstraintsSpec `json:"resourceConstraints" pflag:"-,Defines resource constraints on how many executions to be created per project/overall at any given time."`
+
+	DefaultCluster string `json:"defaultWarehouse" pflag:",Defines the default warehouse to use when running on Databricks unless overwritten by the task."`
+
+	TokenKey string `json:"databricksTokenKey" pflag:",Name of the key where to find Databricks token in the secret manager."`
+
+	DatabricksInstance string `json:"databricksInstance" pflag:",Databricks workspace instance name."`
+
+	EntrypointFile string `json:"entrypointFile" pflag:",A URL of the entrypoint file. DBFS and cloud storage (s3:/, gcs:/, adls:/, etc) locations are supported."`
```
wild-endeavor, line 66:
```suggestion
	EntrypointFile string `json:"entrypointFile" pflag:",A URL of the entrypoint file. DBFS and cloud storage (s3://, gcs://, adls://, etc) locations are supported."`
```
needs to be specified with // right?  better to match in comment to avoid confusion

wild-endeavor (dismissed):

go/tasks/plugins/webapi/databricks/config.go
pingsutw: it will only be used in [integration test](https://github.com/flyteorg/flyteplugins/blob/a8268d0e6622e1e7c3c21349ded1a7d307a2c2a9/go/tasks/plugins/webapi/databricks/integration_test.go#L37)

go/tasks/plugins/webapi/databricks/config.go
pingsutw: done, thanks

wild-endeavor (dismissed):

wild-endeavor (approved):

PR #300, opened 2022-12-02
sending correct external resources for k8s-array plugin

hamersaw:
# TL;DR
Cache status of map tasks subtasks is not reported correctly. This PR fixes the issue, where all external resources may be correctly annotated with the correct cache status during execution.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
^^^

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/plugins/array/catalog.go
+++ b/go/tasks/plugins/array/catalog.go
@@ -269,7 +269,6 @@ func WriteToDiscovery(ctx context.Context, tCtx core.TaskExecutionContext, state
 	for idx, phaseIdx := range state.ArrayStatus.Detailed.GetItems() {
 		phase := core.Phases[phaseIdx]
 		if !phase.IsSuccess() {
-			// TODO @hamersaw - prove that this is wrong!
 			// tasksToCache is built on the originalArraySize and ArrayStatus.Detailed is the executionArraySize
 			originalIdx := arrayCore.CalculateOriginalIndex(idx, state.GetIndexesToCache())
 			tasksToCache.Clear(uint(originalIdx))
--- a/go/tasks/plugins/array/catalog.go
+++ b/go/tasks/plugins/array/catalog.go
@@ -6,6 +6,7 @@ import (
 	"math"
 	"strconv"
 
+	idlCore "github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
 	idlPlugins "github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/plugins"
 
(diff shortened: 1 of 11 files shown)


go/tasks/plugins/array/catalog.go
```diff
@@ -257,18 +257,21 @@ func WriteToDiscovery(ctx context.Context, tCtx core.TaskExecutionContext, state
...
 	// Do not cache failed tasks. Retrieve the final phase from array status and unset the non-successful ones.
+
 	tasksToCache := state.GetIndexesToCache().DeepCopy()
 	for idx, phaseIdx := range state.ArrayStatus.Detailed.GetItems() {
 		phase := core.Phases[phaseIdx]
 		if !phase.IsSuccess() {
-			tasksToCache.Clear(uint(idx))
+			// tasksToCache is built on the originalArraySize and ArrayStatus.Detailed is the executionArraySize
+			originalIdx := arrayCore.CalculateOriginalIndex(idx, state.GetIndexesToCache())
+			tasksToCache.Clear(uint(originalIdx))
```
EngHabu, line 274: Ouch

go/tasks/plugins/array/core/state.go
```diff
@@ -267,8 +263,9 @@ func SummaryToPhase(ctx context.Context, minSuccesses int64, summary arraystatus
 		return PhaseWriteToDiscoveryThenFail
 	}
 
-	// No chance to reach the required success numbers.
-	if totalRunning+totalSuccesses+totalWaitingForResources+totalRetryableFailures < minSuccesses {
+	// No chance to reach the required success numbers. If any subtasks are running we let them
+	// complete, so results may be written to cache if enabled, before failing.
+	if totalSuccesses+totalWaitingForResources+totalRetryableFailures < minSuccesses && totalRunning == 0 {
```
EngHabu, line 268: Should we just add total running to the +s

hamersaw: OK, so I felt strongly enough to change it, but not enough to defend the change. I reverted this to the previous approach. Basically, I think transitioning to permanent failure should operate differently whether the task is cached or not:
Cached: execute as many of the subtasks as possible to allow the cached outputs to be written.
Not cached: fail as soon as possible so you don't spend any time doing unnecessary work (current approach).

cosmicBboy:
@EngHabu friendly ping: does this PR look good to go? Currently blocking https://github.com/flyteorg/flyte/issues/3315

PR #306, opened 2023-01-06
Add env FLYTE_FAIL_ON_ERROR to aws batch job

pingsutw:

# TL;DR
Add env FLYTE_FAIL_ON_ERROR to aws batch job, so the flytekit will return error code once the job fail.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
image

## Tracking Issue
https://flyte-org.slack.com/archives/C01P3B761A6/p1664972219043289

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/plugins/array/awsbatch/transformer.go
+++ b/go/tasks/plugins/array/awsbatch/transformer.go
@@ -118,7 +118,7 @@ func UpdateBatchInputForArray(_ context.Context, batchInput *batch.SubmitJobInpu
 		envVars = append(envVars, &batch.KeyValuePair{Name: refStr(ArrayJobIndex), Value: refStr("FAKE_JOB_ARRAY_INDEX")},
 			&batch.KeyValuePair{Name: refStr("FAKE_JOB_ARRAY_INDEX"), Value: refStr("0")})
 	}
-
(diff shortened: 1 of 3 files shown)


go/tasks/plugins/array/awsbatch/transformer.go
```diff
@@ -144,7 +144,7 @@ func getEnvVarsForTask(ctx context.Context, execID pluginCore.TaskExecutionID, c
 			Value: val,
 		})
 	}
-
+	finalEnvVars = append(finalEnvVars, v1.EnvVar{Name: failOnError, Value: "true"})
```
hamersaw, line 147: Can we instead use something like:
```
m[failOnError] = "true"
```
on like line 139? Then it will be populated in the `finalEnvVars` as part of the loop and the array with be right-sized rather than reallocating capacity because it's set as `len(m)`. Also, it stops from double-setting the env var.

pingsutw: done, thanks

hamersaw (approved):

PR #310, opened 2023-01-20
Return nil if ResourceMeta is nil when deleting the node

pingsutw:
Fixes https://github.com/flyteorg/flyte/issues/3281

# TL;DR
Transition to failed state if the task isn't created. otherwise, the task will run forever.

1 round (phase: `PhaseNotStarted`)
- Webapi plugin try to [allocate token](https://github.com/flyteorg/flyteplugins/blob/e7bf9a4e1b43ec7be55411fcc31babbe6c769db4/go/tasks/pluginmachinery/internal/webapi/core.go#L82)

2 round (state: `PhaseAllocationTokenAcquired`, phase: `queued`):
- Plugin fail to create a task, and will not update the task state. The state is `PhaseAllocationTokenAcquired` in this round
- Task handler [send](https://github.com/flyteorg/flytepropeller/blob/01ec2147cc830ec8271f3423e88d82cc23a68609/pkg/controller/nodes/task/handler.go#L737) the task event (queued) to admin
- However we transition the phase from queued to running [here](https://github.com/flyteorg/flytepropeller/blob/01ec2147cc830ec8271f3423e88d82cc23a68609/pkg/controller/nodes/task/handler.go#L186)
- And we will [record](https://github.com/flyteorg/flytepropeller/blob/411457274c99094951045c02e003ab0653d9cf5d/pkg/controller/nodes/executor.go#L504) the node event (running) in the node executor

3 round (state: `PhaseAllocationTokenAcquired`, phase: `running`):
- Will try to create a task, and fail again 
- In the screenshot below, you can see the node phase is `running`, but the task state is `queued`

4+ round
- Repeat 3 round

Update: we should just return nil if task metadata is nil

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
Before
image

After
image

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3281

It's related to this [issue](https://flyte-org.slack.com/archives/CP2HDHKE1/p1674074427141929)
propeller panic because [taskCtx.ResourceMeta()](https://github.com/flyteorg/flyteplugins/blob/3ec161aaa3885ebc7ec199da91d605611ef7ab1b/go/tasks/plugins/webapi/databricks/plugin.go#L186) is nil. However, it should never happen. it was happening because we tried to abort the task that doesn't exist.

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/pluginmachinery/core/resource_manager.go
+++ b/go/tasks/pluginmachinery/core/resource_manager.go
@@ -37,20 +37,18 @@ type ResourceRegistrar interface {
 
 // ResourceManager Interface
 // 1. Terms and definitions
-//
 //   - Resource: resource is an abstraction of anything that has a limited quota of units and can be claimed in a
 //     single unit or multiple units at once. At Flyte's current state, a resource means a logical
 //     separation (e.g., a cluster) of an external service that allows a limited number of outstanding
 //     requests to be sent to.
-//
 //   - Token: Flyte uses a token to serve as the placeholder to represent a unit of resource. Flyte resource manager
 //     manages resources by managing the tokens of the resources.
 //
-//     2. Description
+//  2. Description
 //     ResourceManager provides a task-type-specific pooling system for Flyte Tasks. Plugin writers can optionally
 //     request for resources in their tasks, in single quantity.
 //
-//     3. Usage
+//  3. Usage
 //     A Flyte plugin registers the resources and the desired quota of each resource with ResourceRegistrar at the
 //     setup time of Flyte Propeller. At the end of the setup time, Flyte Propeller builds a ResourceManager based on
 //     these registration requests.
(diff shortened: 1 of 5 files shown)


hamersaw:
I think this should be left as is. The problem here was the panic on when attempted to abort on a `nil` resource right? Because right now, if there is a failure to create it, then propeller will retry up to the number of configured attempts and fail the task. Then it should abort the task during the transition to failure. In the k8s plugin manager we [ignore failures if the resource does not exist](https://github.com/flyteorg/flytepropeller/blob/20146304738db96e8a15cf9a824233676a7560ea/pkg/controller/nodes/task/k8s/plugin_manager.go#L378).

IMO this should be handled by only attempting to [delete the resource](https://github.com/flyteorg/flyteplugins/blob/4634a81403e501882e3b1f39bcbd229b78768a4e/go/tasks/pluginmachinery/internal/webapi/core.go#L111) during the abort call if the `incomingState.ResourceMeta != nil`.

hamersaw:
Or it should be handled in the plugin. According to the [docs](https://github.com/flyteorg/flyteplugins/blob/4634a81403e501882e3b1f39bcbd229b78768a4e/go/tasks/pluginmachinery/webapi/plugin.go#L138-L140):

`If the resource has already been deleted, the API should not fail.`

So we might need to be more careful in the plugin.

pingsutw:
@hamersaw thanks, that makes sense. updated it.

hamersaw (approved):

PR #313, opened 2023-01-29
Discard fields when replica number equals zero to avoid api client error

ByronHsu:
# TL;DR

In some versions of tensorflow CRD, there is an enforcement to restrict the minimum number of replica. If users pass in 0 as the number of replicas, the k8s client will **throw an error** at runtime. The field should be discarded if the number of replicas is zero.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

### Bug
In [the order version of tfjob CRD](https://gist.github.com/ByronHsu/2b484d53bcb4a361bf9af9fb303c8896), it will validate the minimum number of replicas. If the replica == 0 but the field exists, the kubeflow client will fail.

## Test

```
byhsu@byhsu-ld1 ~/local-repo/flyte/flyteplugins gh-tf-replica-bug* ❯ go test github.com/flyteorg/flyteplugins/go/tasks/plugins/k8s/kfoperators/tensorflow
ok      github.com/flyteorg/flyteplugins/go/tasks/plugins/k8s/kfoperators/tensorflow    0.072s
```

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

### DIFF
--- a/go/tasks/plugins/k8s/kfoperators/mpi/mpi.go
+++ b/go/tasks/plugins/k8s/kfoperators/mpi/mpi.go
@@ -13,7 +13,7 @@ import (
 	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/k8s"
 	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/utils"
 	"github.com/flyteorg/flyteplugins/go/tasks/plugins/k8s/kfoperators/common"
-	commonKf "github.com/kubeflow/common/pkg/apis/common/v1"
+	commonOp "github.com/kubeflow/common/pkg/apis/common/v1"
 	kubeflowv1 "github.com/kubeflow/training-operator/pkg/apis/kubeflow.org/v1"
 	v1 "k8s.io/api/core/v1"
 	metav1 "k8s.io/apimachinery/pkg/apis/meta/v1"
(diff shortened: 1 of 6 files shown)


goyalankit:
@ByronHsu Perhaps also include the symptoms of this bug in the PR. i.e., chief and ps processes are shown immediately terminated

hamersaw:
@ByronHsu thanks, this is awesome! Can you maybe link to the TF CRD version changes? So if you specify 0 for one of the replica sets (ex. chief, worker, etc) then it defaults to a minimum number of replicas? You can't control the replica configurations (ie. `Template`, `RestartPolicy`) then right?

ByronHsu:
@hamersaw We used the CRD in `tf-operator`, but It might because tf-operator repository is depreacted, so I cannot find it online. The CRD we used is [this](https://gist.github.com/ByronHsu/2b484d53bcb4a361bf9af9fb303c8896).

> So if you specify 0 for one of the replica sets (ex. chief, worker, etc) then it defaults to a minimum number of replicas? 
It will not default to the min. It will fail instead.

hamersaw (approved):

hamersaw:
I'm going to assume that since these are all kf operators this functionality is the same for mpi and pytorch. Can we make the quick change to update those two plugins the same way this is updated?

ByronHsu:
It seems that mpi[https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/plugins/k8s/kfoperators/mpi/mpi.go#L94] already handled, but [pytorch](https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go#L88) didn't. Our use cases are only mpi and tf, can you help on the pytorch part?

ByronHsu:
@hamersaw Thanks for unifying kf components!

hamersaw (approved):

PR #329, opened 2023-03-09
Add support raw container in the map task

pingsutw:
# TL;DR
When using regular task in the map task, the entrypoint is `pyflyte-map-execute`. it does two things before running the task. 
1. Update sub task input / output interface to [list interface](https://github.com/flyteorg/flytekit/blob/be24c52f3fb8f7949db172f490cca95fa0f0e413/flytekit/core/map_task.py#L56) 
2. Append job id to [output prefix](https://github.com/flyteorg/flytekit/blob/be24c52f3fb8f7949db172f490cca95fa0f0e413/flytekit/bin/entrypoint.py#L401)

However, when using [raw container](https://docs.flyte.org/projects/cookbook/en/latest/auto/core/containerization/raw_container.html), the entrypoint will not be `pyflyte-map-execute`. 

Therefore, we should update the output prefix for copilot, and [support upload collection](https://github.com/flyteorg/flytecopilot/pull/54) in the copilot.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
image

## Tracking Issue
https://flyte-org.slack.com/archives/CP2HDHKE1/p1678230956906899

## Follow-up issue
_NA_


go/tasks/plugins/array/k8s/subtask.go
```diff
@@ -186,6 +186,17 @@ func launchSubtask(ctx context.Context, stCtx SubTaskExecutionContext, cfg *Conf
 		Value: strconv.Itoa(stCtx.originalIndex),
 	})
 
+	for sidecarIndex, container := range pod.Spec.Containers {
+		if container.Name == config.GetK8sPluginConfig().CoPilot.NamePrefix+flytek8s.Sidecar {
+			for i, arg := range pod.Spec.Containers[sidecarIndex].Args {
```
pingsutw, line 191: cc @hamersaw should we pass env `FlyteK8sArrayIndex` to copilot, and construct final output prefix in the copilot?

hamersaw: I don't have strong feelings here. How are we passing the array index to the inputs downloader? Because in flytekit we pass the input data ref and a subtask index, IIUC it reads the full list of inputs and only uses the value at the subtask index. We need to do the same thing here right?

go/tasks/plugins/k8s/pod/plugin.go
```diff
@@ -184,8 +191,17 @@ func (plugin) GetTaskPhaseWithLogs(ctx context.Context, pluginContext k8s.Plugin
...
 
+	// When the copilot is running, we should wait until the data is uploaded by the copilot.
+	copilotContainerName, exists := r.GetAnnotations()[flytek8s.FlyteCopilotName]
+	if exists {
+		copilotContainerPhase := flytek8s.DetermineContainerPhase(copilotContainerName, pod.Status.ContainerStatuses, &info)
+		if copilotContainerPhase.Phase() == pluginsCore.PhaseRunning && len(info.Logs) > 0 {
+			return pluginsCore.PhaseInfoRunning(pluginsCore.DefaultPhaseVersion+1, copilotContainerPhase.Info()), nil
+		}
+	}
+
```
hamersaw, line 202: So right now this is done in `ContainerTasks` (not using map task) by just not setting a `primaryContainerName` on the `Pod`, then this code waits for the entire `Pod` to complete. It seems like this is what we should do for subtasks as well.

I think the issue is that [here](https://github.com/flyteorg/flyteplugins/blob/0a681cd1a9e2c6586e6408532767539be65737f4/go/tasks/plugins/array/k8s/subtask.go#L325-L344) we always add a `PrimaryContainerName` to the pod annotation. I think it makes sense to update the code in the subtask.go so that this annotation is only added if necessary, then we shouldn't need to add to `flytek8s.FlyteCopilotName` annotation above either.

hamersaw (commented):
I think this is looking very good - thanks for diving into it! My general feedback is I think we can leave the pod plugin code as is. If we can encapsulate the code for maptasks in the [k8s_array package](https://github.com/flyteorg/flyteplugins/tree/master/go/tasks/plugins/array/k8s) then when we implement `ArrayNode` and ultimately refactor this all out it doesn't leave nasty one-off code to support legacy functionality. What are your thoughts on this? Very open for discussion.

go/tasks/plugins/array/k8s/subtask.go
pingsutw: we pass array index to [primary container](https://github.com/flyteorg/flyteplugins/blob/d20137836713ba83acdd029cb1f83eb28b82ceb4/go/tasks/plugins/array/k8s/subtask.go#L183-L188), and raw container task will read the value at subtask index. here is an example. https://github.com/flyteorg/flytekit/pull/1547.

The problem is that in the regular map task, we construct the final output prefix in the flytekit (output_prefix + array index.), but the raw container doesn't know the output prefix, it write to a local share dir instead. uploader will read the data in the share dir and upload to s3.

go/tasks/plugins/k8s/pod/plugin.go
pingsutw: The issue is the propeller will only pass array index to primary container, so we have to set raw-container to primary. However, if we set it to primary, propeller won't wait for the uploader complete, so I added `flytek8s.FlyteCopilotName` to annotation, and wait for copilot first if we find the uploader container in the pod here.

go/tasks/plugins/array/k8s/subtask.go
```diff
@@ -330,6 +346,10 @@ func getTaskContainerIndex(pod *v1.Pod) (int, error) {
 		if len(pod.Spec.Containers) == 1 {
 			return 0, nil
 		}
+		// Copilot is always the second container if it is enabled.
+		if len(pod.Spec.Containers) == 2 && pod.Spec.Containers[1].Name == config.GetK8sPluginConfig().CoPilot.NamePrefix+flytek8s.Sidecar {
```
hamersaw, line 350: I don't think it is reasonable to assume this. Could there a scenario where the second container was not copilot?

PR #330, opened 2023-03-13
External Plugin Service (grpc) 

pingsutw:
# TL;DR
Add a new grpc plugin that can send create / get / delete task request to the flyteplugins-service deployment.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
^^^

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3282

## Follow-up issue
_NA_


go/tasks/plugins/webapi/grpc/config.go
```diff
@@ -0,0 +1,73 @@
...
+
+	// ResourceConstraints defines resource constraints on how many executions to be created per project/overall at any given time
+	ResourceConstraints core.ResourceConstraintsSpec `json:"resourceConstraints" pflag:"-,Defines resource constraints on how many executions to be created per project/overall at any given time."`
+
+	GrpcTokenKey string `json:"grpcTokenKey" pflag:",Name of the key where to find grpc access token in the secret manager."`
+
+	DefaultGrpcEndpoint string `json:"defaultGrpcEndpoint" pflag:",The default grpc endpoint of flyteplugins service."`
+
+	// Maps endpoint to their plugin handler. {TaskType: Endpoint}
+	EndpointForTaskTypes map[string]string `json:"endpointForTaskTypes" pflag:"-,"`
```
hamersaw, line 64: In the IDL proposal I think there is a specific `task_type` in a few of the requests. Is the multiplexing happening here in plugins (ie. call a different endpoint based on the `task_type`) or in a backend plugin (ie. call a grpc endpoint which hosts multiple plugins and chooses the correct one based on the `task_type` in the request)

go/tasks/plugins/webapi/grpc/plugin.go
```diff
@@ -0,0 +1,186 @@
+package grpc
+
+import (
+	"context"
+	"encoding/gob"
+	"fmt"
+
+	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/ioutils"
```
hamersaw, line 8: Can we cleanup this import?

go/tasks/plugins/webapi/grpc/plugin.go
```diff
@@ -0,0 +1,186 @@
...
+
+func getFinalEndpoint(taskType, defaultEndpoint string, endpointForTaskTypes map[string]string) string {
+	if t, exists := endpointForTaskTypes[taskType]; exists {
+		return t
+	}
+
+	return defaultEndpoint
+}
+
+func getClientFunc(endpoint string) (service.BackendPluginServiceClient, *grpc.ClientConn, error) {
```
hamersaw, line 156: So this creates a new grpc client for each call?

go/tasks/plugins/webapi/grpc/plugin.go
pingsutw: fixed

go/tasks/plugins/webapi/grpc/plugin.go
pingsutw: That's because the endpoint is created at runtime. It's based on task type. different tasks may send to different grpc server.

go/tasks/plugins/webapi/grpc/plugin.go
```diff
@@ -0,0 +1,189 @@
...
+
+func getFinalEndpoint(taskType, defaultEndpoint string, endpointForTaskTypes map[string]string) string {
+	if t, exists := endpointForTaskTypes[taskType]; exists {
+		return t
+	}
+
+	return defaultEndpoint
+}
+
+func getClientFunc(endpoint string) (service.ExternalPluginServiceClient, *grpc.ClientConn, error) {
```
hamersaw, line 156: So the endpoint is determined by the taskType with a fallback to a default endpoint, but we should still at least create a cache of gRPC connections (ie `map[endpoint]grpcConn`) so that we don't incur network overhead on each call. Initializing a new TCP connection for each call will make scaling very inefficient.

go/tasks/plugins/webapi/grpc/plugin.go
```diff
@@ -0,0 +1,207 @@
...
+				grpclog.Infof("Failed to close conn to %s: %v", endpoint, cerr)
+			}
+		}()
+	}()
+	return service.NewExternalPluginServiceClient(conn), nil
+}
+
+func newGrpcPlugin() webapi.PluginEntry {
+	supportedTaskTypes := GetConfig().SupportedTaskTypes
+	supportedTaskTypes = append(supportedTaskTypes, "bigquery_query_job_task")
```
hamersaw, line 186: "bigquery_query_job_task" shouldn't be hardcoded here right?

pingsutw: removed it. yes, I think it's better to add it to the configmap in the helm chart.

hamersaw (approved):

PR #335, opened 2023-03-27
Inject container resource during BuildRawContainer

ByronHsu:
# TL;DR

Inject container resource during BuildRawContainer

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
### Describe the bug

Note: This bug does not affect the behavior of Flyte, but the logic contains a flaw.

In [BuildRawContainer](https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/pluginmachinery/flytek8s/container_helper.go#L213), we do not inject the container resource from the  task template. Instead, we inject the resource from [parameters.TaskExecMetadata.GetOverrides().GetResources()](https://github.com/flyteorg/flytepropeller/blob/master/pkg/compiler/transformers/k8s/node.go#L49).

This is because for both `_with_override(resource=...` and `@task(requests=...)`, they are both inject as the node resource override in [buildNodeSpec](https://github.com/flyteorg/flytepropeller/blob/master/pkg/compiler/transformers/k8s/node.go#L49)

### Expected behavior

For the case of `@task(requests=...)`, the resource override on the node should be empty, and the resource is injected from tasktemplate at [BuildRawContainer](https://github.com/flyteorg/flyteplugins/blob/master/go/tasks/pluginmachinery/flytek8s/container_helper.go#L213)

## Tracking Issue

https://github.com/flyteorg/flyte/issues/3540

### DIFF
--- a/go/tasks/pluginmachinery/flytek8s/container_helper.go
+++ b/go/tasks/pluginmachinery/flytek8s/container_helper.go
@@ -204,6 +204,12 @@ func BuildRawContainer(ctx context.Context, taskContainer *core.Container, taskE
 		containerName = rand.String(4)
 	}
 
+	res, err := ToK8sResourceRequirements(taskContainer.Resources)
+
+	if err != nil {
+		return nil, err
+	}
+
 	container := &v1.Container{
 		Name:                     containerName,
 		Image:                    taskContainer.GetImage(),
@@ -211,6 +217,7 @@ func BuildRawContainer(ctx context.Context, taskContainer *core.Container, taskE
 		Command:                  taskContainer.GetCommand(),
 		Env:                      ToK8sEnvVar(taskContainer.GetEnv()),
 		TerminationMessagePolicy: v1.TerminationMessageFallbackToLogsOnError,
+		Resources:                *res,
 	}
 
 	return container, nil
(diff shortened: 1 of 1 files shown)


go/tasks/pluginmachinery/flytek8s/container_helper.go
```diff
@@ -303,9 +310,7 @@ func AddFlyteCustomizationsToContainer(ctx context.Context, parameters template.
 			" Resources [%v] with mode [%v]", res, platformResources, container.Resources, mode)
 
 		switch mode {
-		case ResourceCustomizationModeAssignResources:
-			container.Resources = ApplyResourceOverrides(*res, *platformResources, assignIfUnset)
-		case ResourceCustomizationModeMergeExistingResources:
+		case ResourceCustomizationModeAssignResources, ResourceCustomizationModeMergeExistingResources:
```
hamersaw, line 313: I don't think that removing support for `ResourceCustomizationModeAssignResources` is the right call here. Could we just update the call to `AddFlyteCustomizationsToContainer` in the [ToK8sContainer function](https://github.com/flyteorg/flyteplugins/blob/18a594ecef8bf744b42d9a824df2d53635aa31d2/go/tasks/pluginmachinery/flytek8s/container_helper.go#L254) to use `ResourceCustomizationModeMergeExistingResources` instead? IIUC this will achieve the same thing, but please correct me.

hamersaw (approved):

PR #341, opened 2023-04-03
Pass `worker_spec_command` to mpi plugin to support horovod

ByronHsu:
# TL;DR
Pass `worker_spec_command` to mpi plugin to support horovod task

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
In this pr in flytekit, we pass `worker_spec_command` to backend to be run on worker pod.

## Tracking Issue
https://github.com/flyteorg/flyte/issues/3567

### DIFF
--- a/go/tasks/plugins/k8s/kfoperators/mpi/mpi.go
+++ b/go/tasks/plugins/k8s/kfoperators/mpi/mpi.go
@@ -72,6 +72,9 @@ func (mpiOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx plu
 
 	// workersPodSpec is deepCopy of podSpec submitted by flyte
 	workersPodSpec := podSpec.DeepCopy()
+
+	// If users don't specify "worker_spec_command" in the task config, the command/args are empty.
+	// In the case of horovod tasks, each worker runs a command launching ssh daemon.
 	workerSpecCommand := []string{}
 	if val, ok := taskTemplateConfig["worker_spec_command"]; ok {
 		workerSpecCommand = strings.Split(val, " ")
--- a/go/tasks/plugins/k8s/kfoperators/mpi/mpi.go
+++ b/go/tasks/plugins/k8s/kfoperators/mpi/mpi.go
@@ -3,6 +3,7 @@ package mpi
 import (
 	"context"
 	"fmt"
+	"strings"
 	"time"
 
 	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/plugins"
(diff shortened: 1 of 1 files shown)


go/tasks/plugins/k8s/kfoperators/mpi/mpi.go
```diff
@@ -69,11 +71,14 @@ func (mpiOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx plu
...
 	workersPodSpec := podSpec.DeepCopy()
+	workerSpecCommand := []string{}
+	if val, ok := taskTemplateConfig["worker_spec_command"]; ok {
+		workerSpecCommand = strings.Split(val, " ")
+	}
 
 	for k := range workersPodSpec.Containers {
-		workersPodSpec.Containers[k].Args = []string{}
+		workersPodSpec.Containers[k].Args = workerSpecCommand
 		workersPodSpec.Containers[k].Command = []string{}
```
pingsutw, line 82: Should we use `workerSpecCommand` for container's command? If not, I think we could rename it to `workerSpecArgs`
```suggestion
		workersPodSpec.Containers[k].Args = []string{}
		workersPodSpec.Containers[k].Command = workerSpecCommand
```

ByronHsu: I followed the same pattern as `_get_container` of `PythonAutoContainer`, where we put [get_command](https://github.com/flyteorg/flytekit/blob/master/flytekit/core/python_auto_container.py#L186) in `args` of the container. 

I think on the user-facing side (flytekit), we can call them commands, but on the backend side (flytepropeller), we can put the command in args.

Putting either in command or args of the container both work I guess. I just want to make it consistent with the other cases

hamersaw (commented):
@ByronHsu I'm not completely following this. Hoping you can clarify faster than I can look through the code. In the current implementation there is a comment `WorkerPodSpec doesn't need any Argument & command. It will be trigger from launcher pod` implying that the kubeflow operator (or launcher Pod) will override the command set in the worker PodSpec. Is that not actually the case? In this PR it seems we're setting the command which means it is maintained through the launcher. Does the kf operator use the command if it's provided and set it if it's not? Perhaps this should be documented here?

hamersaw:
Also, need DCO signoff, thanks!

go/tasks/plugins/k8s/kfoperators/mpi/mpi.go
```diff
@@ -69,11 +71,19 @@ func (mpiOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx plu
...
 	// workersPodSpec is deepCopy of podSpec submitted by flyte
-	// WorkerPodSpec doesn't need any Argument & command. It will be trigger from launcher pod
 	workersPodSpec := podSpec.DeepCopy()
 
+	// If users don't specify "worker_spec_command" in the task config, the command/args are empty.
+	// However, in some cases, the workers need command/args.
+	// For example, in horovod tasks, each worker runs a command launching ssh daemon.
+
+	workerSpecCommand := []string{}
+	if val, ok := taskTemplateConfig["worker_spec_command"]; ok {
```
hamersaw, line 81: can we make `worker_spec_command` a `const`?

ByronHsu: but `worker_spec_command` is mutatable. See line 82

hamersaw: yes, i understand. i mean:
```
const workerSpecCommandKey = "worker_spec_command"

...
if val, ok := taskTemplateConfig[workerSpecCommandKey]; ok {
    ...
```

hamersaw (approved):

PR #345, opened 2023-04-21
Change kubeflow plugins to allow settings specs for different replica

yubofredwang:
# TL;DR
Change kubeflow plugins to allow settings specs for different replica. The settings include: images, resources, command(mpi only), restart policy for different replica groups.

Corresponding PR in flyteidl: https://github.com/flyteorg/flyteidl/pull/386

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Tracking Issue

fixes https://github.com/flyteorg/flyte/issues/3308

### DIFF
--- a/go/tasks/plugins/k8s/kfoperators/common/common_operator.go
+++ b/go/tasks/plugins/k8s/kfoperators/common/common_operator.go
@@ -234,11 +234,16 @@ func OverrideContainerSpec(podSpec *v1.PodSpec, containerName string, image stri
 				podSpec.Containers[idx].Image = image
 			}
 			if resources != nil {
-				resources, err := flytek8s.ToK8sResourceRequirements(resources)
-				if err != nil {
-					return nil, flyteerr.Errorf(flyteerr.BadTaskSpecification, "invalid TaskSpecification on Resources [%v], Err: [%v]", resources, err.Error())
+				// if resources requests and limits both not set, we will not override the resources
+				if len(resources.Requests) >= 1 || len(resources.Limits) >= 1 {
+					resources, err := flytek8s.ToK8sResourceRequirements(resources)
+					if err != nil {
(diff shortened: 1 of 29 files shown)


kumare3:
cc @fg91 you are also working on this file

go/tasks/plugins/k8s/kfoperators/common/common_operator.go
```diff
@@ -180,3 +188,67 @@ func OverridePrimaryContainerName(podSpec *v1.PodSpec, primaryContainerName stri
...
+				podSpec.Containers[idx].Image = image
+			}
+			if resources != nil {
+				// if resources requests and limits both not set, we will not override the resources
+				if len(resources.Requests) >= 1 || len(resources.Limits) >= 1 {
+					resources, err := flytek8s.ToK8sResourceRequirements(resources)
+					if err != nil {
+						return nil, flyteerr.Errorf(flyteerr.BadTaskSpecification, "invalid TaskSpecificat ion on Resources [%v], Err: [%v]", resources, err.Error())
+					}
+					podSpec.Containers[idx].Resources = *resources
```
hamersaw, line 243: IIUC resources on the idl message will override completely. This is exactly how it's handled in other plugins (ie. dask / etc), but I think it's at least worth bringing up. Does it make sense to have more complex resource logic, for example merging the resources set here with the default resources?

hamersaw (commented):
Only nit - can we add docs to the function headers? It's something we trying to be more strict about.

go/tasks/plugins/k8s/kfoperators/common/common_operator.go
yubofredwang: My two cents is: as a user I would prefer to have consistency. If I override the default resources with plugin config resources. I want that to be the source of truth instead of merging. Just my thoughts though

hamersaw (dismissed):

hamersaw (approved):

PR #347, opened 2023-04-26
Fix double http in the Spark Driver UI Link

mucahitkantepe:
Fixes https://github.com/flyteorg/flyte/issues/3623

Spark operator code: https://github.com/GoogleCloudPlatform/spark-on-k8s-operator/blob/5f2efd4ff97e7c0bfdb726a066118d3401576730/pkg/controller/sparkapplication/sparkui.go#L57

This change solves the problem by keeping Flyte compatible with older version of Spark Operator.
Another, simpler implentation is to just pass whatever spark operator gives. Browsers will add https automatically?

### DIFF
--- a/go/tasks/plugins/k8s/spark/spark.go
+++ b/go/tasks/plugins/k8s/spark/spark.go
@@ -407,7 +407,7 @@ func getEventInfoForSpark(sj *sparkOp.SparkApplication) (*pluginsCore.TaskInfo, 
 			})
 		}
 	} else if sj.Status.AppState.State == sparkOp.RunningState && sj.Status.DriverInfo.WebUIIngressAddress != "" {
-		// Older versions of spark-operator does not append http:// but newer versions do. 
+		// Older versions of spark-operator does not append http:// but newer versions do.
 		uri := strings.TrimPrefix(sj.Status.DriverInfo.WebUIIngressAddress, "http://")
 		customInfoMap[sparkDriverUI] = fmt.Sprintf("https://%s", uri)
 		// Custom doesn't work unless the UI has a custom plugin to parse this, hence add to Logs as well.
--- a/go/tasks/plugins/k8s/spark/spark.go
+++ b/go/tasks/plugins/k8s/spark/spark.go
@@ -407,8 +407,9 @@ func getEventInfoForSpark(sj *sparkOp.SparkApplication) (*pluginsCore.TaskInfo, 
 			})
 		}
 	} else if sj.Status.AppState.State == sparkOp.RunningState && sj.Status.DriverInfo.WebUIIngressAddress != "" {
-		// Append https as the operator doesn't currently.
-		customInfoMap[sparkDriverUI] = fmt.Sprintf("https://%s", sj.Status.DriverInfo.WebUIIngressAddress)
+		// Older versions of spark-operator does not have http:// but newer versions have. 
+		customInfoMap[sparkDriverUI] = strings.TrimPrefix(sj.Status.DriverInfo.WebUIIngressAddress, "http://")
+		customInfoMap[sparkDriverUI] = fmt.Sprintf("https://%s", customInfoMap[sparkDriverUI])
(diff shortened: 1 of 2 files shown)


go/tasks/plugins/k8s/spark/spark.go
```diff
@@ -407,8 +407,9 @@ func getEventInfoForSpark(sj *sparkOp.SparkApplication) (*pluginsCore.TaskInfo,
 			})
 		}
 	} else if sj.Status.AppState.State == sparkOp.RunningState && sj.Status.DriverInfo.WebUIIngressAddress != "" {
-		// Append https as the operator doesn't currently.
-		customInfoMap[sparkDriverUI] = fmt.Sprintf("https://%s", sj.Status.DriverInfo.WebUIIngressAddress)
+		// Older versions of spark-operator does not have http:// but newer versions have. 
+		customInfoMap[sparkDriverUI] = strings.TrimPrefix(sj.Status.DriverInfo.WebUIIngressAddress, "http://")
```
kumare3, line 411:
```suggestion
		uri = strings.TrimPrefix(sj.Status.DriverInfo.WebUIIngressAddress, "http://")
```

go/tasks/plugins/k8s/spark/spark.go
```diff
@@ -407,8 +407,9 @@ func getEventInfoForSpark(sj *sparkOp.SparkApplication) (*pluginsCore.TaskInfo,
...
+		customInfoMap[sparkDriverUI] = fmt.Sprintf("https://%s", customInfoMap[sparkDriverUI])
```
kumare3, line 412:
```suggestion
		customInfoMap[sparkDriverUI] = fmt.Sprintf("https://%s", uri)
```

kumare3:
a small code change suggested, then looks good

pingsutw (approved):

go/tasks/plugins/k8s/spark/spark.go
```diff
@@ -407,8 +407,9 @@ func getEventInfoForSpark(sj *sparkOp.SparkApplication) (*pluginsCore.TaskInfo,
 			})
 		}
 	} else if sj.Status.AppState.State == sparkOp.RunningState && sj.Status.DriverInfo.WebUIIngressAddress != "" {
-		// Append https as the operator doesn't currently.
-		customInfoMap[sparkDriverUI] = fmt.Sprintf("https://%s", sj.Status.DriverInfo.WebUIIngressAddress)
+		// Older versions of spark-operator does not append http:// but newer versions do.
+		uri := strings.TrimPrefix(sj.Status.DriverInfo.WebUIIngressAddress, "http://")
+		customInfoMap[sparkDriverUI] = fmt.Sprintf("https://%s", uri)
```
hamersaw, line 412: To maintain backwards compatibility does it make sense to check for the prefix rather than removing it (noop if it doesn't exist) and then appending it? Maybe something like

```
uri := sj.Status.DriverInfo.WebUIIngressAddress
if !strings.HasPrefix("https://") {
    uri = fmt.Sprintf("https://%s", uri)
}
customInfoMap[sparkDriverUI] = uri
```

EngHabu, line 412: We shouldn't assume SSL all the time.
```
uri := sj.Status.DriverInfo.WebUIIngressAddress
if !strings.HasPrefix("https://") && !strings.HasPrefix("http://") {
    uri = fmt.Sprintf("https://%s", uri)
}

customInfoMap[sparkDriverUI] = uri
```

mucahitkantepe: I agree, so we should just pass the `WebUIIngressAddress` value along?

hamersaw: @mucahitkantepe yes, and then if it does not contain `https` or `http` we can prepend with `https` as Haytham suggests.

hamersaw:
work completed in https://github.com/flyteorg/flyteplugins/pull/389

PR #348, opened 2023-05-09
Allow 0 worker in pytorch plugins & Add objectMeta to PyTorchJob

ByronHsu:
# TL;DR
1. Kubeflow PyTorch can be configured to use 0 workers when running distributed PyTorch jobs. In this case, the training job would run on a single machine (the master node), without any additional worker nodes.
2. Pass objectMeta to PyTorchJob

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added


pingsutw (commented):
is there any benefit to use only one master node in pytorch CRD? people can just run pytorch in regular python task for single node training, right?

fg91:
> is there any benefit to use only one master node in pytorch CRD? people can just run pytorch in regular python task for single node training, right?

I can only imagine the env vars set by the operator like world size, rank, ...? In the torch elastic task we opted to run single worker trainings in a normal python task/single pod so that users don't need the training operator.

go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
```diff
@@ -115,23 +111,27 @@ func (pytorchOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx
...
+		}
 	}
 	job := &kubeflowv1.PyTorchJob{
 		TypeMeta: metav1.TypeMeta{
 			Kind:       kubeflowv1.PytorchJobKind,
 			APIVersion: kubeflowv1.SchemeGroupVersion.String(),
 		},
-		Spec: jobSpec,
+		Spec:       jobSpec,
+		ObjectMeta: *objectMeta,
```
hamersaw, line 134: Is this just to set labels / annotations on the CR the same as the replicas? If this is the route we want to go, this should probably be done for all kf operator plugins.

fg91:
> In our case, we start with 0 worker since it's easier to debug, and then adjust to multiple workers.
> 
> Although python task can achieve the same thing, we shouldn't error out 0 worker in PyTorch because it's what PyTorch operator allows

Fair point. Users can still switch to python task if they prefer by removing the `task_config=Pytorch(...`.

kumare3:
I actually want to get rid of a required dependency on PytorchOperator for simple single node training. which can suffice in many cases. This actually makes scaling really nice, you start with one node and simply scale to more nodes - to scale you may need to deploy the operator?

This is why when nnodes=1, we just change the task type itself. WDYT? @fg91  and @ByronHsu

kumare3:
check this out - https://github.com/flyteorg/flytekit/blob/e44b8027bd0f59c0f90eb5f86bfa181aebad0d74/plugins/flytekit-kf-pytorch/flytekitplugins/kfpytorch/task.py#L112

ByronHsu:
Could you elaborate the drawback to use PyTorch Operator to training a single node case?

kumare3:
@ByronHsu - FlytePropeller is way more optimal in allocating resources, retrying, and completing sooner. Also this for single node is faster, runs without needing an operator and does not need a CRD to be created.

ByronHsu:
@kumare3 skipping the CRD part can definitely be faster. Thanks. I will raise a corresponding pr in flytekit

ByronHsu:
Will merge with https://github.com/flyteorg/flyteplugins/pull/345 and do a integration test

fg91:
Is this still supposed to be pushed over the finish line or shell we close it?

PR #349, opened 2023-05-09
Enable pod template and Use copy to construct head/worker in ray plugin

ByronHsu:
## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Why?
1. The current implementation is unable to parse podtemplate
2. The current implementation will lose container information (like `Volume`) when constructing worker/head pod because it created a new container from scratch

What?
1.  Use `flytek8s.ToK8sPodSpec` to parse both container and pod type
2. Copy the container from the original primary container instead of creating a new one


hamersaw (commented):
So calling `ToK8sPodSpec` does a lot of things, like injecting Flyte-specific customizations. I think the original implementation decided to call `ToK8sContainer` directly to keep these out of the Pod definition. I just want to make sure that all of this additional Pod construction logic will not introduce regressions.

The alternative is to call `MergeWithBasePodTemplate` [similar to here](https://github.com/flyteorg/flyteplugins/blob/76a80ec5b7240d16fe5a1a190d51d8b784b88115/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L247-L251) which just merges the `PodSpec` with the default PodTemplate (if exists). This was purposefully refactored to ensure the default PodTemplate could be applied to any `PodSpec`.

hamersaw:
> So calling `ToK8sPodSpec` does a lot of things, like injecting Flyte-specific customizations. I think the original implementation decided to call `ToK8sContainer` directly to keep these out of the Pod definition. I just want to make sure that all of this additional Pod construction logic will not introduce regressions.
> 
> The alternative is to call `MergeWithBasePodTemplate` [similar to here](https://github.com/flyteorg/flyteplugins/blob/76a80ec5b7240d16fe5a1a190d51d8b784b88115/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L247-L251) which just merges the `PodSpec` with the default PodTemplate (if exists). This was purposefully refactored to ensure the default PodTemplate could be applied to any `PodSpec`.

cc @pingsutw thoughts?

bstadlbauer:
@hamersaw Sounds good to me 👍 Should I start right away or wait on this PR?

hamersaw:
> @hamersaw Sounds good to me +1 Should I start right away or wait on this PR?

@bstadlbauer it looks like the dask plugin will call the [MergeWithBasePodTemplate](https://github.com/flyteorg/flyteplugins/blob/76a80ec5b7240d16fe5a1a190d51d8b784b88115/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L247-L251) function directly because we're building the `PodSpec` manually with specific options (ex. [workerSpec](https://github.com/flyteorg/flyteplugins/blob/9a2bbbaf2f3ac9e38222ba7755ac4bda1ac2fdb7/go/tasks/plugins/k8s/dask/dask.go#L192-L210).

Just trying to keep some semblance of consistency across the Flyte maintained plugins. Very interested in hearing thoughts from all involved!

go/tasks/plugins/k8s/ray/ray.go
```diff
@@ -55,9 +56,24 @@ func (rayJobResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsC
...
+	var found bool = false
+	for _, c := range podSpec.Containers {
+		if c.Name == primaryContainerName {
+			container = c
+			found = true
+			break
+		}
+	}
+
+	if found == false {
```
pingsutw, line 75:
```suggestion
	if !found {
```

go/tasks/plugins/k8s/ray/ray.go
```diff
@@ -55,9 +56,24 @@ func (rayJobResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsC
...
-	container, err := flytek8s.ToK8sContainer(ctx, taskCtx)
+	podSpec, objectMeta, primaryContainerName, err := flytek8s.ToK8sPodSpec(ctx, taskCtx)
+
 	if err != nil {
-		return nil, errors.Errorf(errors.BadTaskSpecification, "Unable to create container spec: [%v]", err.Error())
+		return nil, flyteerr.Errorf(flyteerr.BadTaskSpecification, "Unable to create pod spec: [%v]", err.Error())
+	}
+
+	var container v1.Container
+	var found bool = false
```
pingsutw, line 66:
```suggestion
	var container *v1.Container
	found := false
```

pingsutw:
The reason we use `ToK8sContainer` is because flyte didn't support podtemplate while we were working on it.
I think it's fine to use `ToK8sPodSpec`. there are a lot of same logic between `ToK8sContainer` and `ToK8sPodSpec`

- Update cmd and resource. [ToK8sContainer](https://github.com/flyteorg/flyteplugins/blob/1e0faabcc0980c56b6679dafc00abac6783737a5/go/tasks/pluginmachinery/flytek8s/container_helper.go#L253-L262) vs [ToK8sPodSpec](https://github.com/flyteorg/flyteplugins/blob/84030bd81d5434efdbb50667c7fb29dda8a8317a/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L192-L215)
- Copilot is only enabled in rawContainer task, so this [code](https://github.com/flyteorg/flyteplugins/blob/84030bd81d5434efdbb50667c7fb29dda8a8317a/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L222-L239) is fine
- Update security context. [ToK8sContainer](https://github.com/flyteorg/flyteplugins/blob/1e0faabcc0980c56b6679dafc00abac6783737a5/go/tasks/pluginmachinery/flytek8s/container_helper.go#L248-L250) and [ToK8sPodSpec](https://github.com/flyteorg/flyteplugins/blob/84030bd81d5434efdbb50667c7fb29dda8a8317a/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L243-L246)
- [Merge](https://github.com/flyteorg/flyteplugins/blob/84030bd81d5434efdbb50667c7fb29dda8a8317a/go/tasks/pluginmachinery/flytek8s/pod_helper.go#L241-L251) default pod template into podSpec. 

I think all the k8s should use `ToK8sPodSpec` now because we need to merge flytekit podTemplate and default podTemplate into the podSpec. btw, kubeflow plugin already used `ToK8sPodSpec`. @hamersaw wdyt

PR #355, opened 2023-06-02
[Bigquery] Add support for impersonation of GSA bound to task's KSA

jeevb:


go/tasks/pluginmachinery/google/gke_task_workload_identity_token_source_factory.go
```diff
@@ -0,0 +1,92 @@
...
+	if err != nil {
+		return oauth.TokenSource{}, err
+	}
+	return impersonate.CredentialsTokenSource(ctx, impersonate.CredentialsConfig{
+		TargetPrincipal: gcpServiceAccount,
+		Scopes:          impersonationScopes,
+	})
+}
+
+func getKubeClient() (*kubernetes.Clientset, error) {
```
jeevb, line 74: This is hacked together to only support in-cluster for now. Ideally there is a global kube client we can just use instead of creating plugin-specific ones.

ctso (approved):
This all looks pretty reasonable to me.  We should make sure we update the docs to call out this new `gke-task-workload-identity` token source.  We should be sure to call out that the KSA that (I think...) `flytepropeller` uses to make sure it has enough permissions to read ServiceAccount objects from the k8s API.

eapolinario (approved):
Awesome.

PR #359, opened 2023-06-12
fix v1 pytorch job plugin with elastic policy

yubofredwang:
# TL;DR
There was a bug that introduced in PR #345. The pytorch v2 task template is handled incorrectly.
Added Elastic Config parsing for v1. It follows the exact same handling logic as v0 since they are kept as same as v0.

Follow-up PR in flytekit: PR [#1690](https://github.com/flyteorg/flytekit/pull/1690)

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

### DIFF
--- a/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
+++ b/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
@@ -57,12 +57,6 @@ func (pytorchOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx
 		return nil, flyteerr.Errorf(flyteerr.BadTaskSpecification, "nil task specification")
 	}
 
-	pytorchTaskExtraArgs := plugins.DistributedPyTorchTrainingTask{}
-	err = utils.UnmarshalStruct(taskTemplate.GetCustom(), &pytorchTaskExtraArgs)
-	if err != nil {
-		return nil, flyteerr.Errorf(flyteerr.BadTaskSpecification, "invalid TaskSpecification [%v], Err: [%v]", taskTemplate.GetCustom(), err.Error())
-	}
-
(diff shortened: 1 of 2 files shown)


hamersaw (approved):
So the elastic policy for task type version 1 was never being applied because we were only looking at the elastic config from version 0?

The configs for version 0 and 1 have the same functions right? Can we just wrap this in an interface and remove a lot of the boilerplate code? Something like:
```
type elasticConfig interface {
    func GetMinReplicas() uint32
    func GetMaxReplicas() uint32
    // ...
}

var elasticConfig elasticConfig
if taskTemplate.TaskTypeVersion == 0 {
    // ...
    elasticConfig = pytorchTaskExtraArgs.GetElasticConfig()
else {
    // ...
    elasticConfig = kfPytorchTaskExtraArgs.GetElasticConfig()
}

if elasticConfig != nil {
    // leave as is
}
```

yubofredwang:
> So the elastic policy for task type version 1 was never being applied because we were only looking at the elastic config from version 0?
> 
> The configs for version 0 and 1 have the same functions right? Can we just wrap this in an interface and remove a lot of the boilerplate code? Something like:
> 
> ```
> type elasticConfig interface {
>     func GetMinReplicas() uint32
>     func GetMaxReplicas() uint32
>     // ...
> }
> 
> var elasticConfig elasticConfig
> if taskTemplate.TaskTypeVersion == 0 {
>     // ...
>     elasticConfig = pytorchTaskExtraArgs.GetElasticConfig()
> else {
>     // ...
>     elasticConfig = kfPytorchTaskExtraArgs.GetElasticConfig()
> }
> 
> if elasticConfig != nil {
>     // leave as is
> }
> ```

yes, it was not handled initially. very good suggestion. I updated the code! thanks

hamersaw (approved):

PR #360, opened 2023-06-13
Feat: Add pod start and finish time in RFC3339 time format to logging link templating variables #minor

fg91:
# TL;DR

Currently, the the pod start and finish time in unix time format (seconds since 1970) can be used to template logging links.
This time format cannot be used in combination with google cloud logging. Instead, GCP logging [requires timestamps in RFC3339 format](https://cloud.google.com/logging/docs/view/logging-query-language#search_by_time):

```
timestamp >= "2016-11-29T23:00:00Z"
timestamp <= "2016-11-29T23:30:00Z"
```

This PR adds the pod start and finish time in this format to the variables available for templating logging links.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added
 - [x] Code documentation added: https://github.com/flyteorg/flytesnacks/pull/998

## Complete description
__NA__

## Tracking Issue
__NA__

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/logs/logging_utils.go
+++ b/go/tasks/logs/logging_utils.go
@@ -38,16 +38,21 @@ func GetLogsForContainerInPod(ctx context.Context, logPlugin tasklog.Plugin, pod
 		return nil, nil
 	}
 
+	startTime := pod.CreationTimestamp.Unix()
+	finishTime := time.Now().Unix()
+
 	logs, err := logPlugin.GetTaskLogs(
 		tasklog.Input{
-			PodName:           pod.Name,
-			PodUID:            string(pod.GetUID()),
-			Namespace:         pod.Namespace,
-			ContainerName:     pod.Spec.Containers[index].Name,
-			ContainerID:       pod.Status.ContainerStatuses[index].ContainerID,
-			LogName:           nameSuffix,
-			PodUnixStartTime:  pod.CreationTimestamp.Unix(),
-			PodUnixFinishTime: time.Now().Unix(),
+			PodName:              pod.Name,
+			PodUID:               string(pod.GetUID()),
+			Namespace:            pod.Namespace,
+			ContainerName:        pod.Spec.Containers[index].Name,
+			ContainerID:          pod.Status.ContainerStatuses[index].ContainerID,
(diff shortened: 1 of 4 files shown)


hamersaw (commented):
Fantastic, thanks! Should be pretty easy to add ISO 8601 format as well right? Would wrap up [this issue](https://github.com/flyteorg/flyte/issues/3348).

fg91:
> Fantastic! Should be pretty easy to add ISO 8601 format as well right? Would wrap up [this issue](https://github.com/flyteorg/flyte/issues/3348).

Unfortunately, go appears to not have the `time.ISO8601` equivalent to `time.RFC33339`, see [here](https://pkg.go.dev/time#Time.Format).

Users on stackoverflow recommend to use rfc3339 as a proxy for 8601 (see reply to [this](https://stackoverflow.com/a/42217963) answer):

> From what I could tell, RFC3339 is a stricter version of ISO 8601. So it is probably safe to use the RFC format if a system expects the ISO. 

However, [this](https://stackoverflow.com/a/65221179) answer suggests that not everything that is allowed by RFC3339 is allowed by ISO 8601 and vice versa.

That being said, a RFC3339 timestamp propeller generates based on this PR is e.g. `2023-06-13T18:28:16Z`. According to [this](https://www.regextester.com/97766) ISO 8601 timestamp format validator (first google result), it is a match.

So it seems that `time.Unix(startTime, 0).Format(time.RFC3339)` doesn't use such peculiarities?!


We could use our own ISO8601 format string (see [here](https://stackoverflow.com/a/35479255)) even though I couldn't confirm this format string is correct.

Alternatively, we could leave it with RFC3339 and in the documentation explain that it is valid also for ISO 8601. This would be my suggestion. What do you think @hamersaw ?

hamersaw:
> Alternatively, we could leave it with RFC3339 and in the documentation explain that it is valid also for ISO 8601. This would be my suggestion.

Sounds great to me! Thanks for the depth.

hamersaw (approved):

PR #363, opened 2023-06-16
WIP: Show potential soluation

bstadlbauer:
# TL;DR
Fixes an issue

Checkout https://github.com/flyteorg/flyte/issues/3787 for the full issue description.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [ ] Code completed
 - [ ] Smoke tested
 - [ ] Unit tests added

## Complete description
At the moment, this adds the hashed inputs as suffix to the key. The hashed inputs implementation has been copied over from `generateArtifactTagName` ([here](https://github.com/flyteorg/flytepropeller/blob/89dfd5b3bb694d9a19108aa4c8a6d10a18fd3a32/pkg/controller/nodes/task/catalog/datacatalog/transformer.go#L164)) so in case we stick with this solution, we should probably refactor out into `flyteplugins`

## Tracking Issue
fixes Checkout https://github.com/flyteorg/flyte/issues/3787 for the full issue description


hamersaw:
@bstadlbauer thanks so much for finding this issue, and more for submitting a fix! I think this will work, two things:
(1) lets add some unit tests just to make sure subsequent submissions with different input values result in different workItemIDs.
(2) looks like the hashify function was taken from propeller catalog. not sure if we want to go through the hassle of refactoring this to live entirely in plugins (ie. move function here and call from propeller) rather than duplicating it. not sure how long until this code will be deprecated with the looming ArrayNode work.

bstadlbauer:
@hamersaw I've added a test to check that work item ids are different and refactored the literal map hashing out of propeller into here. The testing for the hash function has been copied and adapted from [here](https://github.com/flyteorg/flytepropeller/blob/4d1a3713d1daae3e1bddf9922e92c6ccc9d56f13/pkg/controller/nodes/task/catalog/datacatalog/transformer_test.go).

go/tasks/pluginmachinery/catalog/async_client_impl.go
```diff
@@ -14,11 +14,14 @@ import (
...
 
 	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/workqueue"
+
+	"github.com/flyteorg/flyteidl/gen/pb-go/flyteidl/core"
 )
 
 const specialEncoderKey = "abcdefghijklmnopqrstuvwxyz123456"
 
 var base32Encoder = base32.NewEncoding(specialEncoderKey).WithPadding(base32.NoPadding)
+var emptyLiteralMap = core.LiteralMap{Literals: map[string]*core.Literal{}}
```
hamersaw, line 24: should this be defined in the `hashing.go` file instead? I believe it's the only place it's being used right?

hamersaw (commented):
My nit all over flyte repos is to alphabetize import statements. I've seen place where the same package is imported under two different names, this would help mitigate these scenarios. Do you mind updating the added imports?

go/tasks/pluginmachinery/catalog/async_client_impl.go
bstadlbauer: Thanks for spotting - moved it over 👍 30ca1642ce7fe834781ebad7916bed5d90a84ecf

bstadlbauer:
@hamersaw Thanks for reviewing!

Not being familiar with best practices in `go`, I've read up a bit on it, so let me know if the following conflicts with the default within Flyte.

My main information source has been the [google style guide](https://google.github.io/styleguide/go/best-practices#import-ordering).

1. Imports are organized in blocks separated by newlines
2. The first block should be imports from stdlib 
3. The second block should be all other imports

The `goimports` tool only understands those two sections (unless started with `-local` then it could do a third block with local package imports)

Given the above, it seems like as long as there are only two blocks separated by one space, imports should be ordered automatically (or otherwise fail CI). 

Does that sound about right? I've not run things assuming there `-local` flag

hamersaw:
@bstadlbauer re: imports - I try to stick with exactly as you have described.

> Given the above, it seems like as long as there are only two blocks separated by one space, imports should be ordered automatically (or otherwise fail CI).
Exactly.

>  it seems like it's a bit of a wild west out there
And it drives me crazy! It's not a high priority item to clean all of this up (especially across repos), which is why it hasn't been done yet. I just try to keep it from getting worse.

hamersaw (approved):

PR #368, opened 2023-07-03
Support gRPC config for agent-service plugin

honnix:
# TL;DR
Support customized gRPC config for agent-service plugin.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [x] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

Support configuring gRPC endpoint with:

* insecure
* default service config
* timeouts

Details can be found in https://github.com/flyteorg/flyte/issues/3823

**Note that this is a breaking change!** I'm not sure how mature the `agent-service` is and how many users are using it, but if backward compatibility is important in this case, I can go an extra mile to achieve that.

## Tracking Issue
Closes https://github.com/flyteorg/flyte/issues/3823

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/plugins/webapi/agent/config.go
+++ b/go/tasks/plugins/webapi/agent/config.go
@@ -39,7 +39,7 @@ var (
 				Value: 50,
 			},
 		},
-		DefaultGrpcEndpoint: GrpcEndpoint{
+		DefaultAgent: Agent{
 			Endpoint:       "dns:///flyte-agent.flyte.svc.cluster.local:80",
 			Insecure:       true,
 			DefaultTimeout: config.Duration{Duration: 10 * time.Second},
@@ -58,21 +58,21 @@ type Config struct {
 	// ResourceConstraints defines resource constraints on how many executions to be created per project/overall at any given time
 	ResourceConstraints core.ResourceConstraintsSpec `json:"resourceConstraints" pflag:"-,Defines resource constraints on how many executions to be created per project/overall at any given time."`
 
-	// The default grpc endpoint if there does not exist a more specific matching against task types
-	DefaultGrpcEndpoint GrpcEndpoint `json:"defaultGrpcEndpoint" pflag:",The default grpc endpoint of agent service."`
+	// The default agent if there does not exist a more specific matching against task types
+	DefaultAgent Agent `json:"defaultAgent" pflag:",The default agent."`
 
-	// The grpc endpoints of agent services, which are used to match against specific task types
-	GrpcEndpoints map[string]*GrpcEndpoint `json:"grpcEndpoints" pflag:",The grpc endpoints of agent services."`
+	// The agents used to match against specific task types. {AgentId: Agent}
+	Agents map[string]*Agent `json:"agents" pflag:",The agents."`
 
-	// Maps endpoint to their plugin handler. {TaskType: Endpoint}
-	EndpointForTaskTypes map[string]string `json:"endpointForTaskTypes" pflag:"-,"`
+	// Maps task types to their agents. {TaskType: AgentId}
+	AgentForTaskTypes map[string]string `json:"agentForTaskTypes" pflag:"-,"`
 
 	// SupportedTaskTypes is a list of task types that are supported by this plugin.
 	SupportedTaskTypes []string `json:"supportedTaskTypes" pflag:"-,Defines a list of task types that are supported by this plugin."`
 }
 
-type GrpcEndpoint struct {
-	// Endpoint points to a gRPC service
+type Agent struct {
+	// Endpoint points to an agent gRPC endpoint
 	Endpoint string `json:"endpoint"`
 
 	// Insecure indicates whether the communication with the gRPC service is insecure
--- a/go/tasks/plugins/webapi/agent/config.go
+++ b/go/tasks/plugins/webapi/agent/config.go
@@ -10,8 +10,6 @@ import (
 )
 
 var (
-	defaultTimeout = config.Duration{Duration: 10 * time.Second}
-
 	defaultConfig = Config{
 		WebAPI: webapi.PluginConfig{
 			ResourceQuotas: map[core.ResourceNamespace]int{
@@ -42,8 +40,9 @@ var (
 			},
 		},
 		DefaultGrpcEndpoint: GrpcEndpoint{
-			Endpoint: "dns:///flyte-agent.flyte.svc.cluster.local:80",
-			Insecure: true,
+			Endpoint:       "dns:///flyte-agent.flyte.svc.cluster.local:80",
+			Insecure:       true,
+			DefaultTimeout: config.Duration{Duration: 10 * time.Second},
 		},
 		SupportedTaskTypes: []string{"task_type_1", "task_type_2"},
 	}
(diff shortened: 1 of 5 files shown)


go/tasks/plugins/webapi/agent/plugin.go
```diff
@@ -144,28 +157,44 @@ func (p Plugin) Status(ctx context.Context, taskCtx webapi.StatusContext) (phase
...
+		pool, err := x509.SystemCertPool()
+		if err != nil {
+			return nil, err
+		}
+
+		creds := credentials.NewClientTLSFromCert(pool, "")
+		opts = append(opts, grpc.WithTransportCredentials(creds))
+	}
+
+	if endpoint.DefaultServiceConfig != "" {
```
hamersaw, line 188: I think `len(endpoint.DefaultServiceConfig) > 0` is the idiomatic golang way to do this.

go/tasks/plugins/webapi/agent/config.go
```diff
@@ -10,6 +10,8 @@ import (
 )
 
 var (
+	defaultTimeout = config.Duration{Duration: 10 * time.Second}
```
hamersaw, line 13: do we want to make this configurable?

hamersaw (commented):
@honnix can you elaborate on what is not backwards compatible?

go/tasks/plugins/webapi/agent/config.go
honnix: Good question. I thought about how to deal with the default here. My reasoning is that if user bothers configuring this, they might just configure for each operation instead because it doesn't save much effort. That being said, I don't have a strong opinion and I can make it configurable for sure.

go/tasks/plugins/webapi/agent/plugin.go
honnix: Sure. I just followed existing code I found.

honnix:
> @honnix can you elaborate on what is not backwards compatible?

Both `EndpointForTaskTypes` and `DefaultGrpcEndpoint` are now completely different types so if a user has configured either of those, upgrading to this version will crash their application I guess.

go/tasks/plugins/webapi/agent/config.go
```diff
@@ -78,8 +77,11 @@ type GrpcEndpoint struct {
 	// DefaultServiceConfig sets default gRPC service config; check https://github.com/grpc/grpc/blob/master/doc/service_config.md for more details
 	DefaultServiceConfig string `json:"defaultServiceConfig"`
 
-	// Timeouts defines various RPC timeout values for different plugin operations: CreateTask, GetTask, DeleteTask; if not configured, defaults to 10s
+	// Timeouts defines various RPC timeout values for different plugin operations: CreateTask, GetTask, DeleteTask; if not configured, defaults to DefaultTimeout
 	Timeouts map[string]config.Duration `json:"timeouts"`
+
+	// DefaultTimeout gives the default RPC timeout if a more specific one is not defined in Timeouts
+	DefaultTimeout config.Duration `json:"defaultTimeout"`
```
honnix, line 84: Note that this is now a mandatory config item, otherwise it defaults to `0` I guess, which would fail the RPC immediately.

honnix: This doesn't seem to be nice so I addressed it in 2a706c3e85d7eac4ff4d5214f825f7813cbdb60a

hamersaw:
> upgrading to this version will crash their application I guess

@pingsutw we released flyte agents as an experimental feature right? so no guarantees of backwards compatibility. IMO fixing this grpc configuration issue warrants a backwards incompatible merge. thoughts?

eapolinario:
> > upgrading to this version will crash their application I guess
> 
> @pingsutw we released flyte agents as an experimental feature right? so no guarantees of backwards compatibility. IMO fixing this grpc configuration issue warrants a backwards incompatible merge. thoughts?

Agreed. This feature is tagged as experimental, so let's use this time to break backwards compatibility if needed (which seems like this particular instance is one of such times).

hamersaw:
@honnix thinking about this it will be unable to cover the scenario where there are two flyte agent deployments where each satisfies multiple task types right? I'm thinking something like:
```
grpcConnections:
    - endpoint: x.com:8088
      taskTypes:
      - foo
      - bar
    - endpoint: y.com:8088
      taskTypes:
      - baz
      - bat
```
the current approach would create a separate gRPC connection for each task type right? whereas it would probably be better to reuse a single connection for each endpoint. Thoughts?

honnix:
I don't think so. The task type here is only for routing and there should be only one grpc connection per each endpoint. This PR shouldn't change how that part works I think.

hamersaw:
> I don't think so. The task type here is only for routing and there should be only one grpc connection per each endpoint. This PR shouldn't change how that part works I think.

So the [EndpointForTaskTypes map](https://github.com/flyteorg/flyteplugins/blob/3d2525a9f9373a220219fb2d1c07f0b38101b23b/go/tasks/plugins/webapi/agent/config.go#L60) is keyed on the task type, but the [connectionCache](https://github.com/flyteorg/flyteplugins/blob/3d2525a9f9373a220219fb2d1c07f0b38101b23b/go/tasks/plugins/webapi/agent/plugin.go#L30) is keyed on the endpoint. So you can't have two endpoints configured differently for different task types. For example, differing timeouts will result in unintended behavior:
```
endpointForTaskTypes:
  foo: 
    endpoint: localhost:8088
    timeout: 20
  bar: 
    endpoint: localhost:8088
    timeout: 40
```
IIUC the above would only create a single gRPC connection in the background (with timeout of whatever agent was called first) and reuse it for calls to both task types. I don't think this is a huge issue, but I don't like the unintended behavior for future added gRPC params that might be more important.

honnix:
@hamersaw I did some experiment in https://github.com/flyteorg/flyteplugins/pull/368/commits/d67a01563a504ab1cf33940056e1ca43b3ed3c38 and now the config is like:

```yaml
endpointForTaskTypes:
  foo: agent
  bar: bar_agent
  foo_bar: agent

grpcEndpoints:
  agent:
    endpoint: localhost:8088
    timeouts:
      CreateTask: 1s
  bar_agent:
    endpoint: localhost:8088
    timeouts:
      CreateTask: 2s
```

And the cache is keyed on the whole endpoint config (as a pointer).

honnix:
> Does it make sense to update the naming here? Ex `grpcEndpoints` -> `agentEndpoints` and `endpointForTaskTypes` -> `agentsForTaskTypes` or `taskTypeAgents`?

SGTM but maybe it's better double check with the original author of the config section.

> Also, just want to make sure this doesn't introduce additional confusion. We still need to add all supported plugins to the `supportedTaskTypes` configuration right?

Yes that is correct. This PR does not make any change to the semantics of `supportedTaskTypes`.

honnix:
> Does it make sense to update the naming here? Ex `grpcEndpoints` -> `agentEndpoints` and `endpointForTaskTypes` -> `agentsForTaskTypes` or `taskTypeAgents`?
> 
> Also, just want to make sure this doesn't introduce additional confusion. We still need to add all supported plugins to the `supportedTaskTypes` configuration right?

@hamersaw I did some experiments in the latest commit. PTAL, thanks!

hamersaw (approved):

pingsutw (approved):
@honnix Thank you so much

honnix:
Thank you all for reviewing and providing great comments! No worries about nits at all. If we are gonna ship a breaking change, it's better we do it right.

pingsutw:
@honnix btw, would you mind helping us update the [doc](https://github.com/flyteorg/flytesnacks/blob/5c5380d28cec7ae57d646b7e0cd6bce195c1df5a/examples/extend_flyte/extend_flyte/agent_service.py#L171-L186) here.

honnix:
> @honnix btw, would you mind helping us update the [doc](https://github.com/flyteorg/flytesnacks/blob/5c5380d28cec7ae57d646b7e0cd6bce195c1df5a/examples/extend_flyte/extend_flyte/agent_service.py#L171-L186) here.

Sure. I did not know this. I can do it tomorrow.

PR #372, opened 2023-07-13
Add support for using task execution ID fields in log URI templates

jeevb:


go/tasks/pluginmachinery/tasklog/template.go
```diff
@@ -96,67 +106,28 @@ func (s TemplateLogPlugin) GetTaskLog(podName, podUID, namespace, containerName,
...
+	}
+	err = templateVars.Merge(extraTemplateVars...)
+	if err != nil {
+		return Output{}, err
 	}
 
 	taskLogs := make([]*core.TaskLog, 0, len(s.templateUris))
 	for _, templateURI := range s.templateUris {
+		var buf bytes.Buffer
+		t := template.Must(template.New("uri").Parse(templateURI))
```
jeevb, line 123: Switches to using text templating instead of regexes+replace.

EngHabu: I believe you will find a benchmarking test in one of the test files that's comparing the two and finding a big discrepancy in performance (memory allocations?)... hence the decision... 
Things may have changed and we should run it again... 
Also we may make this as a flag on the log config... the templatingEngine: Default/RegEx, Gotemplate ?

jeevb: Hmm interesting. Didn’t know about this history. The regex implementation is somewhat limiting, but we can work around that I suppose. I’m fine with keeping it as-is if the difference is significant - not sure about adding more configuration.

pmahindrakar-oss: Since we are revamping this logic, can we introduce a way of flagging to switch over to the old way incase of backward compat issues. Haytham's suggestion to use a log config can allow to keep the old and new behavior and we can deprecate the old config and use the new method as default in new versions, but users can switch over in case they find issues and if things look ok, then we remove the deprecated config and method in next releases

jeevb: For performance reasons, I reverted to using regexes+replace.

go/tasks/pluginmachinery/tasklog/template.go
```diff
@@ -1,14 +1,73 @@
...
+		PodName                 string                        `json:"podName,omitempty"`
+		Namespace               string                        `json:"namespace,omitempty"`
+		ContainerName           string                        `json:"containerName,omitempty"`
+		ContainerID             string                        `json:"containerId,omitempty"`
+		LogName                 string                        `json:"logName,omitempty"`
+		PodRFC3339StartTime     string                        `json:"podRFC3339StartTime,omitempty"`
+		PodRFC3339FinishTime    string                        `json:"podRFC3339FinishTime,omitempty"`
+		PodUnixStartTime        string                        `json:"podUnixStartTime,omitempty"`
+		PodUnixFinishTime       string                        `json:"podUnixFinishTime,omitempty"`
+		PodUID                  string                        `json:"podUID,omitempty"`
```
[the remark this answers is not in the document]
jeevb: 🤔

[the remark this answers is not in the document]
jeevb: Is the intent on improving UX or performance? I think using 2 distinct templating schemes would add complexity, but there will be a performance benefit to cutting down the number of variables used for rendering.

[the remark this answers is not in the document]
pmahindrakar-oss: IMO would make it cleaner, structured and also have performance benefits of doing so as it wont have extraneous vars which are not applicable to a particular templae.
Also would make it easier for adding more such templating schemes in the future. Not a blocker for this PR though just a nice to have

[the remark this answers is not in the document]
jeevb: Per offline discussion, we will split into 2 templating schemes to keep rendering performant.

pmahindrakar-oss (approved):

fg91:
Nice, thanks :)
Could you please document these new variables [here](https://docs.flyte.org/projects/cookbook/en/stable/auto_examples/deployment/configure_logging_links.html#how-to-configure)?

PR #374, opened 2023-07-14
feat: Dask add pod template support

bstadlbauer:
# TL;DR
This PR change the `dask` plugin to use `ToK8sPodSpec` to create the basis of the pod template to be used for all `dask` components. This is coherent with other plugins (such as the `ray` plugin) and enables the use of common features such as pod templates.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
This has been pretty straight forward. The only downside was that I had to remove the interruptible config from the pod spec after it has been added, as I couldn't see a way of not adding it in `ToK8sPodSpec`.


bstadlbauer:
@hamersaw I'm done with the work on this, but haven't smoke tested this in a local cluster yet, should hopefully get to that tomorrow.

The test and lint failures are a bit puzzling to me as both pass locally, and it seems like they flag something that's not even in the code (e.g. `dask_test:543` does not contain the function mentioned)

go/tasks/plugins/k8s/dask/dask.go
```diff
@@ -114,50 +118,66 @@ func (p daskResourceHandler) BuildResource(ctx context.Context, taskCtx pluginsC
...
 		return nil, err
 	}
-	schedulerSpec, err := createSchedulerSpec(*daskJob.Scheduler, clusterName, *defaults)
+	nonInterruptiblePodSpec := podSpec.DeepCopy()
+	removeInterruptibleConfig(nonInterruptiblePodSpec, taskCtx)
+
+	// Add labels and annotations to objectMeta as they're not added by ToK8sPodSpec
+	mergeMapInto(taskCtx.TaskExecutionMetadata().GetAnnotations(), objectMeta.Annotations)
+	mergeMapInto(taskCtx.TaskExecutionMetadata().GetLabels(), objectMeta.Labels)
+
```
bstadlbauer, line 138: @hamersaw Is there a better way to do this? The `ObjectMeta` object which is returned does not have those set

go/tasks/plugins/k8s/dask/dask.go
```diff
@@ -30,68 +30,72 @@ const (
...
+		matchExpressions := nst.MatchExpressions
+		newMatchExpressions := []v1.NodeSelectorRequirement{}
+		for _, matchExpression := range matchExpressions {
+			if !nodeSelectorRequirementsAreEqual(matchExpression, *interruptibleNodeSelectorRequirements) {
+				newMatchExpressions = append(newMatchExpressions, matchExpression)
+			}
+		}
+		nst.MatchExpressions = newMatchExpressions
 	}
+}
```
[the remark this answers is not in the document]
hamersaw: An alternative solution would be to create a wrapper over the `TaskExecutionContext` / `TaskExecutionMetadata` which forces`interruptible` to be false. This would stop them from ever being injected without adding more parameters. Something like:
```
struct daskTaskExecutionContext {
    taskExecutionContext
    metadata TaskExecutionMetadata
}

func (d *daskTaskExecutionContext) GetMetadata() TaskExecutionMetadata{
    return metadata
}

type daskTaskExecutionMetadata {
    taskExecutionMetadata
}

func (d *daskTaskExecutionMetadata) IsInterruptible() bool {
    return falsee
}
```
the function and type names are all wrong (don't know off the top of my head), but basically just forcing interruptible to always be false for dask tasks.

go/tasks/plugins/k8s/dask/dask.go
hamersaw: So on the top-levle object these are [set in flytepropeller](https://github.com/flyteorg/flytepropeller/blob/4322012ef07dff4de457473a7b5fb4cd6be4a2ee/pkg/controller/nodes/task/k8s/plugin_manager.go#L106-L107). Obviously this doesn't match the use-case here. The only alternative would be to take them out of flytepropeller, but then they would have to be added everywhere for each plugin and I'm sure it would break some users private plugins. I'm OK leaving as is if you are?

hamersaw:
>  should hopefully get to that tomorrow.

@bstadlbauer have you had a chance to test this locally yet? Everything looks good to me, but always nice to double-check.

As far as the CI checks, I'm not quite sure what's going on. Will plan on addressing as we iterate on this.

go/tasks/plugins/k8s/dask/dask.go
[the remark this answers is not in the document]
bstadlbauer: That sounds good to me - I'll try to implement that!

bstadlbauer:
@hamersaw This would be ready from my end now - again sorry for the incredibly long delay here.

I've swapped the `removeInterruptibleConfig` function in favor of a `nonInterruptibleTaskExecutionMetadata` wrapper as you've [suggested above](https://github.com/flyteorg/flyteplugins/pull/374#discussion_r1268068084).

In addition, I've written a small end-to-end test suite in https://github.com/bstadlbauer/flyte-dev-setup which can automatically end-to-end test the plugin which should hopefully decrease the iteration time on future changes as I was mostly blocked by testing this. I've also found a small bug (an incorrect restart policy due to a dask kubernetes update) which I've fixed along the way.
In addition I've noticed that Labels are somehow not propagated (i.e. [this task](https://github.com/bstadlbauer/flyte-dev-setup/blob/3e581a26685a59faaa452e03b95c7b2026687275/flyte_dev_setup/workflows.py#L68-L114) will fail), but I'm not 100% sure yet what the root cause is - I've done some debugging and it somehow seems like 

https://github.com/flyteorg/flyteplugins/blob/fba2d0b4d829c9822bbe0f6015d9920a36aa7f28/go/tasks/plugins/k8s/dask/dask.go#L116

the output `GetLabels()` here does not contain the label, which would mean it's outside of the plugin.

go/tasks/plugins/k8s/dask/dask.go
bstadlbauer: I'm fine with leaving it 👍

hamersaw (approved):

PR #378, opened 2023-07-15
Carry over hash value for all literal types in remote caching

nicholasjng:
# TL;DR
Previously, the hash on an input literal would be ignored if the literal was of either the `Collection` or `Map` type. This is fixed here by moving the hash-checking branch up in front of the special type logic, so that an empty literal with hash is constructed in either case if a hash is found.

Adds a regression test to verify the hashes are carried over into the representer literals for both map and collection types.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
See the discussion in https://github.com/flyteorg/flytekit/pull/1720 for the problem statement. This PR is a direct follow-up.

The remote cache fix was missing from that PR, and that is exactly what is added here. The code here is the same fix approach, just in Go this time.

## Tracking Issue
N/A

### DIFF
--- a/go/tasks/pluginmachinery/catalog/hashing_test.go
+++ b/go/tasks/pluginmachinery/catalog/hashing_test.go
@@ -550,6 +550,64 @@ func TestHashLiteralMap_LiteralsWithHashSet(t *testing.T) {
 				},
 			},
 		},
+		{
+			name: "literal map containing hash",
+			literal: &core.Literal{
+				Value: &core.Literal_Map{
+					Map: &core.LiteralMap{
+						Literals: map[string]*core.Literal{
+							"hello": {
+								Value: &core.Literal_Scalar{
+									Scalar: &core.Scalar{
+										Value: &core.Scalar_Primitive{
+											Primitive: &core.Primitive{
+												Value: &core.Primitive_StringValue{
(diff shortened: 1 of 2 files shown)


go/tasks/pluginmachinery/catalog/hashing_test.go
```diff
@@ -593,3 +593,33 @@ func TestNoInputValues(t *testing.T) {
 	assert.Equal(t, "GKw-c0PwFokMUQ6T-TUmEWnZ4_VlQ2Qpgw-vCTT0-OQ", hashDupe)
 	assert.Equal(t, hashDupe, hash)
 }
+
+// Ensure that hash values on maps and collections are carried over
+// into representing literals.
+func TestHashCarryOnContainers(t *testing.T) {
```
eapolinario, line 599: can you move these tests to the table-tests in `TestHashLiteralMap_LiteralsWithHashSet` instead?

nicholasjng: Done in f4d830ee7b44b30399082eaaadb069e740fd2ebb.

PR #379, opened 2023-07-19
Fix deletion of elastic task resource requests

fg91:
# TL;DR

Currently, the following elastic task's pods **will not have any resource requests (1.8 release)**:

```python
@task(
    task_config=Elastic(
        nnodes=2,
        nproc_per_node=2,
    ),
    requests=Resources(cpu="1", mem="1Gi", gpu="1"),
)
def train():
    ...
```

This PR fixes this.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description

The kubeflow plugin was recently refactored in flytekit (https://github.com/flyteorg/flytekit/pull/1627) as well as in flyteplugins (https://github.com/flyteorg/flyteplugins/pull/345).

After the refactoring, users can set e.g. different resources for master and worker replicas of `PyTorch` tasks:

```python
@task(
    task_config=PyTorch(
        worker=Worker(
            requests=Resources(cpu="2", mem="2Gi"),
            limits=Resources(cpu="4", mem="2Gi"),
        ),
        master=...
      )
      requests=Resources(cpu="1", mem="1Gi", gpu="1"),
)
....
```

If the user overrides one of the resource requests in the `task_config`, in the backend we override the resources configured on the task level [here](https://github.com/flyteorg/flyteplugins/blob/13ef7d6d6a178b06dffd592c4b49f46b9cda4410/go/tasks/plugins/k8s/kfoperators/common/common_operator.go#L260).

Now, in a `@task(task_config=PyTorch()..)` task, when the user *doesn't* override the resources, we send `task_models.Resources(requests=[], limits=[])` (which is different from `None/nil`, see [here](https://github.com/flyteorg/flytekit/blob/master/flytekit/core/resources.py#L78)) to the backend as part of the worker spec. We go into [this](https://github.com/flyteorg/flyteplugins/blob/13ef7d6d6a178b06dffd592c4b49f46b9cda4410/go/tasks/plugins/k8s/kfoperators/common/common_operator.go#L266) `if` statement but not into [this](https://github.com/flyteorg/flyteplugins/blob/13ef7d6d6a178b06dffd592c4b49f46b9cda4410/go/tasks/plugins/k8s/kfoperators/common/common_operator.go#L268) one within the first. As a consequence, in this case we *don't* overwrite the resource request set in `@task(requests=Resources())`. This makes sense.

However, the `Elastic` task config in `flytekitplugins` currently doesn't allow overwriting the resource request since elastic pytorch training jobs don't have a differentiation between master and worker replicas. There are only workers which means we can simply use the normal `@task(requests=Resources())` to specify their resources.

This, however, means that in this case we send `None/nil` to the backend as part of the worker specs resource request. [Here](https://github.com/flyteorg/flyteplugins/blob/13ef7d6d6a178b06dffd592c4b49f46b9cda4410/go/tasks/plugins/k8s/kfoperators/common/common_operator.go#L275) we go into the `else` case and simply delete what the user set as `@task(requests=Resources())`.

This seems wrong and this PR changes this.

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/plugins/k8s/kfoperators/common/common_operator.go
+++ b/go/tasks/plugins/k8s/kfoperators/common/common_operator.go
@@ -268,12 +268,10 @@ func OverrideContainerSpec(podSpec *v1.PodSpec, containerName string, image stri
 				if len(resources.Requests) >= 1 || len(resources.Limits) >= 1 {
 					resources, err := flytek8s.ToK8sResourceRequirements(resources)
 					if err != nil {
-						return flyteerr.Errorf(flyteerr.BadTaskSpecification, "invalid TaskSpecificat ion on Resources [%v], Err: [%v]", resources, err.Error())
+						return flyteerr.Errorf(flyteerr.BadTaskSpecification, "invalid TaskSpecification on Resources [%v], Err: [%v]", resources, err.Error())
 					}
 					podSpec.Containers[idx].Resources = *resources
 				}
-			} else {
-				podSpec.Containers[idx].Resources = v1.ResourceRequirements{}
 			}
 			if len(args) != 0 {
 				podSpec.Containers[idx].Args = args
--- a/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch_test.go
+++ b/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch_test.go
@@ -720,7 +720,7 @@ func TestBuildResourcePytorchV1WithZeroWorker(t *testing.T) {
 	assert.Error(t, err)
 }
 
-func TestParasElasticConfig(t *testing.T) {
+func TestParseElasticConfig(t *testing.T) {
 	elasticConfig := plugins.ElasticConfig{MinReplicas: 1, MaxReplicas: 2, NprocPerNode: 4, RdzvBackend: "c10d"}
 	elasticPolicy := ParseElasticConfig(&elasticConfig)
 	assert.Equal(t, int32(1), *elasticPolicy.MinReplicas)
--- a/go/tasks/plugins/k8s/kfoperators/common/common_operator_test.go
+++ b/go/tasks/plugins/k8s/kfoperators/common/common_operator_test.go
@@ -314,11 +314,13 @@ func TestOverrideContainerSpecEmptyFields(t *testing.T) {
 
 func TestOverrideContainerNilResources(t *testing.T) {
 	podSpec := dummyPodSpec()
+	podSpecCopy := podSpec.DeepCopy()
+
 	err := OverrideContainerSpec(&podSpec, "primary container", "", nil, []string{})
 	assert.NoError(t, err)
 	assert.Equal(t, 2, len(podSpec.Containers))
-	assert.Nil(t, podSpec.Containers[0].Resources.Limits)
-	assert.Nil(t, podSpec.Containers[0].Resources.Requests)
+	assert.Equal(t, podSpec.Containers[0].Resources.Limits, podSpecCopy.Containers[0].Resources.Limits)
+	assert.Equal(t, podSpec.Containers[0].Resources.Requests, podSpecCopy.Containers[0].Resources.Requests)
 }
 
 func dummyTaskContext() pluginsCore.TaskExecutionContext {


go/tasks/plugins/k8s/kfoperators/common/common_operator.go
```diff
@@ -268,12 +268,10 @@ func OverrideContainerSpec(podSpec *v1.PodSpec, containerName string, image stri
 				if len(resources.Requests) >= 1 || len(resources.Limits) >= 1 {
 					resources, err := flytek8s.ToK8sResourceRequirements(resources)
 					if err != nil {
-						return flyteerr.Errorf(flyteerr.BadTaskSpecification, "invalid TaskSpecificat ion on Resources [%v], Err: [%v]", resources, err.Error())
+						return flyteerr.Errorf(flyteerr.BadTaskSpecification, "invalid TaskSpecification on Resources [%v], Err: [%v]", resources, err.Error())
 					}
 					podSpec.Containers[idx].Resources = *resources
 				}
-			} else {
```
fg91, line 275: This is the problematic line.

go/tasks/plugins/k8s/kfoperators/pytorch/pytorch_test.go
```diff
@@ -720,7 +720,7 @@ func TestBuildResourcePytorchV1WithZeroWorker(t *testing.T) {
 	assert.Error(t, err)
 }
 
-func TestParasElasticConfig(t *testing.T) {
+func TestParseElasticConfig(t *testing.T) {
```
fg91, line 723: Scope creep: fix typo

fg91:
@yubofredwang I somehow can't tag you as reviewer but could you please sanity check this? Thanks

hamersaw:
@fg91 just want to make sure I understand the logic here. Basically for the pytorch, MPI, and tensorflow plugins there are three states resources for each replica can be:
- _nil_: this means that they were not provided and we need to set them to empty lists
- _empty list_: we do nothing, this essentially means they were not set in flytekit
- _list_: if provided then we override

This introduces issues because in the elastic plugin the resources will always be `nil` because they are not set. So this PR is changing the handling of `nil` resources. My concern is whether there is a good reason that `nil` resources are set to empty lists rather then leaving them as is.

Setting the resources for each replica in flytekit to empty lists for the elastic plugin would achieve the same result right?

fg91:
> @fg91 just want to make sure I understand the logic here. Basically for the pytorch, MPI, and tensorflow plugins there are three states resources for each replica can be:
> 
> * _nil_: this means that they were not provided and we need to set them to empty lists
> * _empty list_: we do nothing, this essentially means they were not set in flytekit
> * _list_: if provided then we override

Your understanding is correct that there are these three possible states for resources for each replica and that this is the current behaviour.

> Setting the resources for each replica in flytekit to empty lists for the elastic plugin would achieve the same result right?

It would, yes, I first considered doing this, but I personally find the current logic [here](https://github.com/flyteorg/flyteplugins/blob/13ef7d6d6a178b06dffd592c4b49f46b9cda4410/go/tasks/plugins/k8s/kfoperators/common/common_operator.go#L266) odd:

Is there a scenario where we would ever want to [end up with a pod without any resource requests](https://github.com/flyteorg/flyteplugins/blob/13ef7d6d6a178b06dffd592c4b49f46b9cda4410/go/tasks/plugins/k8s/kfoperators/common/common_operator.go#L276) despite having set them in the task decorator's `requests` arg? I could imagine that one sets resources in the task decorator and then chooses lower/higher resources for one specific replica. But no resources at all?

From the perspective of the elastic plugin I find it unintuitive that in order for the resources set in `@task(requests=` to not be deleted, I have to include an empty list in the worker spec's resources and that if I completely omit the resources field there (which is the default), the ones I specified in `@task(requests=` are deleted.

Considering the function [`OverrideContainerSpec`](https://github.com/flyteorg/flyteplugins/blob/13ef7d6d6a178b06dffd592c4b49f46b9cda4410/go/tasks/plugins/k8s/kfoperators/common/common_operator.go#L260C85-L260C94), which is passed a pod spec with certain resources and override `resources`, I find it unintuitive that when I pass empty `resources`, the original ones in the pod spec are preserved but if I pass `resources=nil`, the resources in the pod spec are set to empty resources.

This is why I personally think we should address this in the backend and not in `flytekit` but of course I'm happy to discuss!

I'd be interested in your thoughts on this @yubofredwang

yubofredwang:
> > @fg91 just want to make sure I understand the logic here. Basically for the pytorch, MPI, and tensorflow plugins there are three states resources for each replica can be:
> > 
> > * _nil_: this means that they were not provided and we need to set them to empty lists
> > * _empty list_: we do nothing, this essentially means they were not set in flytekit
> > * _list_: if provided then we override
> 
> Your understanding is correct that there are these three possible states for resources for each replica and that this is the current behaviour.
> 
> > Setting the resources for each replica in flytekit to empty lists for the elastic plugin would achieve the same result right?
> 
> It would, yes, I first considered doing this, but I personally find the current logic [here](https://github.com/flyteorg/flyteplugins/blob/13ef7d6d6a178b06dffd592c4b49f46b9cda4410/go/tasks/plugins/k8s/kfoperators/common/common_operator.go#L266) odd:
> 
> Is there a scenario where we would ever want to [end up with a pod without any resource requests](https://github.com/flyteorg/flyteplugins/blob/13ef7d6d6a178b06dffd592c4b49f46b9cda4410/go/tasks/plugins/k8s/kfoperators/common/common_operator.go#L276) despite having set them in the task decorator's `requests` arg? I could imagine that one sets resources in the task decorator and then chooses lower/higher resources for one specific replica. But no resources at all?
> 
> From the perspective of the elastic plugin I find it unintuitive that in order for the resources set in `@task(requests=` to not be deleted, I have to include an empty list in the worker spec's resources and that if I completely omit the resources field there (which is the default), the ones I specified in `@task(requests=` are deleted.
> 
> Considering the function [`OverrideContainerSpec`](https://github.com/flyteorg/flyteplugins/blob/13ef7d6d6a178b06dffd592c4b49f46b9cda4410/go/tasks/plugins/k8s/kfoperators/common/common_operator.go#L260C85-L260C94), which is passed a pod spec with certain resources and override `resources`, I find it unintuitive that when I pass empty `resources`, the original ones in the pod spec are preserved but if I pass `resources=nil`, the resources in the pod spec are set to empty resources.
> 
> This is why I personally think we should address this in the backend and not in `flytekit` but of course I'm happy to discuss!
> 
> I'd be interested in your thoughts on this @yubofredwang

The reason of the code logic is that I assume when user does not specify resources at all. They want to use the default resources user set for that namespace. Considering the link [here](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/cpu-default-namespace/)

How about having an extra check like this

```
if resources != nil {
  ...
} else if podSepc[containeridx].resources == nil {
  podSpec.Containers[idx].Resources = v1.ResourceRequirements{}
}

Does this fix this bug?

**Comment by fg91** [comment]
> The reason of the code logic is that I assume when user does not specify resources at all. They want to use the default resources user set for that namespace. Considering the link [here](https://kubernetes.io/docs/tasks/administer-cluster/manage-resources/cpu-default-namespace/)

When not specifying any resources at all in a task, e.g.

```python
@task
def train() -> str:
    print("Trained")
```

the respective pod should have the [task defaults configured on the Flyte platform side ](https://github.com/flyteorg/flyte/blob/713f3f86a4d74d555e1a55b52065119df39117e3/charts/flyte-core/values.yaml#L621), not defaults configured e.g. via a `kind: LimitRange`.

In our case the task above runs with

```yaml
    resources:
      limits:
        cpu: "1"
        memory: 2Gi
      requests:
        cpu: "1"
        memory: 2Gi
```

I would expect the same resources for this task:

```python
@task(
    task_config=Elastic(
        nnodes=2,
        nproc_per_node=2,
    )
)
def train() -> str:
    print("Trained")
```

When using flyte 1.8.0 (helm chart, `flytekit`, `flytekitplugins-kfpytorch`), however, the task doesn't have the default resources configured on the platform side:

```yaml
resources: {}
```

In my opinion this is not the expected behaviour.

hamersaw (approved):

PR #387, opened 2023-08-12
Transition to Queue if the JobCondition is empty 

pingsutw:
# TL;DR
image

The plugin manager keeps trying to recreate CRD since [ExtractCurrentCondition](https://github.com/flyteorg/flyteplugins/blob/c528bb88937b4732c9cb5537ed8ea6943ff4fb56/go/tasks/plugins/k8s/kfoperators/common/common_operator.go#L65) returns an error when the condition is empty.

Create CRD -> Condiction is empty -> return error -> retry -> retry limit exceeds -> fail.

I've seen this error many times, it seems like it only happens when using the training operator 1.5.0+.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [x] Unit tests added

## Complete description
image

## Tracking Issue
_NA_

## Follow-up issue
_NA_


hamersaw:
@fg91 @yubofredwang mind taking a look?

hamersaw:
@pingsutw without this fix, if the kf operator is not deployed the same failure will occur right? namely there will be no conditions because no operator picks up the CR. However, with this change, if the operator is not deployed the task will be forever stuck in the `QUEUED` phase right?

pingsutw:
@fg91 correct. Which version of Pytorch operator you was using? seems like the error only happens when using pytorch-operator 1.5+

yubofredwang:
The PR itself looks good. However, if the operator is not installed, can we use [StartTime](https://github.com/kubeflow/training-operator/blob/master/pkg/apis/kubeflow.org/v1/common_types.go#L52C2-L52C11) as an indication of whether operator is installed? 
Like if the StartTime is set, then that means the operator is not missing. We are only waiting for conditions to be updated.

fg91:
> @fg91 correct. Which version of Pytorch operator you was using? seems like the error only happens when using pytorch-operator 1.5+

Yes, we use 1.5+ 👍

yubofredwang:
> > The PR itself looks good. However, can we use [StartTime](https://github.com/kubeflow/training-operator/blob/master/pkg/apis/kubeflow.org/v1/common_types.go#L52C2-L52C11) as an indication of whether operator is installed?
> 
> [app.status](https://github.com/flyteorg/flyteplugins/blob/7ddd99393cb67c837885db21c55c976be7dd7368/go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go#L236) is None when the task is started.

I was thinking maybe we can leverage `objectMeta.CreationTimestamp` and set a timeout for the creation of the TFJob resource? 
objectMeta.CreationTimestamp

pingsutw:
I think that's a good idea, and we should make timeout configurable. wdyt @hamersaw

pingsutw:
updated it. 
image

fg91:
> updated it. image

Nice descriptive error message 👌 
(Only nit-pick: most of - at least our - users won't know what CR means, let's maybe not abbreviate)

yubofredwang:
Looks perfect! Thanks for the quick implementation!

go/tasks/plugins/k8s/kfoperators/pytorch/config.go
```diff
@@ -0,0 +1,31 @@
...
+
+var (
+	defaultConfig = Config{
+		Timeout: config.Duration{Duration: 1 * time.Minute},
+	}
+
+	configSection = pluginsConfig.MustRegisterSubSection("pytorch", &defaultConfig)
+)
+
+// Config is config for 'ray' plugin
```
yubofredwang, line 19: Fix the comment.

pingsutw: good catch. updated it.

go/tasks/plugins/k8s/kfoperators/pytorch/config.go
```diff
@@ -0,0 +1,32 @@
...
+	pluginsConfig "github.com/flyteorg/flyteplugins/go/tasks/config"
+	"github.com/flyteorg/flytestdlib/config"
+)
+
+//go:generate pflags Config --default-var=defaultConfig
+
+var (
+	defaultConfig = Config{
+		Timeout: config.Duration{Duration: 1 * time.Minute},
+	}
```
hamersaw, line 15: qq is this functionality the same for all kfoperator plugins? should we add a config for timeout on tensorflow and mpi as well?

go/tasks/plugins/k8s/kfoperators/common/config.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+//go:generate pflags Config --default-var=defaultConfig
+
+var (
+	defaultConfig = Config{
+		Timeout: config.Duration{Duration: 1 * time.Minute},
+	}
+
+	configSection = pluginsConfig.MustRegisterSubSection("pytorch", &defaultConfig)
```
hamersaw, line 17: if this is in `common` it shouldn't be named `pytorch` right?

pingsutw: sorry, my bad, updated it

go/tasks/plugins/k8s/kfoperators/pytorch/pytorch.go
```diff
@@ -231,6 +231,9 @@ func (pytorchOperatorResourceHandler) GetTaskPhase(_ context.Context, pluginCont
 		return pluginsCore.PhaseInfoUndefined, err
 	}
 
+	if app.Status.StartTime == nil && app.CreationTimestamp.Add(common.GetConfig().Timeout.Duration).Before(time.Now()) {
+		return pluginsCore.PhaseInfoUndefined, fmt.Errorf("kubeflow operator hasn't updated the pytorch coustum resource since creation time %v", app.CreationTimestamp)
```
hamersaw, line 235: "coustum" -> "custom"

same in tensorflow / mpi

go/tasks/plugins/k8s/kfoperators/common/config.go
```diff
@@ -0,0 +1,32 @@
...
+)
+
+//go:generate pflags Config --default-var=defaultConfig
+
+var (
+	defaultConfig = Config{
+		Timeout: config.Duration{Duration: 1 * time.Minute},
+	}
+
+	configSection = pluginsConfig.MustRegisterSubSection("common", &defaultConfig)
```
hamersaw, line 17: Does this result in configuration like:
```
plugins:
    common:
         timeout: 10s
     k8s:
         ...
     ....
```
IMO `common` is not descriptive enough here. Maybe something like `kf-operator`. Thoughts?

pingsutw: That makes sense, updated it, thanks.

hamersaw (approved):

PR #388, opened 2023-08-14
Feat: Expose nodeID to container

fellhorn:
# TL;DR
Exposes the NodeId to the kubernetes Pod as environment variable.

```
- name: FLYTE_INTERNAL_NODE_ID
  value: n0/dn0/dn0/dn0/dn0/dn0/dn0/dn0/dn0/dn0
```

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
Currently the environment variables do not provide enough information to know the exact task being executed. 

For example we intended to create a unique ID for all retries of a task. Without the nodeID this is not possible as tasks might be reused. 

One can retrieve this information also from the hostname but this fails for long names due to k8s pod name length limits.

### DIFF
--- a/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go
+++ b/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go
@@ -31,6 +31,15 @@ func GetContextEnvVars(ownerCtx context.Context) []v1.EnvVar {
 			},
 		)
 	}
+
+	if nodeId := contextutils.Value(ownerCtx, contextutils.NodeIDKey); nodeId != "" {
+		envVars = append(envVars,
+			v1.EnvVar{
+				Name:  "FLYTE_INTERNAL_NODE_ID",
+				Value: nodeId,
+			},
+		)
+	}
 	return envVars
 }
 
--- a/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go
+++ b/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go
@@ -32,11 +32,11 @@ func GetContextEnvVars(ownerCtx context.Context) []v1.EnvVar {
 		)
 	}
 
-	if nodeId := contextutils.Value(ownerCtx, contextutils.NodeIDKey); nodeId != "" {
+	if nodeID := contextutils.Value(ownerCtx, contextutils.NodeIDKey); nodeID != "" {
 		envVars = append(envVars,
 			v1.EnvVar{
 				Name:  "FLYTE_INTERNAL_NODE_ID",
-				Value: nodeId,
+				Value: nodeID,
 			},
 		)
 	}


go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go
```diff
@@ -31,6 +31,15 @@ func GetContextEnvVars(ownerCtx context.Context) []v1.EnvVar {
 			},
 		)
 	}
+
+	if nodeID := contextutils.Value(ownerCtx, contextutils.NodeIDKey); nodeID != "" {
```
hamersaw, line 35: It looks like this ID will be a parent / child hierarchy separated by '/' ([here](https://github.com/flyteorg/flytestdlib/blob/dc81c279d511f271126443b2855843014727e34c/contextutils/context.go#L99-L106)). Do we instead want the nodeID to be identical to that displayed in the UI? We can instead use the `NodeExecutionID.NodeID` and inject [here](https://github.com/flyteorg/flyteplugins/blob/03dc9db2fe948de997734b0d7f2621506a84dd99/go/tasks/pluginmachinery/flytek8s/k8s_resource_adds.go#L73) using something like `id.GetID().NodeExecutionId.NodeId`. Thoughts?

fellhorn: Hi @hamersaw 
For our case we need the entire hierarchy, similar to how is is displayed here in the UI:
image

(ideally without it being shortened to a specific length)

`id.GetID().NodeExecutionId.NodeId` only gives me the child nodeID, e.g. `dn0`, which might occur multiple times in a workflow. Instead we need the entire hierarchy to uniquely identify a single node in the graph (e.g. `n0/dn0/dn0/dn0/dn0`)

hamersaw: If you use something like:
```
currentNodeUniqueID, err := common.GenerateUniqueID(nCtx.ExecutionContext().GetParentInfo(), nodeExecutionID.NodeId)
```
then it should prepend the node ID with the parent info which should give you the `n0-0-dn0` / `n0-0-dn0-0-dn0` being separated by `-` rather than `/`. I think this would be cleaner?

eapolinario:
Hi, we are moving all Flyte development to a monorepo. In order to help the transition period, we're moving open PRs to monorepo automatically and your PR was moved to . Notice that if there are any conflicts in the resulting PR they most likely happen due to the change in the import path of the flyte components.

eapolinario:
Hi, we are moving all Flyte development to a monorepo. In order to help the transition period, we're moving open PRs to monorepo automatically and your PR was moved to https://github.com/flyteorg/flyte/pull/4135. Notice that if there are any conflicts in the resulting PR they most likely happen due to the change in the import path of the flyte components.

PR #391, opened 2023-08-17
Create a FileOutput reader if the agent produce file output

pingsutw:
# TL;DR
Fixed https://flyte-org.slack.com/archives/C05GQ16TQLF/p1692227368597599

Create a `RemoteFileOutputReader` in the agent plugin if the task writes the output to the blob store.
For example, an agent may submit a job to aws batch, which will write the output to s3, so we should add a file output reader to task context.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
^^^

## Tracking Issue
_NA_

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/plugins/webapi/agent/plugin.go
+++ b/go/tasks/plugins/webapi/agent/plugin.go
@@ -5,6 +5,7 @@ import (
 	"crypto/x509"
 	"encoding/gob"
 	"fmt"
+
 	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/io"
 	"github.com/flyteorg/flytestdlib/logger"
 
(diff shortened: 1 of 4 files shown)


pingsutw:
cc @hamersaw would you mind taking a look

go/tasks/plugins/webapi/agent/plugin.go
```diff
@@ -169,17 +171,38 @@ func (p Plugin) Status(ctx context.Context, taskCtx webapi.StatusContext) (phase
...
+	}
+
+	if taskTemplate.Interface == nil || taskTemplate.Interface.Outputs == nil || taskTemplate.Interface.Outputs.Variables == nil {
+		logger.Infof(ctx, "The task declares no outputs. Skipping writing the outputs.")
+		return nil
+	}
+
+	var opReader io.OutputReader
+	if resource.Outputs != nil {
+		logger.Infof(ctx, "Agent returned an output")
```
hamersaw, line 197: These probably shouldn't be `Info` level. If we log every time an agent successfully returns an input that's a ton of unnecessary logs.

go/tasks/plugins/webapi/agent/plugin.go
```diff
@@ -169,17 +171,38 @@ func (p Plugin) Status(ctx context.Context, taskCtx webapi.StatusContext) (phase
...
+		opReader = ioutils.NewInMemoryOutputReader(resource.Outputs, nil, nil)
+	} else {
+		logger.Infof(ctx, "Agent didn't return any output, assuming file based outputs.")
+		opReader = ioutils.NewRemoteFileOutputReader(ctx, taskCtx.DataStore(), taskCtx.OutputWriter(), taskCtx.MaxDatasetSizeBytes())
+	}
+	return taskCtx.OutputWriter().Put(ctx, opReader)
```
hamersaw, line 203: can you give an example of these locations? I'm not following here - we're using the `taskCtx.OutputWriter` to construct an `OutputReader` that just writes back to the `taskCtx.OutputWriter`?

pingsutw: TBH, I'm a bit confused as well. Most of our plugins do this.
- [k8s plugin](https://github.com/flyteorg/flytepropeller/blob/8446baf92748f6079f43aaa94d1b0b97df233a9e/pkg/controller/nodes/task/k8s/plugin_manager.go#L291-L297)
- [map task](https://github.com/flyteorg/flyteplugins/blob/f577f62727a16a282edfa7a49d985e47a4ccef79/go/tasks/plugins/array/outputs.go#L232-L237)
- [sagemaker](https://github.com/flyteorg/flyteplugins/blob/e7bf9a4e1b43ec7be55411fcc31babbe6c769db4/go/tasks/plugins/k8s/sagemaker/custom_training.go#L194-L199)

1. flytekit writes the proto to s3
2. propeller reads proto from s3
3. propeller writes the proto to another s3 path

However, if we don't call `OutputWriter().Put()`, the downstream task won't be able to find the input.

pingsutw: error: `Last Error: UNKNOWN::Output Reader was nil. Plugin/Platform problem.`

pingsutw: Update:
OutputWriter is a [NewBufferedOutputWriter](https://github.com/flyteorg/flyteplugins/blob/e7bf9a4e1b43ec7be55411fcc31babbe6c769db4/go/tasks/pluginmachinery/ioutils/buffered_output_writer.go#L9), so it won't actually write the output to another s3 path, it just [adds the output reader to NewBufferedOutputWriter](https://github.com/flyteorg/flyteplugins/blob/e7bf9a4e1b43ec7be55411fcc31babbe6c769db4/go/tasks/pluginmachinery/ioutils/buffered_output_writer.go#L15), and then propeller will validate the output by reading the data from output reader, which is in the output writer.

hamersaw: thanks for figuring this out! looks great.

go/tasks/plugins/webapi/agent/plugin.go
```diff
@@ -169,17 +171,38 @@ func (p Plugin) Status(ctx context.Context, taskCtx webapi.StatusContext) (phase
...
 }
 
+func writeOutput(ctx context.Context, taskCtx webapi.StatusContext, resource *ResourceWrapper) error {
+	taskTemplate, err := taskCtx.TaskReader().Read(ctx)
+	if err != nil {
+		return err
+	}
+
+	if taskTemplate.Interface == nil || taskTemplate.Interface.Outputs == nil || taskTemplate.Interface.Outputs.Variables == nil {
+		logger.Infof(ctx, "The task declares no outputs. Skipping writing the outputs.")
```
hamersaw, line 191: Can we make this debug too?

pingsutw: done

hamersaw (approved):

PR #395, opened 2023-08-25
Make rendering template in agent client temporary

edwinyyyu:
# TL;DR
A render of the task template in the agent client can persist across multiple nodes. Solution stores the unrendered template in a separate variable and uses it to recover the unrendered template.

## Type
 - [X] Bug Fix
 - [ ] Feature
 - [ ] Plugin

### DIFF
--- a/go/tasks/plugins/webapi/agent/plugin.go
+++ b/go/tasks/plugins/webapi/agent/plugin.go
@@ -68,6 +68,7 @@ func (p Plugin) Create(ctx context.Context, taskCtx webapi.TaskExecutionContextR
 		return nil, nil, err
 	}
 
+	var argTemplate []string
 	if taskTemplate.GetContainer() != nil {
 		templateParameters := template.Parameters{
 			TaskExecMetadata: taskCtx.TaskExecutionMetadata(),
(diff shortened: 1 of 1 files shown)


go/tasks/plugins/webapi/agent/plugin.go
```diff
@@ -101,6 +103,10 @@ func (p Plugin) Create(ctx context.Context, taskCtx webapi.TaskExecutionContextR
 		return nil, nil, err
 	}
 
+	if taskTemplate.GetContainer() != nil {
+		taskTemplate.GetContainer().Args = argTemplate
```
pingsutw, line 107: qq: why do we need to revert args here? did you see any error in the propeller?

edwinyyyu: For tasks with the same task function, the first render will persist, causing all prefixes to correspond to the same node. The {{}} templates will have been replaced so later renders will not find anything to replace.

edwinyyyu: This solution was chosen over makeing a copy of the entire taskTemplate. I'm not familiar with if there is a better place to solve this problem.

pingsutw: LGTM, could you add a comment at line 105?

edwinyyyu: Done.

pingsutw: thanks!

pingsutw (approved):

hamersaw (approved):

PR #396, opened 2023-08-25
Fix: Mounting of secrets in kfoperators #minor

fellhorn:
# TL;DR
First step to solve https://github.com/flyteorg/flyte/issues/3907 . It fixes the bug for kfoperator plugins and allows similar fixes for other k8s based plugins to mount secrets.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested: We tested it with `Elastic`, `Pytorch`, `Tensorflow`, `MPI` and `Horovod`
 - [ ] Unit tests added

## Complete description
Pods need specific labels and annotations that the mutating webhook picks them up and performs the mutations to mount the respective secrets. For plugins this did not work as these annotations were applied on the plugin object and not the `Pod` object (e.g. `PytorchJob` instead of `Pod`)

## Follow-up issue
https://github.com/flyteorg/flyte/issues/3907 still needs to be fixed for e.g. the `Dask` plugin

### DIFF
--- a/go/tasks/pluginmachinery/flytek8s/pod_helper.go
+++ b/go/tasks/pluginmachinery/flytek8s/pod_helper.go
@@ -14,6 +14,7 @@ import (
 	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/core/template"
 	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/flytek8s/config"
 	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/utils"
+	"github.com/flyteorg/flyteplugins/go/tasks/pluginmachinery/utils/secrets"
 
 	"github.com/flyteorg/flytestdlib/logger"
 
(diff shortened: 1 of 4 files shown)


go/tasks/pluginmachinery/flytek8s/pod_helper.go
```diff
@@ -115,6 +116,27 @@ func mergeMapInto(src map[string]string, dst map[string]string) {
...
 
+func AppendSecretMountingMetadata(objectMeta *metav1.ObjectMeta, taskTemplate core.TaskTemplate) error {
+	secretsMap := make(map[string]string)
+	injectSecretsLabel := make(map[string]string)
+	var err error
+
+	if taskTemplate.SecurityContext != nil && len(taskTemplate.SecurityContext.Secrets) > 0 {
+		secretsMap, err = secrets.MarshalSecretsToMapStrings(taskTemplate.SecurityContext.Secrets)
+		if err != nil {
+			return err
```
EngHabu, line 127:
```suggestion
			return fmt.Errorf("failed to marshal secrets. Error: %w", err)
```

go/tasks/plugins/k8s/kfoperators/mpi/mpi.go
```diff
@@ -157,6 +157,12 @@ func (mpiOperatorResourceHandler) BuildResource(ctx context.Context, taskCtx plu
 		return nil, fmt.Errorf("number of launch worker should be more then 0")
 	}
 
+	errSecret := flytek8s.AppendSecretMountingMetadata(objectMeta, *taskTemplate)
+
+	if errSecret != nil {
+		return nil, errSecret
```
EngHabu, line 163: nit:
```suggestion
		return nil, fmt.Errorf("failed to append secrets. Error: %w", errSecret)
```

hamersaw (commented):
It looks like these annotations are added for the k8s container / pod objects (and all top-level objects) in the [plugin manager in propeller](https://github.com/flyteorg/flytepropeller/blob/476d4165f6e9d1df52d00f17b0e9ea3726e0bdc8/pkg/controller/nodes/task/k8s/plugin_manager.go#L106-L107) after being set during[ TaskExecutionContext creation](https://github.com/flyteorg/flytepropeller/blob/476d4165f6e9d1df52d00f17b0e9ea3726e0bdc8/pkg/controller/nodes/task/k8s/task_exec_context.go#L46-L66). Is the TaskExecutionContext where you got this code from?

I'm wondering if there is a better solution here than copying this to plugins as well. Could we just use the former (ie. merging the `tCtx.GetAnnotations()` with those already on the `objectMeta`) and save ourselves duplicated code? This would result in all of the annotations / labels associated with the task being inlcuded as well, which may mitigate future issues. And if this is the route we choose to go, we should probably merge with the [default annotations and labels](https://github.com/flyteorg/flyteplugins/blob/552f145c4632944a23e7e2d9fa2b2f065c4b14b9/go/tasks/pluginmachinery/flytek8s/config/config.go#L78-L81) as well right?

fellhorn:
As @hamersaw pointed out this is better solved in a more general approach and I would prefer to include https://github.com/flyteorg/flyteplugins/pull/401 instead of this PR. Thus will close it.

Thanks for your review @EngHabu

PR #398, opened 2023-09-02
Add supportTaskTypes for agentservice without write it in config twice.

Future-Outlier:
# TL;DR
As title.

## Type
 - [ ] Bug Fix
 - [x] Feature
 - [x] Plugin

## Are all requirements met?

 - [x] Code completed
 - [ ] Smoke tested
 - [x] Unit tests added

## Complete description
Right now, the agent-service section doesn't need to add supportedTaskTypes anymore.
We can get it by "default-for-task-types" from tasks.
image
image

## Tracking Issue
https://github.com/flyteorg/flytepropeller/pull/612


pingsutw (approved):

pingsutw (approved):

honnix:
I'm not entirely sure this is a good direction to go. As I understood "default-for-task-types" is only needed when there is registration conflict, ie two plugins claiming to support the same type of task. This change to me feels like slightly abusing the semantics of "default-for-task-types". In our production setup for example, we have never needed to use "default-for-task-types", and now with this change we will be forced to use it.

pingsutw:
> As I understood "default-for-task-types" is only needed when there is registration conflict, ie two plugins claiming to support the same type of task

@honnix sorry, This makes sense. We'll revert it shortly.

PR #400, opened 2023-09-05
Fix resources requests for AWS Batch plugin

hamersaw:
# TL;DR
This PR uses the container resources as the base before merging with resource overrides and platform defaults. The logic involved in [this PR](https://github.com/flyteorg/flytepropeller/pull/544) updated always setting the resource overrides even if there were none (if missing it would set overrides to the container resources), which introduced this regression.

## Type
 - [x] Bug Fix
 - [ ] Feature
 - [ ] Plugin

## Are all requirements met?

 - [x] Code completed
 - [x] Smoke tested
 - [ ] Unit tests added

## Complete description
^^^

## Tracking Issue
https://github.com/flyteorg/flyte/issues/4003

## Follow-up issue
_NA_

### DIFF
--- a/go/tasks/plugins/array/awsbatch/transformer.go
+++ b/go/tasks/plugins/array/awsbatch/transformer.go
@@ -88,11 +88,9 @@ func FlyteTaskToBatchInput(ctx context.Context, tCtx pluginCore.TaskExecutionCon
 		return nil, err
 	}
 
-
 	if overrideResources := tCtx.TaskExecutionMetadata().GetOverrides().GetResources(); overrideResources != nil {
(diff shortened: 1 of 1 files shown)


pingsutw (approved):

eapolinario (approved):
Does https://github.com/flyteorg/flytepropeller/pull/544 also affect other plugins?

hamersaw:
> Does [flyteorg/flytepropeller#544](https://github.com/flyteorg/flytepropeller/pull/544) also affect other plugins?

Good question, as far as I can tell no. Other plugins use the container parsing we define. AWS Batch is unique because it uses the resources to populate a different object (ie. aws batch job).

